PR #30, opened 2022-12-15
fix(store): fix typings

FabianLars:


FabianLars:
hmm, no idea how that happened

PR #32, opened 2022-12-16
fix(persisted-scope): fix patterns usage (fix #25)

### LINKED ISSUE #25: [persisted-scope] Asset protocol not configured correctly for directories after restart
I may be mistaken as I don't know all the intricate of Tauri or Rust, but it seems to me if I use a `dialog.open` as a directory selector it will [allow the path with allow_directory](https://github.com/tauri-apps/tauri/blob/2901145c497299f033ba7120af5f2e7ead16c75a/core/tauri/src/endpoints/dialog.rs#L199). While the current code when deserializing only use allow_file, rendering is useless ?

I think it might be the cause of the problem I currently have, at least when working with the asset-protocol (feature is enabled)

[Minimal reproduction](https://github.com/OrIOg/vanilla-folder/)
___
Versions:
tauri-plugin-persisted-scope: Used with a github

Tauri info
```
Environment
  › OS: Windows 10.0.19042 X64
  › Webview2: 101.0.1210.32
  › MSVC:
      - Visual Studio Community�2019
      - Visual Studio Build Tools�2019
  › Node.js: 17.3.0
  › npm: 8.3.0
  › pnpm: Not installed!
  › yarn: 1.22.19
  › rustup: 1.25.1
  › rustc: 1.65.0
  › cargo: 1.65.0
  › Rust toolchain: stable-x86_64-pc-windows-msvc

Packages
  › @tauri-apps/api [NPM]: 1.2.0
  › tauri [RUST]: 1.2.2,
  › tauri-build [RUST]: 1.2.1,
  › tao [RUST]: 0.15.6,
  › wry [RUST]: 0.23.3,

App
  › build-type: bundle
  › CSP: default-src 'self' https://asset.localhost; media-src 'self' https://asset.localhost; img-src 'self' https://asset.localhost; script-src 'self' https://asset.localhost; connect-src https://asset.localhost
  › distDir: ../dist
  › devPath: http://localhost:3000/
  › framework: Vue.js
  › bundler: Vite

App directory structure
  ├─ .git
  ├─ .vscode
  ├─ dist
  ├─ node_modules
  ├─ public
  ├─ scripts
  ├─ src
  ├─ src-tauri
  └─ tauri-plugin-persisted-scope
```

OrIOg:
PR Draft for fixing #25([explanation](https://github.com/tauri-apps/plugins-workspace/issues/25#issuecomment-1353677934))

Current change breaks old `.persistent-scope` files deserialization.
Can be fixed by moving the pattern conversion in deserialization.

Using [metadata](https://github.com/tauri-apps/tauri/commit/911587692cd22071c8f8fc8361d90a17a71806ef) with Tauri v1.

[Suggestion](https://github.com/tauri-apps/tauri/issues/5875#issue-1503785783) for Tauri v2
Could be easier/cleaner if `FsScopeEvent` gave the "type" of allowed path(i.e : file, directory, recursive directory).
Might want to make a PR in Tauri itself.

As a Rust/Tauri neophyte, there must be a better way to handle that mutex/remove it.
Might be fixable when if we reuse the old serialization method if we move the pattern conversion in deserialization.

Any suggestions are welcome

### DIFF
--- a/plugins/fs-watch/README.md
+++ b/plugins/fs-watch/README.md
@@ -57,7 +57,7 @@ import { watch, watchImmediate } from "tauri-plugin-fs-watch-api";
 const stopWatching = await watch(
   "/path/to/something",
   (event) => {
-    const { type, payload } = event;
+    const { kind, path } = event;
   },
   { recursive: true }
 );
@@ -65,7 +65,7 @@ const stopWatching = await watch(
 const stopRawWatcher = await watchImmediate(
   ["/path/a", "/path/b"],
   (event) => {
-    const { path, operation, cookie } = event;
+    const { type, paths, attrs } = event;
   },
   {}
 );


vasilvestre:
I think I have this issue. Should we wait for a merge or I can apply this patch somehow ?

FabianLars:
We can't merge it as long as it depends on a git version of tauri (and it would be ideal if we can implemented it in a way that does not need a tauri update to not break apps that can't upgrade tauri for some reason - that part is not a requirement)

PR #50, opened 2023-01-06
refactor: use builder style

JonasKruckenberg:
Use the `plugin::Builder` and `init` style to unify plugin usage.

This affects the following plugins:
- `auth`
- `fs-watch`
- `fs-extra`
- `sql`
- `upload`


FabianLars:
One thing this PR made me notice is that the path in lint-rust.yml is wrong :sweat_smile: i guess it should be something like `"plugins/*/src/**"` instead?

FabianLars (approved):

PR #52, opened 2023-01-06
refactor(store): rename plugin

JonasKruckenberg:
Rename the plugin builder struct from `PluginBuilder` to `Builder` to align with conventions

### DIFF
--- a/plugins/store/src/lib.rs
+++ b/plugins/store/src/lib.rs
@@ -244,12 +244,12 @@ async fn save<R: Runtime>(
 }
 
 #[derive(Default)]
-pub struct PluginBuilder {
+pub struct Builder {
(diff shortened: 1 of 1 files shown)


FabianLars:
This one is a bit tricky because it also has `StoreBuilder`. I guess that could cause confusion?

JonasKruckenberg:
Hmm I mean yeah, maybe? But rustc will tell you if you mixed the types up and IMO it's not thaaaat hard after all 🤔

FabianLars:
I'm a bit biased because of the help desk damage but I'll trust you here :P

FabianLars (approved):

JonasKruckenberg:
If this causes any helpdesk posts you are allowed to ping me 😉

PR #54, opened 2023-01-06
Sync all repos

JonasKruckenberg:
Previously only the websocket plugin was being synced as our test case. This PR pulls the trigger on syncing all plugin repos

### DIFF
--- a/plugins/mirrors.txt
+++ b/plugins/mirrors.txt
@@ -1,1 +1,14 @@
+authenticator
+autostart
+fs-extra
+fs-watch
+localhost
+log
+persisted-scope
+positioner
+sql
+store
+stronghold
+upload
 websocket
+window-state


Beanow:
Something I'd like to see, for now and future sanity is perhaps a GH Action test:
Do repos with the expected names exist?
Maybe some other high-level quick to validate sanity check.

JonasKruckenberg:
> Do repos with the expected names exist?

It actually does that, check the shell script!

It will also not commit and push any changes if that specific subdirectory didn't get any changes.

Beanow:
Right what I meant to say was, as part of the pull request test suite.
Not only as emergency stop when the merge was done.

JonasKruckenberg:
> Right what I meant to say was, as part of the pull request test suite.

Ahh 🤔 hm yeah some additional PR checks is definitely a good idea

Beanow:
:laughing: ok I was reviewing.
But I guess watching the build log will be faster now.

PR #55, opened 2023-01-06
[sql] Make sqlite the default database driver

JonasKruckenberg:
We talked about how annoying the current state of the sql plugin is and that is breaks virtually all code-related tooling there is. Since the vast majority of people use sqlite anyway (and they should) this PR makes it the actual default flag.


JonasKruckenberg:
Actually, no let's come up with something nicer here

PR #56, opened 2023-01-06
feat: add single-instance plugin

JonasKruckenberg:


Beanow:
@JonasKruckenberg is there any way we can preserve the history of the repos? It's something I missed on the other plugins, where blames/contributor analysis/etc are not visible other than the mirror repos.

JonasKruckenberg:
I mean I'm all for going with your PR, it just seems to have been a bunch more work than I wanted to put into it 😅
And I was sort of expecting we need to do major overhauls on most plugins anyway, but you're right having the history preserved is much better

PR #72, opened 2023-01-07
feat: Add single-instance

Beanow:
Alternative to #56
Here preserving git history.

:warning: Should not be squash-merged. As that would probably squash the entire history again.

~~:warning: Missing some improvements & integrations available in #56.
This is just a "dumb mv" so far.~~
Edit: this now has all the changes.

I'm all for emotional safety, but I've got to blame people :grimacing:

- [Blame on #56](https://github.com/tauri-apps/plugins-workspace/blame/64fc10e147ce306432cb30366b55f9c6724a186d/plugins/single-instance/src/platform_impl/windows.rs)
- [Blame on #72](https://github.com/tauri-apps/plugins-workspace/blame/a9020b7b9099800ae9da7d388329477540ba47ca/plugins/single-instance/src/platform_impl/windows.rs)

 

That said, references by number may be wrong here.
image


JonasKruckenberg (approved):
Has too many unwraps and missing error handling for my liking, but since that's not a concern of this PR, looks good to me 👍🏻 

Edit: the failing checks aren't a concern

amrbashir (changes requested):
I would actually remove the `emit-event` example, and just include a call to `app.emit_all` inside the `vanilla` example

JonasKruckenberg:
> I would actually remove the `emit-event` example, and just include a call to `app.emit_all` inside the `vanilla` example

The idea was that we wanted to preserve the old repo as much as possible first and them do some cleanup, but I agree that in this case it would make sense to remove this

amrbashir (approved):

PR #78, opened 2023-01-10
fix: wrong path patterns in `files` property, fixes #76

### LINKED ISSUE #76: d.ts is not included
After running `npm add https://github.com/tauri-apps/tauri-plugin-store` I couldn't import the module

`import { Store } from "tauri-plugin-store-api";` doesn't work.

I checked the node_modules, it seems like it's because no `d.ts` is present in the dist folder.

FabianLars:


tsconfig.base.json
```diff
@@ -16,5 +16,5 @@
     "declaration": true,
     "declarationDir": "dist"
```
FabianLars, line 17: Does this need to be changed too? doesn't seem to make a difference 🤔

PR #80, opened 2023-01-10
docs(autostart): add import to `README.md` usage example

osbre:
This change can make life a little bit easier for beginners :)


JonasKruckenberg (approved):
Oups yeah nice catch 👍🏻

PR #85, opened 2023-01-11
chore: Add tauri-plugin-websocket history

JonasKruckenberg:
This PR adds the history of the `websocket` plugin for better git blame and for preserving history like #72


plugins/single-instance/examples/vanilla/src-tauri/Cargo.toml
```diff
@@ -1,5 +1,7 @@
+[workspace]
+
 [package]
-name = "single-instance-example"
+name = "app"
```
Beanow: Looks like a change in different plugin.

plugins/sql/src/plugin.rs
```diff
@@ -348,7 +348,9 @@ impl Builder {
                     drop(lock);
 
                     app.manage(instances);
-                    app.manage(Migrations(Mutex::new(self.migrations.take().unwrap_or_default())));
+                    app.manage(Migrations(Mutex::new(
+                        self.migrations.take().unwrap_or_default(),
+                    )));
```
Beanow, line 353: Ditto, change in other plugin.

JonasKruckenberg: it's `cargo fmts` fault 🤷🏻‍♂️ 😂

plugins/single-instance/examples/vanilla/src-tauri/Cargo.toml
JonasKruckenberg: true, though the name really doesn't matter much

JonasKruckenberg:
```suggestion
name = "single-instance-example"
```

plugins/sql/src/plugin.rs
FabianLars: And that's why i should stop using the web editor for small changes in rust files :))

PR #88, opened 2023-01-12
fix: change dist for js files, fixes #76 for yarn3

### LINKED ISSUE #76: d.ts is not included

FabianLars:
See https://github.com/tauri-apps/plugins-workspace/issues/76 for reference. yarn3 likes to be really weird again :)) Not the only solution but the one i'd like the most i think. didn't like adding the whole guest-js folder to files as shown in the linked issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 target
 node_modules
 dist
+dist-js
--- a/plugins/authenticator/package.json
+++ b/plugins/authenticator/package.json
@@ -7,20 +7,20 @@
     "Tauri Programme within The Commons Conservancy"
   ],
   "type": "module",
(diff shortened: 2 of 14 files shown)


JonasKruckenberg (approved):
I have no idea why this works when the old setup didn't, but if you can confirm it does that's enough for me 👍

FabianLars:
yeah no idea why. i guess yarn berry is being yarn berry again 😞

PR #104, opened 2023-01-23
feat(window-state): add flags to control what is saved/restored

### LINKED ISSUE #101: [window-state] persist to much information (decoration: true/false is persisted)
Only tested on Linux! Gnome 43

We have an app with `decorations: false` and used the window-state plugin.

We then changed our mind about the decorations and flipped the flag `decorations: true`

But if the window-state plugin has been started with decorations: false that state is restored, so our window doesn't get the decorations :)

amrbashir:
closes #101

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -235,7 +235,7 @@ checksum = "d9b39be18770d11421cdb1b9947a45dd3f37e93092cbf377614828a319d5fee8"
 dependencies = [
  "hermit-abi 0.1.19",
  "libc",
- "winapi 0.3.9",
+ "winapi",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


JonasKruckenberg:
I like this because our saving/restoring of the fullscreen property isn't handled correctly

amrbashir:
> I like this because our saving/restoring of the fullscreen property isn't handled correctly

How so? as far as my testing goes, it saves/restores correctly

JonasKruckenberg (approved):
Works great in elk and with your PR that fullscreen restoring bug seems to be gone too 👍🏻

FabianLars:
Before merging it we need to update the plugin's readme as it still shows the old ShowMode

FabianLars:
2 seconds too slow lol

PR #108, opened 2023-01-24
feat(store): expose Store to Rust code

### LINKED ISSUE #13: [store] Expose `Store.cache` as `pub` to be easily used from Rust code
Currently it's `pub(crate)` preventing me from directly using the `Store` in my Rust logic. I can submit a PR.

### LINKED ISSUE #73: [store] Public struct StoreCollection, so that I can get store from state to access store from backend
I need to access store from backend. But the StoreCollection is private, so that I can not get the store from manage state.

JonasKruckenberg:
closes #73, closes #13

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -235,7 +235,7 @@ checksum = "d9b39be18770d11421cdb1b9947a45dd3f37e93092cbf377614828a319d5fee8"
 dependencies = [
  "hermit-abi 0.1.19",
  "libc",
- "winapi 0.3.9",
+ "winapi",
 ]
 
 [[package]]
@@ -610,7 +610,7 @@ checksum = "f4ffc801dacf156c5854b9df4f425a626539c3a6ef7893cc0c5084a23f0b6c59"
 dependencies = [
  "atty",
  "lazy_static",
- "winapi 0.3.9",
+ "winapi",
 ]
 
 [[package]]
@@ -2458,7 +2458,7 @@ checksum = "f346ff70e7dbfd675fe90590b92d59ef2de15a8779ae305ebcbfd3f0caf59be4"
 dependencies = [
  "autocfg",
  "bitflags",
- "cfg-if 1.0.0",
+ "cfg-if",
  "libc",
  "memoffset",
  "pin-utils",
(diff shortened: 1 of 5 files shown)


FabianLars:
I feel like a small usage example would be really beneficial here - ig a code comment should be enough? (i actually just dislike example apps lol)

Also the fields of StoreCollection are private without a getter function so it still doesn't work right?

JonasKruckenberg:
> I feel like a small usage example would be really beneficial here - ig a code comment should be enough? (i actually just dislike example apps lol)

Yeah I have to figure some things out first, because this actually doesn't work as-is bc of generic inference and things. 

> Also the fields of StoreCollection are private without a getter function so it still doesn't work right?

People should use `with_store` since that handles the whole fallback and freezing logic. But maybe we can make that a method of `StoreCollection` for easier use?

FabianLars:
> People should use with_store since that handles the whole fallback and freezing logic

nevermind, i misinterpreted what `with_store` is.

> But maybe we can make that a method of StoreCollection for easier use?

Interesting idea, yeah

acurrieclark:
Been doing some experimenting on this branch. I am hoping to be able to share and access store values in rust and on the client.

```ts
const store = new Store(".settings.dat");
await store.set("test-documents", ["a", "b"]);
console.log(await store.get("test-documents")); // outputs ["a", "b"]
```

```rust
use tauri_plugin_store::with_store;
use tauri_plugin_store::StoreCollection;
use std::path::PathBuf;
use tauri::Manager;
use tauri::Wry;

let stores = app.state::<StoreCollection<Wry>>();
let path = PathBuf::from(".settings.dat");
with_store(app.handle(), stores, path, |store| {
    let docs = store.get("test-documents").ok_or("couldn't find value");
    println!("{:?}", docs);
    Ok(())
})?;
```

Which outputs `couldn't find value` on the rust end. Removing the `set` from the client code results in the value still being found on the client, but the same error on rust.

Attempting to use `set` in the closure throws an error: 
```
no method named `set` found for mutable reference `&mut Store<tauri_runtime_wry::Wry<EventLoopMessage>>` in the current scope
```

NB. I have already posted this on discord, posting here as it seemed a better place to report (what appear to me to be) bugs.

FabianLars:
> Attempting to use set in the closure throws an error:

On the rust side it's called `insert` not `set`.

FabianLars:
i think the ergonomics could be improved slightly by using generics like `Into<String>` or something but lets stick with this for a while and see how it works 👍 

(p.s. the above discussion was resolved on discord and it was basically a misunderstanding)

FabianLars (approved):

PR #125, opened 2023-02-13
fix(autostart): fix auto start choosing wrong app path on macOS

morajabi:
I'm releasing my fork into production on noor.to, will report back if it indeed worked as I expect or needs further refinement.

More description in the code comments


morajabi:
CleanShot 2023-02-13 at 13 41 26@2x

It worked. 
`auto_start path /Applications/Noor.app`

FabianLars (approved):

PR #129, opened 2023-02-15
fix(stronghold) return null if there is no record

Ludea:
return null if there is no record into store

### DIFF
--- a/plugins/stronghold/guest-js/index.ts
+++ b/plugins/stronghold/guest-js/index.ts
@@ -300,10 +300,10 @@ export class Store {
       key: toBytesDto(key),
     }).then((v) => {
          if (v !== null || undefined) {
-          Uint8Array.from(v)
+          return Uint8Array.from(v);
          }
          else {
-          null
+          return null;
         }
       })
   }
--- a/plugins/stronghold/guest-js/index.ts
+++ b/plugins/stronghold/guest-js/index.ts
@@ -300,7 +300,7 @@ export class Store {
       key: toBytesDto(key),
     }).then((v) => {
          if (v !== null || undefined) {
-          Uint8Array.from(v);
+          Uint8Array.from(v)
          }
          else {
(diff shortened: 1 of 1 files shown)


plugins/stronghold/guest-js/index.ts
```diff
@@ -298,7 +298,14 @@ export class Store {
...
       key: toBytesDto(key),
-    }).then((v) => Uint8Array.from(v));
+    }).then((v) => {
+         if (v !== null || undefined) {
+          Uint8Array.from(v);
+         }
+         else {
+          null
+        }
+      })
```
FabianLars: You're not returning anything here so it will always be null (or undefined?) no matter if the value exists or not.

FabianLars (changes requested):

plugins/stronghold/guest-js/index.ts
Ludea: Better ?

FabianLars: same issue still. That only work in Rust, in javascript/typescript you'll have to use the `return` keyword here

plugins/stronghold/guest-js/index.ts
```diff
@@ -298,7 +298,14 @@ export class Store {
...
       client: this.client,
       key: toBytesDto(key),
-    }).then((v) => Uint8Array.from(v));
+    }).then((v) => {
+         if (v !== null || undefined) {
+          Uint8Array.from(v)
+         }
+         else {
+          null
+        }
```
FabianLars: Something like this i guess
```suggestion
         if (v !== null || undefined) {
          return Uint8Array.from(v)
         }
         else {
          return null
        }
```

Ludea: done

FabianLars (approved):

PR #255, opened 2023-02-15
chore: Add MSRV to all readmes

FabianLars:


plugins/authenticator/README.md
```diff
@@ -4,6 +4,8 @@ Use hardware security-keys in your Tauri App.
 
 ## Install
 
+_This plugin requires a Rust version of at least**1.64**_
```
lorenzolewis: Add a space before the version on all the readme's

lorenzolewis:
I have some other changes on content not in this PR but will take a pass at that when I start to pull content into the docs site

plugins/authenticator/README.md
FabianLars: Whoops using find&replace is not one of my strengths apparently

PR #259, opened 2023-02-20
feat: add path plugin

lucasfernog:


lucasfernog:
This is a good example for a mobile/desktop plugin, but we will move it to the core crates.

Issue #204, opened 2022-03-09
[sql] MySql TIMESTAMP returns null


simmslabs:
MySql TIMESTAMP returns null

FabianLars:
We only really support a small-ish subset of data types, mainly focused on sqlite, which we consider the main use case, especially since 99% of the times actual sql servers should be behind some kind of api server and not directly accessed from clients...

That said, another PR added `DATETIME` support so maybe you can use this instead (although tbh i'm thinking about removing it again 🤔)

Going to close this for now unless @lucasfernog thinks this is indeed in scope (if it is, we should think about other datatypes too)

HarshilSharaf:
@FabianLars Even i am facing when trying a simple `select * from users ;` query the type `created_on` returns null as it is of type datetime in MySQL db..... Can we expect this feature in near future or not as this are some basic types used in every project?

FabianLars:
@HarshilSharaf 

Can you run `cargo update` in your src-tauri dir and try again (make sure you consume the rust side of the plugin from this repo, not the mirror)? There was a commit which was supposed to fix this.

If it still doesn't work please create a new issue with a minimal reproduction example. Thank you.

HarshilSharaf:
> @HarshilSharaf
> 
> Can you run `cargo update` in your src-tauri dir and try again (make sure you consume the rust side of the plugin from this repo, not the mirror)? There was a commit which was supposed to fix this.
> 
> If it still doesn't work please create a new issue with a minimal reproduction example. Thank you.

@FabianLars Thank you for a quick reply.
Just FYI i just created a new tauri repo 2 days back.. But i still try running cargo update

HarshilSharaf:
> @HarshilSharaf
> 
> Can you run `cargo update` in your src-tauri dir and try again (make sure you consume the rust side of the plugin from this repo, not the mirror)? There was a commit which was supposed to fix this.
> 
> If it still doesn't work please create a new issue with a minimal reproduction example. Thank you.

Hi @FabianLars i've tried running `cargo update` and after that also the datatypes like `TIMESTAMP` , `DATETIME` and `DATE` are returned as null

I've tried this on the tables of my own database as well as some inbuilt databases that came with `MySQL` workbench

Can you please check what is wrong with it?

Also, if possible please add support for `ENUM` dataType in future.


MySQL and MySQL workbench version : `8.0.31`

The code that i am using in angular frontend to query database is as follows:
[...]

PR #266, opened 2023-03-02
feat(sql): Improve type decoding. Adds support for time and dates

FabianLars:
and mysql's enum type. ref https://github.com/tauri-apps/plugins-workspace/issues/204#issuecomment-1450518137

The string returned from date/datetime columns may be different in format than what was inserted but working with these types was super annoying that i want to leave it as-is for now...


FabianLars:
(reviewed by awesome community members)

PR #269, opened 2023-03-06
Allow to use local timezone instead of UTC timezone when writing log

pqnga:
We want to print the local time in the log file instead of the UTC (we are in UTC+7). This PR allows to choose the preferable timezone when writing the log.

### DIFF
--- a/plugins/log/src/lib.rs
+++ b/plugins/log/src/lib.rs
@@ -21,9 +21,11 @@ use tauri::{
 };
 
 pub use fern;
+use time::OffsetDateTime;
 
 const DEFAULT_MAX_FILE_SIZE: u128 = 40000;
 const DEFAULT_ROTATION_STRATEGY: RotationStrategy = RotationStrategy::KeepOne;
+const DEFAULT_TIMEZONE_STRATEGY: TimezoneStrategy = TimezoneStrategy::UseUtc;
 const DEFAULT_LOG_TARGETS: [LogTarget; 2] = [LogTarget::Stdout, LogTarget::LogDir];
 
 /// An enum representing the available verbosity levels of the logger.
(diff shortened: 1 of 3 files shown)


plugins/log/src/lib.rs
```diff
@@ -83,6 +85,24 @@ pub enum RotationStrategy {
...
+}
+
+impl TimezoneStrategy {
+    pub fn get_now(&self) -> OffsetDateTime {
+        match self {
+            TimezoneStrategy::UseUtc => OffsetDateTime::now_utc(),
+            TimezoneStrategy::UseLocal => match OffsetDateTime::now_local() {
+                Ok(v) => v,
+                Err(_) => OffsetDateTime::now_utc() // Fallback to UTC since Rust cannot determine local timezone
+            }
```
FabianLars: I think this could be simplified to something like this
```suggestion
            TimezoneStrategy::UseLocal => OffsetDateTime::now_local().unwrap_or_else(|_| OffsetDateTime::now_utc())
```

FabianLars (changes requested):
Thanks for contributing! Just one small thing and then i think it's good to go. Edit: Plus the compiler error of course 😅

plugins/log/src/lib.rs
FabianLars: Aaand please run `cargo fmt` so that CI is happy again

FabianLars (approved):

pqnga:
@FabianLars Hey I do not know how to fix the audit-rust https://github.com/tauri-apps/plugins-workspace/actions/runs/4343174730/jobs/7584917093

Is it required to pass audit-rust to merge this PR? Thanks.

FabianLars:
you can ignore that. it's caused by some nested dependency and fails for all PRs

pqnga:
@FabianLars Thank you. Could you merge it to dev branch as well?

PR #276, opened 2023-03-10
fix(window-state): Use outer position

parker-codes:
Previously, using `.inner_position()` would shift the window down slightly (approximately the size of the title bar) each time the app was started. Using `.inner_size()` still seems to correctly set the window size though.


amrbashir (approved):

amrbashir:
thanks @parker-codes

PR #294, opened 2023-03-31
ci: Replace actions-rs/audit-check with rustsec fork

FabianLars:
They finally decided to maintain it for now and recently released an update. rustsec is the same org that hosts the advisory database so imo it should be fairly trustworthy.

This PR is a follow-up of https://github.com/tauri-apps/plugins-workspace/pull/45 where we replaced all other actions-rs actions.

### DIFF
--- a/.github/workflows/audit-rust.yml
+++ b/.github/workflows/audit-rust.yml
@@ -28,6 +28,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions-rs/audit-check@v1
+      - uses: rustsec/audit-check@v1
         with:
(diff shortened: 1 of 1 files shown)


FabianLars:
`@Beanow` Requesting your review because you participated in #45.

Beanow:
I think you're referring to https://github.com/rustsec/audit-check/issues/2#issuecomment-1484963758

Seems like we should report a bug though:
> Warning: Unknown warning kind unsound found, please, file a bug

FabianLars:
Alrighty, tried to fix it myself but i don't understand the action at all 😅

FabianLars:
https://github.com/rustsec/audit-check/issues/8

PR #295, opened 2023-03-31
fix(store): Load state from disk and extend defaults

rming:


FabianLars:
Could you fix the issue CI is complaining about (run `cargo fmt`)? Then we can merge this i think :)

rming:
> Could you fix the issue CI is complaining about (run `cargo fmt`)? Then we can merge this i think :)

format fixed

FabianLars (approved):

PR #306, opened 2023-04-14
feat(dialog): add plugin

lucasfernog:


lucasfernog:
Includes the existing desktop implementation + Android and iOS APIs (except folder picker and file save dialog, which aren't possible and doesn't make sense on mobile).

handexing:
@lucasfernog  Hello. Can folder picker be supported? Some scenarios still require it, such as the export function. I hope users can select a directory before exporting. Or are there any other solutions?

PR #308, opened 2023-04-15
feat(fs): add plugin

lucasfernog:


lucasfernog:
This is also just a port of existing and audited code from the tauri codebase.

PR #309, opened 2023-04-17
feat(window-state): add js api, closes #254

### LINKED ISSUE #254: [window-state] Does not work with custom titlebar (appWindow.close)
When I close the app with [`appWindow.close`](https://tauri.studio/en/docs/usage/guides/visual/window-customization/#js) and reopen the app, the window position is not preserved.

It seems that closing app with `appWindow.close` does not trigger a `CloseRequested` event, so the state is not read.

If I close the window with system titlebar and reopen the app, the window position is preserved. So I can confirm it's about `appWindow.close`.

amrbashir:
closes #254


amrbashir:
Not sure why prettier doesn't like the new `tsconfig.json` even though it is just a copy of the one in `upload` plugin

FabianLars:
> Not sure why prettier doesn't like the new tsconfig.json even though it is just a copy of the one in upload plugin

It's interpreted as a text file instead of a symlink. Newer git versions _should_ be able to convert the windows mklink stuff to git symlinks if configured correctly (didn't try it myself yet). If that doesn't work you'll need wsl or something to create a unix symlink :/

I meant to explore this a bit and write down some form of guide since this repo is not really windows friendly at the moment 😅

amrbashir:
Ah, that makes sense

plugins/window-state/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "tauri-plugin-window-state-api",
+  "version": "0.1.0",
+  "description": "Save window positions and sizse and restore them when the app is reopened.",
```
FabianLars:
```suggestion
  "description": "Save window positions and sizes and restore them when the app is reopened.",
```

FabianLars:
I'm not the biggest fan of the `js_` prefix, what do you think about moving the commands into a `mod cmd` in that same file instead?
Otherwise lgtm 👍

plugins/window-state/guest-js/index.ts
```diff
@@ -0,0 +1,35 @@
...
+/**
+ *  Restore the specified window state from disk
+ */
+async function restoreState(label: WindowLabel, flags: StateFlags) {
+  invoke("plugin:window-state|js_restore_state", { label, flags });
+}
+
+/**
+ *  Restore the current window state from disk
+ */
```
FabianLars:
```suggestion
/**
 *  Save the state of all open windows to disk.
 */
async function saveWindowState(flags: StateFlags) {
  invoke("plugin:window-state|js_save_window_state", { flags });
}

/**
 *  Restore the state for the specified window from disk.
 */
async function restoreState(label: WindowLabel, flags: StateFlags) {
  invoke("plugin:window-state|js_restore_state", { label, flags });
}

/**
 *  Restore the state for the current window from disk.
 */
```

FabianLars (approved):

PR #310, opened 2023-04-17
ci: use original commiter author instead of actor

amrbashir:


FabianLars (approved):
Let's just try it out lol

PR #317, opened 2023-04-17
feat: add API example

lucasfernog:
This PR moves the API example from the tauri repository to this repo, since we'll be moving all our APIs to individual plugins.


amrbashir:
I don't think we should do this tbh, the API example has purposes that could only be full-filled if it is kept in the tauri repo:
1. test-bed for new features developed in tauri repo
2. an example for new users to see what tauri could do which will be only discoverable if it exists in the main tauri repo

amrbashir:
And even though the APIs are moved here, we are probably going to be re-exporting them through tauri and the `api` example shouldn't be affected by this at all

FabianLars:
> we are probably going to be re-exporting them through tauri

I sure hope we won't. At least not all of them (my prime example is always the cli ~~abomination~~ module).

With the plugin approach we also solve the "How do i use the rust functions without enabling the js counterparts", provided we re-add the allowlist into the outsourced plugins of course.

That said, i kinda agree that this repo may not be the best place for the example, but i'm not sure if the main repo is any better either from a DX pov at least 😅

lucasfernog:
For core testbed I think we should have a dedicated example on the tauri repo. Discoverability should be managed by the website IMO, it's hard to find the app even in the main repo, so ideally we distribute the binary and expose it in the website IMO.

PR #322, opened 2023-04-18
Fixed that types are invalid

### LINKED ISSUE #321: tauri-plugin-store-api does not work with ts
It seems that `tauri-plugin-store-api` does not specify the correct path for the types:

in package.json the path is:
```
"types": "./dist-js/index.d.ts",
```
However the correct path would be:    
```
"types": "./dist-js/dist/index.d.ts",
```

DavidSichau:
Fixed the path to the types file.

Closes #321

### DIFF
--- a/plugins/store/package.json
+++ b/plugins/store/package.json
@@ -9,10 +9,10 @@
   "type": "module",
   "browser": "dist-js/index.min.js",
   "module": "dist-js/index.mjs",
-  "types": "dist-js/index.d.ts",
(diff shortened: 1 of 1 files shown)


FabianLars:
Thanks for your contribution! I fixed the root issue in https://github.com/tauri-apps/plugins-workspace/commit/7acf865ffbb82c6d00334926a4527dd5ff3163a4 (especially because it was an issue in all plugins) so this PR is not needed anymore, thank you again though!

PR #324, opened 2023-04-18
feat(global-shortcut): add plugin

lucasfernog:


plugins/global-shortcut/src/lib.rs
```diff
@@ -0,0 +1,286 @@
...
+            .map_err(Into::into)
+    }
+
+    pub fn unregister_all<S: IntoIterator<Item = HotKey>>(&self, shortcuts: S) -> Result<()> {
+        acquire_manager(&self.manager)?
+            .unregister_all(&shortcuts.into_iter().collect::<Vec<HotKey>>())
+            .map_err(Into::into)
+    }
+
+    pub fn is_registered(&self, shortcut: HotKey) -> bool {
```
amrbashir: We should note that this function only checks if this app already registered this shortcut and not whether the OS has the shortcut registered

plugins/global-shortcut/src/lib.rs
```diff
@@ -0,0 +1,286 @@
+use std::{
+    collections::HashMap,
+    str::FromStr,
+    sync::{Arc, Mutex},
+};
+
+pub use global_hotkey::hotkey::{Code, HotKey, Modifiers};
```
amrbashir: if we are exporting `HotKey`, should we re-export it as `Shortcut` to have consistency with the naming in the plugin?

plugins/global-shortcut/guest-js/index.ts
```diff
@@ -0,0 +1,128 @@
...
+  handler: ShortcutHandler
+): Promise<void> {
+  return await invoke('plugin:globalShortcut|register_all', {
+    shortcuts,
+    handler: transformCallback(handler)
+  })
+}
+
+/**
+ * Determines whether the given shortcut is registered by this application or not.
```
amrbashir, line 79:
```suggestion
 * Determines whether the given shortcut is registered by this application or not.
 * 
 * If the shortcut is registered by another application, it will still return `false`.
 *
```

amrbashir (changes requested):

lucasfernog:
I'm mostly just porting existing code to this repo for now, documentation is something we'll need to work a lot later. But i'll resolve these suggestions soon.

amrbashir (approved):

PR #326, opened 2023-04-19
feat(notification): add plugin

lucasfernog:


plugins/notification/guest-js/index.ts
```diff
@@ -0,0 +1,189 @@
...
+ * ```
+ * It is recommended to allowlist only the APIs you use for optimal bundle size and security.
+ * @module
+ */
+
+import { invoke } from '@tauri-apps/api/tauri'
+
+type PermissionState = 'granted' | 'denied'
+
+  ; (function () {
```
FabianLars: I feel like we have to move this blog into `js_init_script`, right?

FabianLars (changes requested):

plugins/notification/guest-js/index.ts
lucasfernog: Yeah, thanks for catching it :D

PR #328, opened 2023-04-24
fix(persisted-scope): Prevent memory leaks, fixes #274

### LINKED ISSUE #274: [persisted-scope] Memory leak in tauri-plugin-persisted-scope
### Describe the bug

[tauri-plugin-persisted-scope](https://github.com/tauri-apps/tauri-plugin-persisted-scope) v0.1.0 causes a memory leak until OOM.
Not using the plugin fixes the OOM.

### Reproduction

```
tauri::Builder::default()
    .plugin(tauri_plugin_persisted_scope::init())
    .expect("error while running tauri application")
```

### Expected behavior

No memory leak :)

### Platform and versions

```bash
$ tauri info

Environment
  › OS: Arch Linux Unknown X64
  › Node.js: 16.13.2
  › npm: 9.3.0
  › pnpm: 6.11.0
  › yarn: 3.3.1
  › rustup: 1.25.1
  › rustc: 1.63.0
  › cargo: 1.63.0
  › Rust toolchain: stable-x86_64-unknown-linux-gnu 

Packages
WARNING: no lock files found, defaulting to npm
  › @tauri-apps/api [NPM]: Not installed!
  › tauri [RUST]: 1.2.3,
  › tauri-build [RUST]: 1.2.1,
  › tao [RUST]: 0.15.8,
  › wry [RUST]: 0.23.4,
```

### Stack trace

_No response_

### Additional context

_No response_

FabianLars:
It's not actually a memory _leak_ the plugin just breaks the .persisted-scope file, bloating it up (to 100mb after a few uses/restarts if you actually use the scope) to a point where it gets too large to work correctly.

This PR adds more pattern than strictly necessary to fix existing files seamlessly - well, hopefully at least.

Fixes #274

### DIFF
--- a/.changes/persisted-scope-fix-oom.md
+++ b/.changes/persisted-scope-fix-oom.md
@@ -1,0 +1,5 @@
+---
+persisted-scope: patch
+---
+
+Recursively unescape saved patterns before allowing/forbidding them. This effectively prevents `.persisted-scope` files from blowing up, which caused Out-Of-Memory issues, while automatically fixing existing broken files seamlessly.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4362,6 +4362,7 @@ dependencies = [
 name = "tauri-plugin-persisted-scope"
 version = "0.1.0"
 dependencies = [
+ "aho-corasick",
  "bincode",
  "log",
  "serde",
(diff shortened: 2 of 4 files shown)


FabianLars:
We may have to look into deleting files that are too large before reading them in the future because some users reported OOM crashes. In the PR tests Windows didn't report any memory usage raise when reading a 50MB, but no idea if Rust just handles this extremely well or if Windows just doesn't want to report it in time (even though it has like 5 secs lol)

Edit: Tested it with the current dev branch and Windows does report the insane memory usage correctly so maybe Rust is indeed doing a great job here 👀 
Edit2: Yes, both of these edits were added before i initially posted this comment 😂

PR #329, opened 2023-04-24
ci: set up covector

FabianLars:
I saw #53 too late but this PR has a few fixes more (plus more local testing probably) anyway so whatever 🤷

### DIFF
--- a/.changes/config.json
+++ b/.changes/config.json
@@ -1,0 +1,177 @@
+{
+  "gitSiteUrl": "https://github.com/tauri-apps/plugins-workspace/",
+  "pkgManagers": {
+    "javascript": {
+      "version": true,
+      "getPublishedVersion": "pnpm view ${ pkgFile.pkg.name } version",
+      "prepublish": ["pnpm build"],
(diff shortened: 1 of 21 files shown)


.changes/config.json
```diff
@@ -0,0 +1,171 @@
...
+          "dryRunCommand": true,
+          "pipe": true
+        }
+      ]
+    }
+  },
+  "packages": {
+    "authenticator": {
+      "path": "./plugins/authenticator",
+      "manager": ""
```
amrbashir, line 38: why aren't you specifying a manager?

FabianLars: so that covector skips it since we don't want to publish it yet

FabianLars: would be prettier if json would support comments :D (or if covector would have an explicit `disabled` setting maybe? idk)

amrbashir: ah, well, everything looks fine to me

FabianLars:
Thanks Amr :) Since this whole repo is like one big experiment i'll just go ahead and see what happens 😂

PR #339, opened 2023-05-02
feat(ci): update covector getPublishedVersion

lucasfernog:
Use the same script as tauri to support bumps across different releases e.g. v1 vs v2 vs alphas.

### DIFF
--- a/.changes/config.json
+++ b/.changes/config.json
@@ -3,12 +3,12 @@
   "pkgManagers": {
     "javascript": {
       "version": true,
-      "getPublishedVersion": "pnpm view ${ pkgFile.pkg.name } version",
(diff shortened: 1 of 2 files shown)


FabianLars (approved):
Kinda expected the cargo one to fail because of workspace inheritance but looks like it works, cool.

lucasfernog:
Current version is received from covector, and all crate versions are read from the crates.io API (`https://crates.io/api/v1/crates/$crate-name`).

FabianLars:
Yeah exactly, and covectors inheritance support seems to be a bit lacking, but anyway, the current approach would have failed too if that was an issue x)

PR #340, opened 2023-05-02
feat(notification): implement Android and iOS APIs

lucasfernog:


FabianLars (approved):
(Take this review with a grain of salt since i can't test it, but it _looks_ good :D )

lucasfernog:
Heavily inspired by https://github.com/ionic-team/capacitor-plugins/tree/main/local-notifications/android/src/main/java/com/capacitorjs/plugins/localnotifications (I still need to find out where I should put this info). I basically took their implementation and refactored it, fixed some TODOs and changed the API to work better with our Rust backend (that's why you'll see an "enum with associated data" in Kotlin lol).

PR #345, opened 2023-05-07
feat(app): add plugin

lucasfernog:


plugins/app/README.md
```diff
@@ -0,0 +1,65 @@
+![plugin-app](banner.jpg)
```
beanow-at-crabnebula: Think the banner isn't included.

plugins/app/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+<!-- description -->
```
beanow-at-crabnebula: Missing an important: what does it do, what is it for?

wusyong (commented):
Do we need `plugins/app/Cargo.lock`? Other plugin doesn't seem to have it.
Beside plugin serves as library. I feel like it should exclude the lock file.

plugins/app/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Install
+
+_This plugin requires a Rust version of at least **1.64**_
```
beanow-at-crabnebula, line 7: Should probably add listing to top-level readme as well.

plugins/app/README.md
lucasfernog: We'll need to request it form Alve later for all new plugins

plugins/app/README.md
FabianLars: If you mean the MSRV, it's already there all the way at the bottom. If you mean the plugin should be added to the table then you're right, completely slipped my mind for the other plugins too...

FabianLars (approved):

PR #346, opened 2023-05-08
feat(os): add plugin

lucasfernog:


FabianLars (changes requested):
The same reviews from the `app` plugin PR apply here, about the plugin's lockfile and the readme description.

FabianLars (approved):
This one is really a prime candidate for a refactoring in the future imo, i don't think the struct approach maps well from the old `Cmd` to this one.

lucasfernog:
@FabianLars can you come up with a proposal in the future? :P thanks

plugins/os/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+            commands::kind,
+            commands::arch,
+            commands::tempdir
+        ])
+        .setup(|app, _api| {
+            app.manage(OperatingSystem(app.clone()));
+            Ok(())
+        })
+        .build()
+}
```
FabianLars:
```suggestion
use std::{fmt::Display, path::PathBuf};

pub use os_info::Version;
use tauri::{
    plugin::{Builder, TauriPlugin},
    AppHandle, Manager, Runtime,
};

mod commands;
mod error;

pub use error::Error;
type Result<T> = std::result::Result<T, Error>;

pub enum Kind {
    Linux,
    Windows,
    Darwin,
    IOS,
    Android,
}

impl Display for Kind {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        match self {
            Self::Linux => write!(f, "Linux"),
            Self::Windows => write!(f, "Linux_NT"),
            Self::Darwin => write!(f, "Darwin"),
            Self::IOS => write!(f, "iOS"),
            Self::Android => write!(f, "Android"),
        }
    }
}

pub fn platform() -> &'static str {
    match std::env::consts::OS {
        "windows" => "win32",
        "macos" => "darwin",
        _ => std::env::consts::OS,
    }
}

pub fn version() -> Version {
    os_info::get().version().clone()
}

pub fn kind() -> Kind {
    #[cfg(target_os = "linux")]
    return Kind::Linux;
    #[cfg(target_os = "windows")]
    return Kind::Windows;
    #[cfg(target_os = "macos")]
    return Kind::Darwin;
    #[cfg(target_os = "ios")]
    return Kind::IOS;
    #[cfg(target_os = "android")]
    return Kind::Android;
}

pub fn arch() -> &'static str {
    std::env::consts::ARCH
}

pub fn tempdir() -> PathBuf {
    std::env::temp_dir()
}

pub fn init<R: Runtime>() -> TauriPlugin<R> {
    Builder::new("os")
        .invoke_handler(tauri::generate_handler![
            commands::platform,
            commands::version,
            commands::kind,
            commands::arch,
            commands::tempdir
        ])
        .build()
}
```

FabianLars (commented):
hmm, i guess that actually depends on if we will add more apis to that plugin, that may actually end up using the apphandle for something.
If you expect that something like this will be added then nevermind me and apologies!
If not, then i really only thought about simply removing the struct wrapper and having plain functions, like this:

plugins/os/src/lib.rs
FabianLars: holyyy, whole file suggestions look pretty horrible

lucasfernog: Well I guess we won't need the handle, this is a pretty simple plugin anyway.

PR #347, opened 2023-05-08
Avoid type conflict

ptazithos:
Avoid type conflict.

> Argument of type '{ file: string; line: number; }' is not assignable to parameter of type 'LogOptions'.
> Type '{ file: string; line: number; }' is not assignable to type 'Record<string, string>'.
> Property 'line' is incompatible with index signature.
> Type 'number' is not assignable to type 'string'.ts(2345)
> No quick fixes available

### DIFF
--- a/plugins/log/guest-js/index.ts
+++ b/plugins/log/guest-js/index.ts
@@ -4,7 +4,7 @@ import { listen, UnlistenFn } from "@tauri-apps/api/event";
 export type LogOptions = {
   file?: string;
   line?: number;
-  extras?: Record<string, string | undefined>;
+  keyValues?: Record<string, string | undefined>;
 };
 
 enum LogLevel {
@@ -51,7 +51,7 @@ async function log(
     return name.length > 0 && location !== "[native code]";
   });
 
-  const { file, line, extras = {} } = options ?? {};
+  const { file, line, keyValues = {} } = options ?? {};
 
   let location = filtered?.[0]?.filter((v) => v.length > 0).join("@");
   if (location === "Error") {
@@ -64,7 +64,7 @@ async function log(
     location,
     file,
     line,
-    keyValues: extras,
+    keyValues,
   });
 }
 
--- a/plugins/log/guest-js/index.ts
+++ b/plugins/log/guest-js/index.ts
@@ -51,7 +51,7 @@ async function log(
     return name.length > 0 && location !== "[native code]";
   });
 
-  const { file, line, keyValues = {} } = options ?? {};
+  const { file, line, keyValues } = options ?? {};
 
   let location = filtered?.[0]?.filter((v) => v.length > 0).join("@");
   if (location === "Error") {
--- a/plugins/log/guest-js/index.ts
+++ b/plugins/log/guest-js/index.ts
@@ -37,7 +37,7 @@ enum LogLevel {
    *
    * Designates very serious errors.
    */
-  Error
+  Error,
 }
 
 async function log(
@@ -64,7 +64,7 @@ async function log(
     location,
     file,
     line,
-    keyValues: extras
+    keyValues: extras,
   });
 }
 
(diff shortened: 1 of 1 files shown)


FabianLars:
Thanks for contributing! Could you also share a code example that's not working for you so we see more clearly what you're fixing here? 🙏

ptazithos:
> Thanks for contributing! Could you also share a code example that's not working for you so we see more clearly what you're fixing here? 🙏

Sure. The line's type here is number, but Record<string, string | undefind> requires all values in this JSON to be string | undefined.
```typescript
export type LogOptions = {
    file?: string;
    line?: number;
} & Record<string, string | undefined>;
```
Screenshot 2023-05-08 232258

FabianLars:
Ah i see, thanks. The problem with this fix though is that it's also a logic change (or rather a change to the function signature) since this:
```ts
info(message, { file: "test.rs", line: 12, whatever: "some string" });
```
will now be
```ts
info(message, { file: "test.rs", line: 12, extra: { whatever: "some string" } });
```

I _think_ the correct approach would be something like this?
```ts
export type LogOptions = {
  file?: string;
  line?: number;
  [key: string]: string | unknown;
};
```

----

I didn't create the Plugin so maybe i'm completely misinterpreting the actual intentions of the original code 😅

ptazithos:
> Ah i see, thanks. The problem with this fix though is that it's also a logic change (or rather a change to the function signature) since this:
> 
> ```ts
> info(message, { file: "test.rs", line: 12, whatever: "some string" });
> ```
> 
> will now be
> 
> ```ts
> info(message, { file: "test.rs", line: 12, extra: { whatever: "some string" } });
> ```
> 
> I _think_ the correct approach would be something like this?
> 
> ```ts
> export type LogOptions = {
>   file?: string;
>   line?: number;
>   [key: string]: string | unknown;
> };
> ```
> 
> I didn't create the Plugin so maybe i'm completely misinterpreting the actual intentions of the original code sweat_smile

Unknown does make more sense for compatibility. But the extra key-values will later be passed to Rust side. And their type has been fixed to String. Using Unknown will accept other types more than string, which will lead to an error.
```Typescript
  const { file, line, ...keyValues } = options ?? {};

  await invoke("plugin:log|log", {
    level,
    message,
    location,
    file,
    line,
    keyValues,
  });
```
```Rust
fn log(
    level: LogLevel,
    message: String,
    location: Option<&str>,
    file: Option<&str>,
    line: Option<u32>,
    key_values: Option<HashMap<String, String>>,
) {
    let location = location.unwrap_or("webview");
    let mut builder = RecordBuilder::new();
    builder
        .level(level.into())
        .target(location)
        .file(file)
        .line(line);

    let key_values = key_values.unwrap_or_default();
    let mut kv = HashMap::new();
    for (k, v) in key_values.iter() {
        kv.insert(k.as_str(), v.as_str());
    }
    builder.key_values(&kv);

    logger().log(&builder.args(format_args!("{message}")).build());
}
```

FabianLars:
`unknown` was a typo, i meant `undefined`, sorry

ptazithos:
> `unknown` was a typo, i meant `undefined`, sorry

undefined doesn't work for me.

> Argument of type '{ file: string; line: number; }' is not assignable to parameter of type 'LogOptions'.
> Property 'line' is incompatible with index signature.
> Type 'number' is not assignable to type 'string'.

 To be honest, I think ts doesn't have a good solution for this case.

FabianLars:
Okay i guess i'm misremembering my last use-case for this 😩 Alrighty, then let's go with your approach but let's rename `extras` to `keyValues` to match the rust side :)

ptazithos:
> Okay i guess i'm misremembering my last use-case for this weary Alrighty, then let's go with your approach but let's rename `extras` to `keyValues` to match the rust side :)

Sure. Thanks.
Renaming has been done.

FabianLars (approved):
thank you :)

FabianLars:
Ah, i just approved but i noticed one last thing, sorry. We don't need to initialize it with `= {}` since key_values on the Rust side is wrapped in an Option so it can handle undefined too.

FabianLars:
Thanks again 😅

PR #355, opened 2023-05-12
refactor: move deleted tauri APIs, prepare for next release

lucasfernog:
Keeping up with https://github.com/tauri-apps/tauri/pull/6925


plugins/app/src/commands.rs
```diff
@@ -12,8 +12,7 @@ pub fn name<R: Runtime>(app: AppHandle<R>) -> String {
 
 #[tauri::command]
 pub fn tauri_version() -> &'static str {
-    // TODO: return actual tauri version with `tauri::VERSION`
-    env!("CARGO_PKG_VERSION")
+    tauri::VERSION
```
amrbashir, line 15: should we change this to use the newly added `tauri::version()` in dev? https://github.com/tauri-apps/tauri/commit/7c237209207cd2938df660b6fd87d3b7d728bd03

amrbashir (approved):

plugins/app/src/commands.rs
amrbashir: or should we change the api in `dev` (still unreleased) to a `const` variable like `next`?

lucasfernog: It should be a const unless we parse the version string to semver::Version IMO

amrbashir: cool, I will update the `dev` branch then

PR #361, opened 2023-05-16
feat: add init scripts

lucasfernog:
Ref https://github.com/tauri-apps/tauri/pull/6975


plugins/window/src/scripts/hotkey.js
```diff
@@ -0,0 +1,6 @@
+// Copyright 2019-2023 Tauri Programme within The Commons Conservancy
```
amrbashir: do we really need this hotkey script? seems like it could be replaced with a simple `keydown` event handler

amrbashir (approved):

plugins/window/src/scripts/hotkey.js
lucasfernog: Fixed it

amrbashir (approved):

PR #362, opened 2023-05-16
feat: documentation for download func.

louremipsum:
# Need
Even though the `upload-plugin` supports download functionality too, there is no documentation regarding this so I added 
a brief documentation about the download functionality along with a small example similar to the upload function

## Changes Implemented

image

image
In the console.log message, it was written download even though it is an upload function, same with the associated comment

image
Code example for download function

### DIFF
--- a/plugins/upload/README.md
+++ b/plugins/upload/README.md
@@ -1,6 +1,7 @@
 ![plugin-upload](banner.png)
 
 Upload files from disk to a remote server over HTTP.
+Download files from a remote HTTP server to disk
 
 ## Install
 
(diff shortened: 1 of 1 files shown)


FabianLars:
Thanks for contributing!

First time i really look at it tbh, but shouldn't it be Content-Type instead of ContentType? 🤔

louremipsum:
> Thanks for contributing!
> 
> First time i really look at it tbh, but shouldn't it be Content-Type instead of ContentType? 🤔

hmm, didn't really think much about that
I tested it and weirdly both 'Content-Type ' and 'ContentType' seems to work for some reason
like this
image
and this both works
image

plugins/upload/README.md
```diff
@@ -1,6 +1,7 @@
 ![plugin-upload](banner.png)
 
 Upload files from disk to a remote server over HTTP.
+Download files from a remote HTTP server to disk
```
FabianLars:
```suggestion
Download files from a remote HTTP server to disk.
```

A bit picky but that triggered me :D

FabianLars (commented):
Hmm that's interesting! I think we should change it to Content-Type nonetheless though to match all the resources out there.

Also, the Content-Type header probablyyy doesn't make much sense for the download example but i can't think of a better example right now either so let's keep it until we add proper plugin docs (where we can go a bit more in-depth)

plugins/upload/README.md
louremipsum: haha, my bad
please check the commit now

FabianLars (approved):
Thank you!

PR #366, opened 2023-05-17
feat(ci): also test on Windows and macOS

lucasfernog:


amrbashir (approved):

lucasfernog:
Sucks that we need to wait 10 min just to install openssl :(

amrbashir:
why do we need it in the first place? shouldn't the vendored feature flag take care of that? also we may need to have a smarter tests, eg not test all crates but only the crates that changed and the ones that depend on it.

lucasfernog:
Yeah we do need smarter tests. Don't ask me about the feature flag, I thought it would work too. The test is kinda dumb in enabling all features too, could be problematic in the future (we're already enabling both `native-tls` and `rustls` which is weird).

lucasfernog:
For the smarter tests we could do something clever with https://github.com/dorny/paths-filter but it'll be nasty. Specially with dependencies between plugins.

PR #368, opened 2023-05-17
feat(plugins): use `@tauri-apps` namespace on NPM

lucasfernog:
Initially I wanted to use `@tauri-apps` but the conflict between the tauri CLI `@tauri-apps/cli` and the CLI plugin name sucks.


amrbashir:
I think we should instead be using `@tauri-apps/plugin-*` so it feels familiar and more authentic and if we ever move to `@tauri/*` we can still have the CLI and the CLI plugin under the same org.

plugins/dialog/README.md
```diff
@@ -18,19 +18,19 @@ Install the Core plugin by adding the following to your `Cargo.toml` file:
 
 ```toml
 [dependencies]
-tauri-plugin-dialog-api = { git = "https://github.com/tauri-apps/plugins-workspace", branch = "v2" }
+@tauri-apps/plugin-dialog = { git = "https://github.com/tauri-apps/plugins-workspace", branch = "v2" }
```
amrbashir:
```suggestion
tauri-plugin-dialog = { git = "https://github.com/tauri-apps/plugins-workspace", branch = "v2" }
```

plugins/fs-watch/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "tauri-plugin-fs-watch-api",
+  "name": "@tauri-apps/plugin-fs-watch",
```
amrbashir, line 2: could we just remove `fs-wtach` and `fs-extra` plugins and merge them to `fs` plugin instead?

amrbashir (changes requested):

plugins/fs-watch/package.json
lucasfernog: `fs-extra` was removed already, but there's some documentation to fix.

lucasfernog: and yeah `fs-watch` should be on `fs` under a feature flag. We can do that later.

amrbashir: definitely before release, seems redundant to release it now and remove it later.

amrbashir (approved):

plugins/fs-watch/package.json
lucasfernog: I'll work on it.

PR #375, opened 2023-05-18
fix: adjust Cargo features to fix android build

lucasfernog:


lucasfernog:
I'm not sure if `native-tls-vendored` should be a default feature on the plugins, even though it's the easy way to make CI pass.

wusyong:
I don't believe rustls has less problem than nativetls since the platform it supports is more limited.
Could we just update the CI script to fix it?

lucasfernog:
For Android we need to use rustls anyway though. I'll try do adjust the CI.

PR #376, opened 2023-05-18
chore(ci): reuse path filters

lucasfernog:


.github/plugins-filter.yml
```diff
@@ -0,0 +1,81 @@
+tauri-plugin-app:
+  - .github/workflows/lint-rust.yml
```
amrbashir, line 2: note that this file is specific to `lint-rust.yml` only and should be changed to `test-rust.yml` for the test workflow.

lucasfernog: Ahh let's keep it duplicated for now then

PR #382, opened 2023-05-21
chore: add remaining mirrors

lucasfernog:


lucasfernog:
I'll create the missing repos tomorrow.

wusyong (approved):

FabianLars:
I went ahead and created the repos. Still had the muscle memory from the other 20 😂 

Do we really need one for single-instance? I made that old repo read-only a while back instead of adding as a mirror because it only had a rust side. Do we want it for consistency too?

FabianLars:
Okay nevermind, i lost access to the single-instance mirror so you gotta do it if you think we should use it again :)

lucasfernog:
We can't really use the mirror for Rust-only deps since we do not resolve the workspace fields in Cargo.toml when mirroring. So we should also deprecate localhost and window-state.

lucasfernog:
Oh window-state has JS API too. Weird :joy:

FabianLars:
window-state recently got js apis, but persisted-scope doesn't and won't

FabianLars:
> Oh window-state has JS API too. Weird 😂

Yeah for undecorated windows where the CloseRequested api isn't triggered

lucasfernog:
Thanks for creating the repos btw. I went on checking which ones were missing and they were all there already, so I was like `wtf???? is there a workflow creating this too?` ofc the workflow is fabian.yml :joy:

plugins/mirrors.txt
```diff
@@ -6,14 +7,17 @@ dialog
 fs
 global-shortcut
 http
-localhost
 log
+os
 persisted-scope
```
FabianLars:
```suggestion
```

PR #387, opened 2023-05-22
fix: exclude debug only function in release mode

simonhyll:
Resolves the following error
image


lucasfernog (approved):
I was sure I had added that, thanks!

PR #391, opened 2023-05-23
feat(os): add `locale` API

lucasfernog:
Ref https://github.com/tauri-apps/tauri/pull/5960


plugins/os/guest-js/index.ts
```diff
@@ -127,5 +127,22 @@ async function tempdir(): Promise<string> {
...
+ * @example
+ * ```typescript
+ * import { locale } from '@tauri-apps/plugin-os';
+ * const locale = await locale();
+ * if (locale) {
+ *    // use the locale string here
+ * }
+ * ```
+ *
+ * @since 1.3.0
```
amrbashir: Should probably say 2.0 since this plugin won't have a 1.x version

amrbashir (changes requested):

PR #392, opened 2023-05-23
feat(window): add `setEffects` API

lucasfernog:
Ref https://github.com/tauri-apps/tauri/pull/6442

Needs alpha.10 :)


examples/api/src-tauri/src/lib.rs
```diff
@@ -71,6 +71,11 @@ pub fn run() {
                     .decorations(false);
             }
 
+            #[cfg(target_os = "macos")]
```
amrbashir, line 74: Why not windows as well?

plugins/window/guest-js/index.ts
```diff
@@ -900,6 +900,30 @@ class WindowManager extends WebviewWindowHandle {
...
+      value: effects,
+    });
+  }
+
+  /**
+   * Clear any applied effects if possible.
+   *
+   * @since 2.0
+   */
+  async clearEffects(): Promise<void> {
```
amrbashir, line 970: Should we add an explicit clear effects method on tauri's side? Instead of relying on None?

amrbashir (approved):

examples/api/src-tauri/src/lib.rs
lucasfernog: Windows is already handled on the block above.

plugins/window/guest-js/index.ts
lucasfernog: I think setting None is quite idiomatic for Rust isn't it? For JavaScript it looks weird..

PR #415, opened 2023-06-05
fix(stronghold) data race on initialization and wrong argument name

vdang-crabnebula:
Small PR on two bugs that I found while looking at the Stronghold plugin

- [minor: Interface breaks] Data race on initialization of Stronghold (you could call Stronghold functions during initialization). New initialization function
- [patch] Fix wrong argument name

I have poor knowledge of JS so the data race fix might be improvable to not change the Stronghold interface


lucasfernog:
I didn't see your PR and pushed #416 sorry :D looking forward for the other fix though!

PR #427, opened 2023-06-07
refactor(os): refactor EOL const

amrbashir:


lucasfernog (approved):
Fixed the init script for you @amrbashir remember to always use the API example :)

PR #428, opened 2023-06-08
feat(http): refactor and improvements

amrbashir:


lucasfernog:
We need to keep client options somehow though.
```ts
interface ClientOptions {
  /**
   * Defines the maximum number of redirects the client should follow.
   * If set to 0, no redirects will be followed.
   */
  maxRedirections?: number;
  connectTimeout?: number | Duration;
}
```

amrbashir:
yeah I think, I will add a function overload to take a 3rd argument for these options

lucasfernog:
Sounds good. Could also add it to the second argument I think (along with body, headers etc right?).

amrbashir:
It could, I will have to test, but I think it makes sense to have it separated so it doesn't conflict with web spec at all now or in the future.

amrbashir:
I ended up extending the second argument.

james138138:
@lucasfernog @amrbashir Thanks to create plugin. We have use case where application needs to call self signed intranet api. Could we expand scope from `HttpAllowlistScope(pub Vec<String>)` to HttpAllowlistScope(pub Vec<(String, bool)>)` , the 2nd item can be used to accept invalid certs for that domain? I'm happy to help if agreed.

amrbashir:
@james138138 Please open another issue to discuss your feature.

plugins/http/guest-js/index.ts
```diff
@@ -30,519 +30,138 @@ declare global {
...
-  fileName?: string;
+async function readBody<T>(rid: number, kind: "blob" | "text"): Promise<T> {
+  return await window.__TAURI_INVOKE__("plugin:http|fetch_read_body", {
+    rid,
+    kind,
+  });
 }
 
-type Part = string | Uint8Array | FilePart<Uint8Array>;
+class TauriResponse extends Response {
```
amrbashir: Is there a reason why we need to extend and override `Response` methods? Was `new Response(dataAsBytes)` not enough?

lucasfernog (dismissed):

lucasfernog (approved):

PR #429, opened 2023-06-08
fix: remove tsconfig symlink

amrbashir:
git symlinks on Windows behaves weirdly and rollup is not able to pick it up, some plugins were already not using a symlink so decided to remove the symlink.


JonasKruckenberg (approved):
After last week I think we should remove ts entirely actually

amrbashir:
I agree, our JavaScript logic is pretty simple that we don't need typescript or bundling

PR #431, opened 2023-06-10
feat(updater): refactor and improvements

### LINKED ISSUE #461: [updater] allow user to change the updater endpoint during runtime
## Description

Sometimes due to the host server domain change,  we may let the user to change the endpoint to upgrade the app instead of shipping the new binaries.

amrbashir:
needs https://github.com/tauri-apps/tauri/pull/7174 and tauri@2.0.0-alpha.10


plugins/updater/src/lib.rs
```diff
@@ -3,38 +3,26 @@
...
-
-struct UpdaterState {
-    target: Option<String>,
-    config: Config,
-}
 
-struct PendingUpdate<R: Runtime>(Mutex<Option<UpdateResponse<R>>>);
-
-#[derive(Default)]
-pub struct Builder {
```
lucasfernog, line 41: We shouldn't remove the Builder here.. installer_args is needed (even the integration test will fail without it IIRC) and a custom target option is nice to support custom configurations (for instance universal macOS bundles, different update channels..)

lucasfernog (changes requested):

plugins/updater/src/lib.rs
amrbashir: sure yeah, I will provide both options

plugins/updater/guest-js/index.ts
```diff
@@ -52,11 +61,12 @@ class Update {
   }
 }
 
-async function check(options?: CheckOptions): Promise<Update> {
-  return invoke<UpdateResponse>("plugin:updater|check", { ...options }).then(
-    (response) => new Update(response)
+/** Check for updates, returns `null` if no updates are available */
```
lucasfernog:
```suggestion
/** Check for updates, resolves to `null` if no updates are available */
```

plugins/updater/src/commands.rs
```diff
@@ -4,59 +4,56 @@
...
-    }
-}
-
 #[tauri::command]
 pub(crate) async fn check<R: Runtime>(
     app: AppHandle<R>,
-    pending: State<'_, PendingUpdate<R>>,
-    headers: Option<HeaderMap>,
+    pending: State<'_, PendingUpdate>,
+    headers: Option<Vec<(String, String)>>,
```
lucasfernog, line 43: You changed the headers format without changing it in the TS API.

I prefer the object way btw, it's closer to other JS libraries like `fetch`, `axios`.

plugins/updater/src/lib.rs
```diff
@@ -99,3 +141,7 @@ impl Builder {
             .build()
     }
 }
+
+pub fn init<R: Runtime>() -> TauriPlugin<R, Config> {
```
lucasfernog: So far we never exposed an init function when we have a Builder.. I feel it hides that API and could confuse the user IMO.

lucasfernog (changes requested):

plugins/updater/src/lib.rs
lucasfernog: Pushed a change to remove it.

plugins/updater/src/commands.rs
lucasfernog: Pushed a commit to revert this.

amrbashir: Yeah my bad I was working on the http in parallel and mixed them up. The JS api should definitely be an object like maybe even use [`Headers`](https://developer.mozilla.org/en-US/docs/Web/API/Headers) as fetch?

plugins/updater/src/lib.rs
amrbashir: I feel like most users won't configure the plugin and just plug it and forget about it and the init function has a nicer syntax that I don't really think we should give that up.

plugins/updater/src/commands.rs
lucasfernog: I believe it already supports both a plain object and a Headers instance (which is nice).

amrbashir: Our `headers` field is typed `Record<string, unknown>` while fetch `headers` filed is typed `[string, string][] | Record<string, string> | Headers`

lucasfernog (approved):

PR #445, opened 2023-06-20
fix(dialog): using AsyncMessageDialog for non linux, closes tauri#7182

pewsheen:
Issue https://github.com/tauri-apps/tauri/issues/7182
Although the issue requests to center the dialog on the window, I think someone might want to center the dialog on the screen.

I'm trying to follow how `FileDialog` deals with the async dialog. Non-Linux systems will use the `AsyncMessageDialog`.

Tested in macOS, Windows, and Linux. Only macOS will center the dialog.

### DIFF
--- a/.changes/dialog-async-message-dialog.md
+++ b/.changes/dialog-async-message-dialog.md
@@ -1,0 +1,5 @@
+---
+"dialog": "patch"
+---
+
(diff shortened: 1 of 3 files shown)


lucasfernog:
Looks like we actually need rfd to add an option to center the dialog on the window instead.

lucasfernog (approved):

PR #447, opened 2023-06-20
Fix README Javascript example

dukedorje:
The example in the README doesn't work. Checking the examples/ directory, the import needs to be referenced differently.


FabianLars (approved):
thank you

PR #449, opened 2023-06-20
feat(stronghold): add an init function that uses argon2

vdang-crabnebula:
Stronghold plugin asks users to provide a key derivation function for initialization.
This is both insecure and anti user-friendly. 

This PR provides:
- an `init_and_build_with_argon2` function that uses argon2 as kdf with salt generation
- `argon2` and other crates are behing a `kdf` feature gate


plugins/stronghold/src/lib.rs
```diff
@@ -405,15 +408,35 @@ impl Builder {
         }
     }
 
+    /// Initializes a stronghold plugin with argon2 as a default kdf
+    pub fn init_and_build_with_argon2<R: Runtime>() -> TauriPlugin<R> {
```
lucasfernog: to reduce duplicated code, I'd actually name this as `pub fn with_argon2` that calls the `new` function with the appropriate password hash function.

lucasfernog (changes requested):

lucasfernog (approved):

PR #459, opened 2023-06-30
fix(persisted-scope): Save asset protocol patterns

FabianLars:
https://discord.com/channels/616186924390023171/1124318735809904710


FabianLars:
actually, we're never reading these paths so this pr is still useless, gonna update it...

amrbashir (approved):

PR #468, opened 2023-07-06
Merge v1 into v2

FabianLars:


FabianLars:
let's wait for the other persisted-scope pr tho

Edit: Converting to a draft just be safe since the persisted-scope plugin is broken in this pr

lucasfernog (approved):

PR #473, opened 2023-07-10
ci: Switch to dependabot

FabianLars:
tldr: Renovate is broken, dependabot works, so why not just switch 🤷 
I'll leave this PR open for a while so that everyone has a chance to raise their concerns.

for context my question about "Why renovate?" on discord: https://discord.com/channels/616186924390023171/631158878108909588/1127989246372876341


amrbashir:
since renovate doesn't yet fully support pnpm, we either switch to dependabot or switch to yarn. Now I don't actively maintain this repo so I will keep the decision for those who do.

FabianLars:
Soooo, as it turns out it was my bad again. I did not notice that this repo uses `^` in the versions (though tbf renovate did not open a pr to lock them like it usually does iirc). So i'll open a PR for that which should make renovate mostly work again.

buuuut i'll leave this PR open because i still think switching may not be a bad idea.
Edit: Hmmm, since it's semi-working now i think this should probably be an org-wide decision and idk if this PR is the best place to discuss that.

p.s. hard no for the yarn option, i'd much rather update the deps all by myself or go with plain npm 🤷

FabianLars:
no review is also a kind of review so let's close this then

PR #477, opened 2023-07-11
chore: Lock npm deps versions

FabianLars:
I did keep the example deps unlocked because renovate doesn't update the example deps iirc


FabianLars:
nvm, locked the example deps too

wusyong (approved):

PR #501, opened 2023-07-16
MySQL examples, v2

writeDavid:
Okay. I think I have everything worked out. Thank you for your patience with my trivial fumblings.


FabianLars (approved):
big thanks for sticking to it :D

Issue #494, opened 2023-07-12
[single-instance] Allow multiple instances for semver incompatible versions


matthme:
At least in my use case, versions that are incompatible according to SemVer are supposed to be able to run alongside each other, while versions that are compatible according to SemVer shouldn't. I *think* this logic makes sense in general, beyond my use case.

As I understand, on macOS there is a "single-instance-policy" in general, so I'm not sure whether it's possible to still allow
running multiple, SemVer-incompatible instances alongside each other by modyfing this plugin.

matthme:
I created a draft modification of the single-instance plugin [here](https://github.com/tauri-apps/plugins-workspace/compare/v1...matthme:plugins-workspace:single-instance-semver) (tested on Linux but not on Windows). I'm happy to make a PR if this looks reasonable and there is general agreement that the semver logic is desired. Otherwise one may add it behind a boolean flag or similar to have it be optional.

FabianLars:
imo this could be a nice addition, though it should be behind a non-default feature flag.

cc @amrbashir

amrbashir:
yeah sure, let's see a PR, I don't mind it to be the default either.

PR #502, opened 2023-07-17
Adding semver compatibility check for Windows and Linux

matthme:
Addressing #494 . This PR is assuming that this is desired default behavior. If it should rather be hidden behind a feature flag, I can try to add that.


amrbashir (dismissed):
The implementation looks alright, but I am not sure I like the compatibility thing though, I think mismatched versions, even if its a patch version shouldn't launch a different version.

However, I think this maybe a desired behavior for some people so I'd suggest:
1. Add a behavior enum
	```rs
	enum LaunchBehavior {
		Same,
		#[cfg(feature = "compatible")]
		Compatible
	}
	```
2. Add a builder for the plugin that will have a method to specify which behavior they need
3. make the behavior that requires `semver` behind a feature flag just to not add unnecessary dependencies for default behavior.

does that sound alright?

matthme:
Okay, I will try to add that when I have some time.

qlaffont:
Any update on this one ?

matthme:
I did not have time to implement the feature flag as it will probably take quite some time and I had lots of other things on my plate so far. For now I'm using my fork instead for my projects: https://github.com/matthme/plugins-workspace/tree/single-instance-semver maybe that's useful for you as well.

FabianLars:
@amrbashir would it be alright to have the feature flag itself change the behavior?

amrbashir:
I guess so yeah

FabianLars:
coolio. mainly asking cause i'm lazy 🙃

FabianLars (approved):

PR #504, opened 2023-07-17
feat: Add deep link plugin for mobile

FabianLars:
~~**This is not meant to be merged!** It's only to discuss the changes presented here, changes that are still TODO / need to be discussed, and to track what changes are required outside of the plugin (in tauri's bundler f.e.)~~ nvm i guess

The biggest open question is how to handle the required Manifest change. I think ideally there'd be a way to reference the consuming app inside the plugin for example via a handlebar thingy that the cli replaces if that's possible. If it isn't then i think that leaves us with 2 options:
1) The user has to modify the manifest in their gen folder
2) the cli modifies that same manifest (or some other one?) at build time, or creates a new one somewhere and merge it into it

2) is basically like 0) with less plugin-specific which may help for other plugins in the future
-> after a small convo with lucas this sounds pretty difficult

todo list (don't have to be resolved, but talked about at least):
- [x] Android Manifest changes
- [x] Document how to set up assetlinks.json (short)
- [ ] Explore "stupid" deep links on Android for easier testing without having to host assetlinks.json with a debug key. -> The current idea is to register the desktop schemes on mobile in development.
- [x] After that we can hopefully not depend on my website (not that i actually care about that)
- [x] Replace the `window.__TAURI_INVOKE__` usage with `listen` in the next alpha release (marked as `TODO` in the code)
- [ ] Consider splitting up the initial intent and last intent (marked as `TODO` in the code)
- [x] Remove the other event `TODO`s i forgot in there - waiting for Lucas' changes to not cause any conflicts.
- [x] Obviously the iOS part is missing (Lucas is working on it)
- [x] Switch branch back to lucas' fork -> probably part of the incoming iOS changes
- [ ] With the iOS part we'll also need to figure out an API that makes sense on both platform (we mostly already know how it works on desktop).
- [ ] Get some feedback on the api and see what issues users have consuming them
- [ ] The 3 desktop platforms should not be added in this PR imo
- [x] Ask the tauri-plugin-deep-link maintainer if they want to transfer the crate name to us 😂


lucasfernog:
I see we're duplicating the config for build vs runtime. We can reuse the same config in an ugly way using `#[path = "src/config.rs"] mod config` in the build.rs file :|

FabianLars:
yeah, i saw that in some Stackoverflow answer but wasn't sure if we're comfortable with that. Or well, could be worse i guess, so why not

plugins/deep-link/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+version = "2.0.0-alpha.0"
+description = "Set your Tauri application as the default handler for an URL"
+#authors = { workspace = true }
+#license = { workspace = true }
+#edition = { workspace = true }
+#rust-version = { workspace = true }
+edition = "2021"
+authors = [ "Tauri Programme within The Commons Conservancy" ]
+license = "Apache-2.0 OR MIT"
+rust-version = "1.65"
```
FabianLars:
```suggestion
authors = { workspace = true }
license = { workspace = true }
edition = { workspace = true }
rust-version = { workspace = true }
```

plugins/deep-link/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+serde_json = "1"
+tauri-build = "2.0.0-alpha.8"
+
+[dependencies]
+serde = "1"
+serde_json = "1"
+tauri = "2.0.0-alpha.11"
+log = "0.4"
+thiserror = "1"
+url = "2"
```
FabianLars:
```suggestion
[build-dependencies]
serde = { workspace = true }
serde_json = { workspace = true }
tauri-build = { workspace = true }

[dependencies]
serde = { workspace = true }
serde_json = { workspace = true }
tauri = { workspace = true }
log = { workspace = true }
thiserror = { workspace = true }
url = "2"
```

FabianLars:
marked as ready for review so we can get it merged & released asap as discussed on discord. There are still a few things we/i should/want to look into but this PR is already large enough 👍

lorenzolewis:
Related to https://github.com/tauri-apps/plugins-workspace/pull/536#discussion_r1319681449, I've opened an issue in https://github.com/tauri-apps/governance-and-guidance/issues/42 to discuss this further, pending a non-CrabNebula board member to weigh in on this. CC @Beanow

lucasfernog (approved):

PR #523, opened 2023-07-28
Fix/readme banners

lorenzolewis:
Updates the banners on the readme's to absolute paths so that downstream docs CI has nice banners

### DIFF
--- a/plugins/app/README.md
+++ b/plugins/app/README.md
@@ -1,4 +1,4 @@
-![plugin-app](banner.png)
+![plugin-app](https://github.com/tauri-apps/plugins-workspace/raw/v2/plugins/app/banner.png)
 
 This plugin provides APIs to read application metadata and macOS app visibility functions.
 
--- a/plugins/app/README.md
+++ b/plugins/app/README.md
@@ -1,4 +1,4 @@
-![plugin-app](banner.png)
+![plugin-app](https://github.com/tauri-apps/plugins-workspace/raw/v2/plugins/app/banner.png)
 
 This plugin provides APIs to read application metadata and macOS app visibility functions.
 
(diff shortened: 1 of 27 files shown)


FabianLars (approved):

amrbashir:
I feel like I prefer the relative path, which makes it resilient against branch name changing, merge conflicts, parent directory renaming ...etc. I feel like this is a problem that should be addressed downstream where the script that generates the docs for the README, should rewrite the relative URL to a permanent URL that reference the commit it was generated from.

lorenzolewis:
How do other downstreams like docs.rs and npm handle this? I can look into what it would take to match whatever they're doing. I'd just copied what's done here: https://github.com/tauri-apps/tauri/blob/4db363a03c182349f8491f46ced258d84723b11f/core/tauri/README.md?plain=1#L3 (and in other packages there)

FabianLars:
tbh I don't think any of the reasons listed really matter, at least in the short term, so imo we should focus on getting the docs running first and add this improvement to the Todo list for later :)

amrbashir:
> How do other downstreams like docs.rs and npm handle this? I can look into what it would take to match whatever they're doing.

`Crates.io` seem to transform the relative link into an absolute URL, so `./splash.png` will become `https://raw.githubusercontent.com/tauri-apps/plugins-workspace/e47562f71284457ff77e4c8b6bf02fdbe19ab880/plugins/window-state/splash.png`

>  I'd just copied what's done here: [tauri-apps/tauri@`4db363a`/core/tauri/README.md?plain=1#L3](https://github.com/tauri-apps/tauri/blob/4db363a03c182349f8491f46ced258d84723b11f/core/tauri/README.md?plain=1#L3) (and in other packages there)

Yeah, we probably gonna need to fix those as well.

> tbh I don't think any of the reasons listed really matter, at least in the short term, so imo we should focus on getting the docs running first and add this improvement to the Todo list for later :)

I agree, I am not saying it is a blocker, just something to think about for the future.

Issue #507, opened 2023-07-18
[window-state] add group function


a1156883061:
I want to save window state by group.
For example, the main window save to main group, and other window save to child group.
I expected such api: 
`add_group("main", "main")`
`add_group("child", "child")`
first param is group name, and second is match rule.
while window label contain `main`, save window state to main group, and while window label contain `child`, save window state to child group.
I think it will be very helpful if we can save state by group

amrbashir:
Could you give more info about your use-case? I don't see how groups can be helpful.

a1156883061:
For Example, a comic reader app, it loads comic list in main window, and open comic in reader window, and there could be multiple reader window.
I expect the main window have a window state, and other reader window have another window state. It will be troublesome if user need resize reader window when open a new reader window.
Or you may only have a main group, which every window is main group and have a same window state. 
***
For first use case
`
add_group("main", "main")
.add_group("reader", "reader")
`
First add group in builder, then open new reader window with label start with `reader`(for example: `reader-1`, `reader-2`), the reader can have same window state which I expected.
***
For second use case,
`add_group("main", "main")`, and open new window with label start with `main`, this way, every window could be same state.
***
The match rule maybe more complicated, supported regex rule.

PR #531, opened 2023-08-03
feat(window-state): add group function

a1156883061:
https://github.com/tauri-apps/plugins-workspace/issues/507
Add group function.
If developer add a group, while label match the rule, it save window state to group name, else save to label name like previous logic.

### DIFF
--- a/plugins/window-state/src/lib.rs
+++ b/plugins/window-state/src/lib.rs
@@ -134,7 +134,7 @@ impl<R: Runtime> WindowExt for Window<R> {
 
         let mut should_show = true;
 
-        if let Some(state) = c.get(key) {
+        if let Some(state) = c.get(&key) {
             // avoid restoring the default zeroed state
             if *state == WindowState::default() {
                 return Ok(());
--- a/plugins/window-state/src/lib.rs
+++ b/plugins/window-state/src/lib.rs
@@ -4,7 +4,11 @@
 
 use bitflags::bitflags;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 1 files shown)


amrbashir (changes requested):
Thanks for your patience but I am quite busy with something else atm and can't focus on reviewing PRs right now but I took a quick look and I don't think we should go with the approach you proposed, instead I think we could maybe use a converter function, for example:
```rs
	.transform_label(|label| {
		if label.startsWith("reader") {
			"reader".to_string()
		} else {
			label
		}	
	})
	.build()
```
but I will cc @lucasfernog, maybe he has a better idea

lucasfernog:
I like @amrbashir idea, it's easier to implement and use.

a1156883061:
can you give me more specific reason why add_group is not a good approach. I think it more intuitive than transform_label.
with add_group
```
	.add_group(tauri_plugin_window_state::Builder::default()
            .add_group(Group::new("reader".into(), |label| label.startsWith("reader")))
            .build())
	.build()
```
I think this API is easy enough to understand and use.

amrbashir:
Your implementation is more complex, requires more allocations and usage of more mutexes while the proposed solution could just be a simple mapping function that probably will use zero-allocations.

I admit that `transform_label` is not very expressive and could use a better name though so I'd be happy to change that to a better name

FabianLars:
I'm in the process of fixing the open v1 PRs according to provided feedback myself to finally be able to change the default branch to v2. I have troubles with this PR though, i believe we need to hear about more users' use cases to come up with a good PR because for me #507 is not enough. I just don't see the value here, or rather, how this actually solves problems.

Therefore i am closing this PR to unblock the v1 to v2 migration. Thank you for contributing nonetheless!

PR #536, opened 2023-08-07
feat: add barcode scanner plugin

lucasfernog:


examples/api/src-tauri/gen/android/buildSrc/src/main/java/com/tauri/api/kotlin/BuildTask.kt
```diff
@@ -32,7 +32,7 @@ open class BuildTask : DefaultTask() {
         val rootDirRel = rootDirRel ?: throw GradleException("rootDirRel cannot be null")
         val target = target ?: throw GradleException("target cannot be null")
         val release = release ?: throw GradleException("release cannot be null")
-        val args = listOf("tauri", "android", "android-studio-script");
+        val args = listOf("/Users/lucas/projects/tauri/plugins-workspace/examples/api/./node_modules/.bin/../../../../node_modules/.pnpm/@tauri-apps+cli@2.0.0-alpha.12/node_modules/@tauri-apps/cli/tauri.js", "android", "android-studio-script");
```
amrbashir: local path

plugins/barcode-scanner/sponsors/crabnebula.svg
amrbashir, line 1: is this okay to add? cc @Beanow

amrbashir (dismissed):
I can't say much about the code since I am not very familiar with android APIs but I tested it and it works fine so LGTM! and I will trust you on the iOS bits since I can't test it.

plugins/barcode-scanner/sponsors/crabnebula.svg
Beanow: Thanks for the callout. Let me reply on the readme.

plugins/barcode-scanner/README.md
```diff
@@ -0,0 +1,98 @@
...
+        </a>
+      </td>
+      <td align="center" valign="middle">
+        <a href="https://impierce.com/" target="_blank">
+            <img src="sponsors/impierce.png" alt="Impierce" width="283" height="90">
+        </a>
+      </td>
+    </tr>
+  </tbody>
+</table>
```
Beanow: (cc @amrbashir @tauri-apps/tauri-board @lorenzolewis)

In the interest of not blocking useful functionality based on a sponsors policy discussion, I'm inclined to have the discussion separately. Would appreciate a gut-check from a non-CN Board Director to be sure.

But I have one immediate check I'd like to do.
@lucasfernog as PR author can you confirm that: both CrabNebula and Impierce are aware and OK with being credited?

And to be aware Tauri was pretty much caught off guard here so there's no hard commitments on whether, how and where these credits / logos might go after a followup discussion.

For further discussion I'd like to track in https://github.com/tauri-apps/governance-and-guidance since my main devils advocate question here would be: what would a different company have to do to get the same treatment? Of getting their plugin adopted by `@tauri-apps` like an official plugin, with sponsor callouts and everything. And who is responsible for future maintenance, etc.

lorenzolewis: Thanks @Beanow , I've opened an issue in https://github.com/tauri-apps/governance-and-guidance/issues/42 to discuss this further, pending a non-CrabNebula board member to weigh in on this.

JelleMillenaar: @Beanow Impierce is aware that we are credited but wouldn't complain at all if it is removed. Our goal is not too show up in the credits, but rather sponsor this development without any particular conditions ;)

JelleMillenaar:
Impierce has tested the plug-in on a range of Android and IOS devices and can confirm great and reliable performance. All our feedback and concerns have been tackled so we have no further feedback to provide. In the next weeks we will be rolling out a much wider public test of our mobile application and will report back any issues if they are encountered. 

As said in the comment above, we are happy to be mentioned as sponsor but this is non-conditional for us, so we wouldn't complain if Tauri rather removes it or places it on a different location.

amrbashir (approved):

plugins/barcode-scanner/README.md
Beanow: @JelleMillenaar that is awesome! And thank you for coming in to clarify :100:

PR #564, opened 2023-08-15
feat: Migrate v1 deep-link plugin to workspace repo

FabianLars:
This probably doesn't need to be merged anytime soon and is really only being migrated because i gave the tauri-plugin-deep-link name to the tauri bot account.

Anyway, the only really required change before this could be merged is the readme and banner. An actual tauri app as an example instead of a main.rs file may be cool too


FabianLars:
SERIOUSLY A MSRV ISSUE HERE TOO????????

Edit: Don't take this srsly, don't wanna get cancelled for having fun :)

FabianLars:
Unbenannt

PR #583, opened 2023-08-23
fix: tuple to struct variant

simonhyll:
`FileDropEvent` is now a struct variant instead of a tuple variant.


amrbashir (changes requested):
you forgot the JS part in the window plugin :wink:

lucasfernog:
We fixed it on another commit ;) thanks!

Issue #591, opened 2023-09-05
[updater] v2 - updater fails to install an update on macos


pashokitsme:
MacOS
Download works fine, but fails on install
When updater tries to install update, it iterates through entries in archive and skip every where it can't get path. But, since archive is compressed, updater skips all entries
It doesn't return any error or something and corrupts .app

Binlogo:
@pashokitsme It seems that you're working on fixing this issue. Are you going to create a PR to proceed it?

pashokitsme:
@Binlogo yep, i fixed it for me. And i already created PR #592 but seems everyone (including me) forgot about it

Binlogo:
> @Binlogo yep, i fixed it for me. And i already created PR #592 but seems everyone (including me) forgot about it

Great work, I tested your fix, and it worked for me. Let's merge pull request #592 to fix this issue. cc @amrbashir @lucasfernog 😉

PR #592, opened 2023-09-05
fix(updater): installation on macos #591

pashokitsme:
Using flate2 to decompressing tar.gz


plugins/updater/Cargo.toml
```diff
@@ -32,6 +32,10 @@ tar = "0.4"
 [target."cfg(target_os = \"windows\")".dependencies]
 zip = { version = "0.6", default-features = false }
 
+[target."cfg(target_os = \"macos\")".dependencies]
+flate2 = "1.0.27"
```
amrbashir, line 36: the `zip` crate has `flate2` feature, could we use that instead of adding a new dependencies entry?

amrbashir (changes requested):
cc @lucasfernog since I can't test this

plugins/updater/Cargo.toml
pashokitsme: Sure
I didn't notice zip has such feature

FabianLars:
Since this PR was just mentioned in the other issue? Isn't it enough to literally just enable the feature in the zip crate? at a glance it looks like that's the only difference between the code in the plugin and the one in tauri v1 🤔

pashokitsme:
@FabianLars 
Updater for macos uses `.tar.gz` archives, so they unpacks with crate `tar` which only reads archives, not decompresses. In Tauri v1 [updater](https://github.com/tauri-apps/tauri/blob/1.3/core/tauri/src/updater/core.rs#L846) uses [ArchiveFormat & ArchiveReader](https://github.com/tauri-apps/tauri/blob/1.3/core/tauri/src/api/file/extract.rs#L18) which handles zip and tar archives and requires `flate2` under the hood

FabianLars:
Ah lol right, I forgot about our custom implementation, sorry!

lucasfernog (approved):

PR #596, opened 2023-09-08
fix(notification): function 'invoke' not found in permission request 

LetrixZ:
Use `__TAURI_INVOKE__` instead of `__TAURI__.invoke`


amrbashir (approved):

amrbashir:
thank you

PR #598, opened 2023-09-11
fix(updater): installation on macos

olexiyb:
I know there is a pull request https://github.com/tauri-apps/plugins-workspace/pull/592, but I think my change is less code


lucasfernog:
I'm merging #592 because it keeps the current code that has already been audited, but I'll ask for input from the security team on this change. It reduces a lot of the code, but it also escapes the extraction path with `.join("..")` which might be a vulnerability.

PR #607, opened 2023-09-19
fix(sql): decode float columns as f32, fixes #602

### LINKED ISSUE #602: [sql] v1 - mysql | Can't SELECT float value
This is my table:
image
This is my query that works (so INT, DATETIME, VARCHAR[200], ENUM work):
**"SELECT id, clid as name, src, calldate, dcontext, uniqueid, dstchannel, channel, disposition, dst FROM cdr WHERE calldate >= DATE_SUB(NOW(), INTERVAL 4 WEEK) ORDER BY id DESC"**

But as soon as I select a FLOAT field (duration, billsec, userfield) i get this error:
**thread 'tokio-runtime-worker' panicked at 'range end index 8 out of range for slice of length 4', C:\Users\benit\.cargo\registry\src\index.crates.io-6f17d22bba15001f\byteorder-1.4.3\src\lib.rs:2199:28**

even if i try cast it to INT like so: **CAST(billsec AS INT) AS duration**

All these queries **work in HeidiSQL**

FabianLars:


FabianLars:
Community member confirmed it fixes it: https://discord.com/channels/616186924390023171/1153666777167962142/1153679598811349003

amrbashir (approved):

PR #609, opened 2023-09-21
add support for absolute sqlite db path in case of using tauri dialog…

syokensyo:
Since tauri provide dialog api, we can use it to browse files on the computer and get the file's absolute path, it is more flexible to make the sqlite plugin adapt to that, so user can operate on any sqlite db file in their computer, no need to put db file to the app's path anymore.

### DIFF
--- a/plugins/sql/src/plugin.rs
+++ b/plugins/sql/src/plugin.rs
@@ -73,6 +73,18 @@ fn app_path<R: Runtime>(app: &AppHandle<R>) -> PathBuf {
 /// Maps the user supplied DB connection string to a connection string
 /// with a fully qualified file path to the App's designed "app_path"
 fn path_mapper(mut app_path: PathBuf, connection_string: &str) -> String {
+    if connection_string.contains(std::path::MAIN_SEPARATOR_STR) {
+        let file_path = std::path::Path::new(connection_string);
+        if file_path.exists() {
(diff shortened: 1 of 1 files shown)


FabianLars:
Thanks for contributing! Before anything else, to be able to merge this PR your commit must be [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) :)

Now about the change itself: i also worked on this half a year ago here: https://github.com/tauri-apps/plugins-workspace/pull/279.
I'm afraid that the implementation can't be as simple as in your PR since we have to support both `/` and `\` on all OSes, have to handle the stupid UAC path on Windows (the `\\?\` path prefix) and may or may not have to handle `sqlite:` and `sqlite://` at the same time since both are valid - in my PR i just added comments that only `sqlite://` is supported but not sure if i'm 100% happy with this decision.

syokensyo:
Oh, sorry, didn't notice your PR, and yes you are right, I didn't cover enough scenarios, will look into your pr, and I think absolute path support is a frequency requirement for sqlite user, it's better to support limited scenarios than no scenario.

amrbashir:
you could use `dunce` crate which will remove the UNC on Windows and keep the path as is on other platforms.

syokensyo:
Will try it, thanks.

Issue #122, opened 2023-02-06
[log] Custom .log file name for tauri-plugin-log


Aviv-Frenkel:
I would like to be able to manually set the `.log` file name, I digged a bit in the lib.rs file of this plugin and noticed that `appname` is passed when calling the `get_log_file_path` function as the file name.
I want the log file name to be something like `rust-side.log` and not be set by the app name, maybe through passing it as an argument to the builder function.

PR #633, opened 2023-10-04
feat (log): Custom .log file names

mellobacon:
Addresses #122 and added appropriate docs. 

Probably should have error checking for the file name but just checking if the implementation is good for now.

### DIFF
--- a/plugins/log/src/lib.rs
+++ b/plugins/log/src/lib.rs
@@ -167,6 +167,7 @@ pub struct Builder {
     timezone_strategy: TimezoneStrategy,
     max_file_size: u128,
     targets: Vec<LogTarget>,
+    log_name: String
 }
 
 impl Default for Builder {
@@ -183,12 +184,14 @@ impl Default for Builder {
                 message
             ))
         });
+        let log_name = "".to_owned();
         Self {
             dispatch,
             rotation_strategy: DEFAULT_ROTATION_STRATEGY,
             timezone_strategy: DEFAULT_TIMEZONE_STRATEGY,
             max_file_size: DEFAULT_MAX_FILE_SIZE,
             targets: DEFAULT_LOG_TARGETS.into(),
+            log_name
         }
     }
 }
@@ -262,6 +265,29 @@ impl Builder {
         self
     }
 
+    /// Writes logs to the given file. Default: <app_name>.log)
+    /// 
+    /// Note: This does not modify the directory logs go into. For that refer to `LogTarget::Folder`.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use tauri_plugin_log::Builder;
+    /// let name = "custom-name";
+    /// let builder = Builder::default()
+    ///     .targets([
+    ///         LogTarget::LogDir
+    ///     ])
+    ///     .log_name(name)
+    ///     .build()
+    /// ); // Outputs content to custom-name.log 
+    ///
+    /// ```
+    pub fn log_name(mut self, log_name: &str) -> Self {
+        self.log_name = log_name.to_owned();
+        self
+    }
+
     #[cfg(feature = "colored")]
     pub fn with_colors(self, colors: fern::colors::ColoredLevelConfig) -> Self {
         let format =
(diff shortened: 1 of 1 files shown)


FabianLars:
> Probably should have error checking for the file name but just checking if the implementation is good for now.

I think it makes sense to rename the app_name var to log_name to make the intention more clear. Maybe also merge the if into the let declaration so it doesn't even request the app name when there's a custom name, also would get rid of the mutability.

It looks good otherwise. And i'm not sure if we even have to validate it since that's something we generally don't really do with paths/file names 🤔

FabianLars:
Oh and thanks for contributing of course!!

mellobacon:
> I think it makes sense to rename the app_name var to log_name to make the intention more clear. Maybe also merge the if into the let declaration so it doesn't even request the app name when there's a custom name, also would get rid of the mutability.

Will do!
> Oh and thanks for contributing of course!!

And no problem :)

FabianLars:
Ah one last thing i overlooked (besides the format check failing) is to use an [Option](https://doc.rust-lang.org/std/option/index.html) instead of an empty string.
For what it's worth, this would also enable users to name their log file just `.log` if they want to.

mellobacon:
Let me know if there's anything else I missed 😅

plugins/log/src/lib.rs
```diff
@@ -183,12 +184,14 @@ impl Default for Builder {
...
             ))
         });
+        let log_name = None;
         Self {
             dispatch,
             rotation_strategy: DEFAULT_ROTATION_STRATEGY,
             timezone_strategy: DEFAULT_TIMEZONE_STRATEGY,
             max_file_size: DEFAULT_MAX_FILE_SIZE,
             targets: DEFAULT_LOG_TARGETS.into(),
+            log_name,
```
FabianLars:
```suggestion
        Self {
            dispatch,
            rotation_strategy: DEFAULT_ROTATION_STRATEGY,
            timezone_strategy: DEFAULT_TIMEZONE_STRATEGY,
            max_file_size: DEFAULT_MAX_FILE_SIZE,
            targets: DEFAULT_LOG_TARGETS.into(),
            log_name: None,
```

plugins/log/src/lib.rs
```diff
@@ -262,6 +265,29 @@ impl Builder {
...
+    /// let builder = Builder::default()
+    ///     .targets([
+    ///         LogTarget::LogDir
+    ///     ])
+    ///     .log_name(name)
+    ///     .build()
+    /// ); // Outputs content to custom-name.log
+    ///
+    /// ```
+    pub fn log_name(mut self, log_name: &str) -> Self {
```
FabianLars:
```suggestion
    pub fn log_name<S: Into<String>>(mut self, log_name: S) -> Self {
```

plugins/log/src/lib.rs
```diff
@@ -284,7 +310,11 @@ impl Builder {
         plugin::Builder::new("log")
             .invoke_handler(tauri::generate_handler![log])
             .setup(move |app_handle| {
-                let app_name = &app_handle.package_info().name;
+                let name = self
+                    .log_name
+                    .clone()
+                    .unwrap_or_else(|| app_handle.package_info().name.clone());
+                let log_name = name.as_str();
```
FabianLars:
```suggestion
                let log_name = self
                    .log_name
                    .as_deref()
                    .unwrap_or_else(|| &app_handle.package_info().name);
```

FabianLars (changes requested):
Just some small things left because i'm so picky lol (sorry) - meant to push this directly but didn't have access 🤷

plugins/log/src/lib.rs
```diff
@@ -262,6 +264,29 @@ impl Builder {
...
+    ///     .targets([
+    ///         LogTarget::LogDir
+    ///     ])
+    ///     .log_name(name)
+    ///     .build()
+    /// ); // Outputs content to custom-name.log
+    ///
+    /// ```
+    pub fn log_name<S: Into<String>>(mut self, log_name: S) -> Self {
+        self.log_name = Some(log_name.to_string());
```
FabianLars:
```suggestion
        self.log_name = Some(log_name.into());
```

FabianLars (approved):
thank youuu

PR #642, opened 2023-10-08
Fix incomplete overwriting behaviour in fs.writeFile.

zh217:
This contains the same fix as https://github.com/tauri-apps/tauri/pull/7982 which fixes https://github.com/tauri-apps/tauri/issues/7973 But as v2 moves things around a little bit the fix is necessary here as well.

### DIFF
--- a/plugins/fs/src/commands.rs
+++ b/plugins/fs/src/commands.rs
@@ -117,9 +117,9 @@ pub fn write_file<R: Runtime>(
         .with_context(|| format!("path: {}", resolved_path.display()))
         .map_err(Into::into)
         .and_then(|mut f| {
-            f.write_all(&contents)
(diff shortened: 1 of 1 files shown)


FabianLars:
_copy&paste from your other PR on the tauri repo:_

Hey, thanks for contributing! 2 things before we can merge this:
1) I think this can be simplified by just using [`truncate()`](https://doc.rust-lang.org/std/fs/struct.OpenOptions.html#method.truncate)
2) We require all commits to be signed, here's a guide on how to set it up: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

zh217:
Never mind this one, as the v2 code uses File::create which doesn't have the bug.

PR #662, opened 2023-10-16
fix(updater): Fix endpoint string replacements, fixes #655

### LINKED ISSUE #655: [updater] the config endpoints was not replaced by target, arch and version
```
            // replace {{current_version}}, {{target}} and {{arch}} in the provided URL
            // this is useful if we need to query example
            // https://releases.myapp.com/update/{{target}}/{{arch}}/{{current_version}}
            // will be translated into ->
            // https://releases.myapp.com/update/darwin/aarch64/1.0.0
            // The main objective is if the update URL is defined via the Cargo.toml
            // the URL will be generated dynamically
            let url: Url = url
                .to_string()
                .replace("{{current_version}}", &self.current_version.to_string())
                .replace("{{target}}", &self.target)
                .replace("{{arch}}", self.arch)
                .parse()?;
```
in the updater.rs
the url format {{current_version}} as %7B%7Bcurrent_version%7D%7D, replacement is failed

FabianLars:
i know it's ugly and i'm open for discussion.
One thing i really wanted to avoid though was decoding the url just for the replacement, to be 100% sure not to mess up whatever url the user provided.
So the only 2 solutions i could think of were this, or reverting it to a simple String again but i wanted to keep the Url validation in the Deserialization impl but that would have made error handling ugly at first glance.

Maybe i'm just too tired and overthinking this while missing the obvious...

### DIFF
--- a/.changes/updater-string-replace.md
+++ b/.changes/updater-string-replace.md
@@ -1,0 +1,5 @@
+---
+"updater": "patch"
+---
+
+The plugin now correctly replaces `arch`, `current_version` and `target` again.
--- a/plugins/updater/src/updater.rs
+++ b/plugins/updater/src/updater.rs
@@ -259,7 +259,10 @@ impl Updater {
             let url: Url = url
                 .to_string()
                 // url::Url automatically url-encodes the string
-                .replace("%7B%7Bcurrent_version%7D%7D", &self.current_version.to_string())
+                .replace(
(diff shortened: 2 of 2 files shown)


amrbashir (dismissed):

FabianLars:
for commits like this i really hate the auto dismissal...

amrbashir (approved):

PR #673, opened 2023-10-19
chore: update to tauri alpha.16, api alpha.9

lucasfernog:


plugins/authenticator/guest-js/index.ts
```diff
@@ -2,23 +2,33 @@
...
+    };
   }
 }
 
 export class Authenticator {
   async init(): Promise<void> {
-    return await window.__TAURI_INVOKE__("plugin:authenticator|init_auth");
+    return await window.__TAURI_INTERNALS__.invoke(
+      "plugin:authenticator|init_auth",
+    );
```
amrbashir: We shouldn't really on `window.__TAURI_INTERNALS__` anymore, it is internal and should be treated as such in this repo, we should use `invoke`directly from `@tauri-apps/api/primitives` which provides the abstraction around the internals.

amrbashir (changes requested):

plugins/authenticator/guest-js/index.ts
lucasfernog: we do this for the plugins because otherwise each plugin global JS script would consume an extra 1.5-2KB on the binary :(

lucasfernog: actually I'm not seeing a huge change anymore... weird

lucasfernog: Maybe your cleanup is saving us :)

lucasfernog: pushed, but on the notification init.js script I can't really do anything about it :(

amrbashir: we could turn that script into a ts file that we compile with rollup

lucasfernog: the things we do for love

FabianLars (dismissed):

amrbashir (approved):

amrbashir (approved):

PR #675, opened 2023-10-21
Fix invoke calls in dialog & shell init scripts

morajabi:
This fixes an issue that broke links and dialogs in latest commit. Let me know if I should do it in another way.


amrbashir:
Looks like we missed this one in https://github.com/tauri-apps/plugins-workspace/pull/673, these files should now be converted to a typescript file and compiled using rollup, similar to the `notification` plugins

morajabi:
For people who have this issue, here's a workaround without patching until this is fixed. Add this line in your entry JS/TS file:
```
 window.invoke = window.__TAURI_INTERNALS__.invoke;
```

morajabi:
@amrbashir Feel free to close this PR if this isn't the way you want to handle it.

lucasfernog (approved):

PR #676, opened 2023-10-21
feat: update to alpha.17, typed mobile plugin IPC arguments

lucasfernog:
Updates to the in progress typed IPC arguments that will be released on alpha.17.

Needs https://github.com/tauri-apps/tauri/pull/8076


plugins/notification/guest-js/index.ts
```diff
@@ -150,57 +150,62 @@ type ScheduleInterval = {
...
-  static at(date: Date, repeating = false) {
-    return new Schedule({ kind: "At", data: { date, repeating } });
+  static at(date: Date, repeating = false, allowWhileIdle = false) {
+    return new Schedule({ at: { date, repeating, allowWhileIdle } });
   }
 
-  static interval(interval: ScheduleInterval) {
-    return new Schedule({ kind: "Interval", data: interval });
+  static interval(interval: ScheduleInterval, allowWhileIdle = false) {
+    return new Schedule({ interval: { interval: interval, allowWhileIdle } });
```
amrbashir:
```suggestion
    return new Schedule({ interval: { interval, allowWhileIdle } });
```

plugins/shell/src/process/mod.rs
```diff
@@ -511,7 +511,7 @@ mod tests {
         assert_eq!(String::from_utf8(output.stdout).unwrap(), "");
         assert_eq!(
             String::from_utf8(output.stderr).unwrap(),
-            "cat: test/: Is a directory\n"
+            "cat: test/: Is a directory\n\n"
```
amrbashir, line 514: does that sound okay? why does it have two new lines?

lucasfernog: the std one only gives us a single \n so it's a bug on our end (it was already a bug, since it had a \n where it shouldn't, I guess someone just fixed the test instead of the actual bug). We should track this on the tauri repo (v1 and v2).

amrbashir: Cool

amrbashir (approved):

PR #677, opened 2023-10-22
feat: Add KeepSome to log plugin

kris-ava:
Allows users to only keep some of the old plugins (thus making sure that logging does not hog all available disk space)


FabianLars (commented):
Hi, thanks so much for contributing and sorry for the huge delay!
I'll do a proper review tomorrow but wanted to ask you for a changefile similar to https://github.com/tauri-apps/plugins-workspace/blob/v2/.changes/dialog-async-message-dialog.md already.

tance77:
@FabianLars @kris-ava  Just wanted to circle back on this. I to have a use case where i don't want to keep "everything" but simply the last X days. Also maybe a version 1 port :)

MGough:
Looks like this has been sitting a while! Is there any chance this could be backported to V1?

FabianLars:
We mostly stopped backporting features to v1, but we're still looking at community PRs that do so.

MGough:
Thanks for the speedy response @FabianLars - I may open a PR to do so in the near future if @kris-ava is not interested in doing so themselves as this feature is something I'd love to be able to use before V2 is ready for production use!

Dexus:
@FabianLars Can we please make it to get it into the next release?

FabianLars (approved):
sry for the delay again

PR #702, opened 2023-11-06
chore: Merge branch v1 into v2

FabianLars:


FabianLars:
Alrighty, should have done that more often. Lesson learned.

plugins/websocket/guest-js/index.ts
```diff
@@ -13,6 +13,15 @@ export interface ConnectionConfig {
   headers?: HeadersInit;
 }
 
+export interface ConnectionConfig {
```
lucasfernog: don't we need the change file for this change?

plugins/sql/src/decode/mysql.rs
```diff
@@ -21,7 +21,14 @@ pub(crate) fn to_json(v: MySqlValueRef) -> Result<JsonValue, Error> {
                 JsonValue::Null
             }
         }
-        "FLOAT" | "DOUBLE" => {
+        "FLOAT" => {
```
lucasfernog, line 24: same for this one... why is there a change file only for some of the changes?

FabianLars: Because in the v1 branch we only added changefiles for released packages

lucasfernog (approved):

PR #708, opened 2023-11-09
dialog: pass readData through to mobile plugin

jquesnelle:
This enables passing `readData` as an argument to `OpenDialogOptions` to enable reading selected files on Android. While the Dialog plugin code for Android correctly deals with `readData` and populates `base64Data` (which allows Android `ContentProvider` to be read from guest-js), the argument itself can't currently be populated from guest-js and used 🙂

### DIFF
--- a/plugins/dialog/src/lib.rs
+++ b/plugins/dialog/src/lib.rs
@@ -267,6 +267,8 @@ pub struct FileDialogBuilder<R: Runtime> {
     pub(crate) title: Option<String>,
     #[cfg(desktop)]
     pub(crate) parent: Option<raw_window_handle::RawWindowHandle>,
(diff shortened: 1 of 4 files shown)


plugins/dialog/src/lib.rs
```diff
@@ -341,6 +348,13 @@ impl<R: Runtime> FileDialogBuilder<R> {
         self
     }
 
+    #[cfg(mobile)]
```
FabianLars, line 351: Could you add a small doc comment here too?

lucasfernog:
I actually wanted to change that whole implementation to use the file descriptor implementation someone shared on Discord. But I didn't get time for it yet, so we can merge this one until I do so.

PR #717, opened 2023-11-12
added attachLogger

tommyo:
Adds the ability to hook into the log stream with a callback.


FabianLars (approved):
Thank you and sorry for the delay

PR #722, opened 2023-11-13
refactor: simplify rollup config and match `@tauri-apps/api` output style

amrbashir:


lucasfernog:
I think we should add https://github.com/tauri-apps/tauri/pull/8130 just in case we end up needing it in the future.

amrbashir:
`tslib` is needed in `@tauri-apps/api` because of the use of private class properties which I don't think we are gonna need here at all and I'd rather deal with it later when needed tbh since we could also decide to increase our typescript target which will not need `tslib`

lucasfernog:
The main problem is we won't know when we need that change, I'm sure we'll forget about it. And we'll have to deal with bugs in the community.
It's a cheap solution that does not impact the dist when you do not need tslib anyway.

amrbashir:
if we ever need `tslib`, rollup will error out and won't bundle until you add `tslib` as a dependency which I removed in this PR so if we ever need it, we will know.

shared/rollup.config.js
```diff
@@ -0,0 +1,88 @@
...
+    external = [/^@tauri-apps\/api/],
+    additionalConfigs = [],
+  } = options;
+
+  const pkg = JSON.parse(readFileSync(join(cwd(), "package.json"), "utf8"));
+
+  const pluginJsName = pkg.name
+    .replace("@tauri-apps/plugin-", "")
+    .replace(/-./g, (x) => x[1].toUpperCase());
+  const iifeVarName = `__TAURI_PLUGIN_${pluginJsName.toUpperCase()}__`;
```
FabianLars, line 34: Shouldn't `-` be converted to `_` for the global vars?

FabianLars:
1) The linked PR was technically not tslib specific, tslib was just what we used in core that needed it.
2) The problem is that the PR didn't fix a build error/crash. The issue was only visible after pulling the lib from the registry (locally reproducible with npm pack). So we only noticed it after users reported it _after a release_, i doubt it'd be different here tbh.

amrbashir:
> 1. The linked PR was technically not tslib specific, tslib was just what we used in core that needed it.

I know The PR is a general solution for external code that is placed in a `node_modules` directory but as far as I can see, rollup only needs `tslib` to polyfill some features, and never seen it requiring other libraries for polyfills but in case we do, rollup will error out.

> 2. The problem is that the PR didn't fix a build error/crash. The issue was only visible after pulling the lib from the registry (locally reproducible with npm pack). So we only noticed it after users reported it _after a release_, i doubt it'd be different here tbh.

That was because `tslib` was a dependency, however in this PR `tslib` was removed so if we ever need a feature that needs to be polyfilled using `tslib`, rollup will produce an error.

shared/rollup.config.js
amrbashir: I don't know about this logic, it is the same as the old one, I only added `_PLUGIN_`

amrbashir: but the second `.replace()` removes `-` and capitalizes the character after it so `@tauri-apps/plugin-global-shortcut` will be `globalShortcut` and the final iife var will be `__TAURI_PLUGIN_GLOBALSHORTCUT__`

amrbashir: We don't need to preserve the separator between `global` and `shortcut` since it is a temporary variable and ultimately is assigned to `window.__TAURI__.globalShortcut`

FabianLars:
> That was because tslib was a dependency, however in this PR tslib was removed so if we ever need a feature that needs to be polyfilled using tslib, rollup will produce an error.

Yes, it will produce an error and then _we_ have to remember that we also need to add the generator and not just the npm dependency.

But idgaf anymore so let's merge this before we have merge conflicts.

FabianLars (approved):

amrbashir:
> Yes, it will produce an error and then we have to remember that we also need to add the generator and not just the npm dependency.

Or decide to increase our typescript target to `esnext` which is why I want to leave this decision for later.

PR #742, opened 2023-11-20
fix: add back `main`, `module` and `types` field in top-level package.json

### LINKED ISSUE #738: [v2 branch] The way of exporting module seems not friendly in package.json
### Description

In my project, I use typescript(v4.7) with config `"module": "esnext", "moduleResolution": "node"`.

If I use the lasted release of the tauri plugin package(eg: @tauri-apps/plugin-log). I will get type error which says can't recognize package `@tauri-apps/plugin-log`. Then I check the package.json of the plugin, I notice below code.

https://github.com/tauri-apps/plugins-workspace/blob/5a8bbe296704f679063ef8b820d582d7f728400b/plugins/log/package.json#L9-L14

After some research, they told me should use typescript(v5) with config `"module": "esnext", "moduleResolution": "bundler"`. They are right. The plugin package now are recognizable.

But new type error appears. `tsc` now can't recognize subpath of package. Eg:

```ts
// Cannot find module '@zing/helpers/dist/tippy' or its corresponding type declarations.
import { getTippy } from '@zing/helpers/dist/tippy';

// package.json of @zing/helpers
{
  ...
  "main": "dist/index.js",
  "module": "dist/index.js",
  "types": "dist/index.d.ts",
  "files": ["dist"],
  "exports": {
    ".": "./dist/index.js",
    "./dist/": "./dist/"
  },
}
```

### Hope

Change the way of exporting module to a more friendly one. 

Here is a example. https://github.com/ProseMirror/prosemirror-model/blob/master/package.json#L5-L12

amrbashir:
closes #738

`"moduleResolution": "node"` doesn't handle the `exports` field

### DIFF
--- a/plugins/authenticator/package.json
+++ b/plugins/authenticator/package.json
@@ -7,6 +7,9 @@
     "Tauri Programme within The Commons Conservancy"
   ],
   "type": "module",
+  "types": "./dist-js/index.d.ts",
+  "main": "./dist-js/index.cjs",
+  "module": "./dist-js/index.js",
   "exports": {
     "types": "./dist-js/index.d.ts",
     "import": "./dist-js/index.js",
--- a/plugins/autostart/package.json
+++ b/plugins/autostart/package.json
@@ -6,6 +6,9 @@
     "Tauri Programme within The Commons Conservancy"
   ],
   "type": "module",
+  "types": "./dist-js/index.d.ts",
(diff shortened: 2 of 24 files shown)


FabianLars (approved):
Can't ever have nice things 🥲

amrbashir:
One day, hopefully

Issue #287, opened 2023-03-22
[single-instance] Will single-instance support MacOS in the future?


Pylogmon:

FabianLars:
Probablyyy not since apps on macos are single instance by default (unless started manually in the terminal). If you do have a use-case where you need the plugin on top of macos' default behavior we'd like to hear about it :)

Pylogmon:
As you said, I need to start it from the terminal. I have implemented the function of calling the software from the terminal when the software is running. I can get the command line parameters through this plugin, so that my software can be called through other software (such as popclip). It works well under Windows and Linux, but this plugin does not support macos。

FabianLars:
alright, understandable use-case, thanks! Still no promises that we will work on it anytime soon but i'll bring it up in a few conversations and see what happens.

p.s. for anyone interested, community contributions are always welcome!

PR #749, opened 2023-11-20
Make single instance plugin MacOS compatible

JuliaMerz:
This got referenced again in [https://github.com/tauri-apps/plugins-workspace/issues/287] so it reminded me that I'd actually built a fix for this for one of my own projects.

See the following discord conversation: https://discord.com/channels/616186924390023171/1127879371835846656/1127879371835846656

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2661,6 +2661,7 @@ dependencies = [
  "cfg-if",
  "libc",
  "memoffset 0.7.1",
+ "pin-utils",
  "static_assertions",
 ]
 
@@ -4729,6 +4730,7 @@ name = "tauri-plugin-single-instance"
 version = "0.0.0"
 dependencies = [
  "log",
+ "nix 0.26.2",
  "serde",
  "serde_json",
  "tauri",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4735,6 +4735,7 @@ dependencies = [
  "serde_json",
  "tauri",
  "thiserror",
+ "tokio",
  "windows-sys 0.52.0",
  "zbus",
 ]
@@ -5038,9 +5039,23 @@ dependencies = [
  "libc",
  "mio",
  "num_cpus",
+ "parking_lot",
  "pin-project-lite",
+ "signal-hook-registry",
  "socket2",
+ "tokio-macros",
  "windows-sys 0.48.0",
+]
+
+[[package]]
+name = "tokio-macros"
+version = "2.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "630bdcf245f78637c13ec01ffae6187cca34625e8c63150d424b59e55af2675e"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.28",
 ]
 
 [[package]]
--- a/plugins/single-instance/Cargo.toml
+++ b/plugins/single-instance/Cargo.toml
@@ -34,3 +34,5 @@ zbus = "3"
 
 [target.'cfg(target_os = "macos")'.dependencies]
 nix = "0.26.2"
+tokio = { version = "1.28.0", features = ["full"] }
+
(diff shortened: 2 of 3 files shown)


FabianLars:
I'll leave the review to Lucas but i wonder if this should be behind a feature flag? I assume the majority of tauri apps can rely on macos' single instance mechanism just fine, which imo should be preferred in that case.

(Also, if you use tokio just to spawn a task you can use `tauri::async_runtime::spawn` instead. It's basically the same.)

JuliaMerz:
Yeah there's a very specific use case here where I wanted to build a CLI that would work even when the app was running, leading to this code:

```rust
    let builder = tauri::Builder::default().setup(move |app| {
        // TODO: break this out
        //CLI Then kill
        match app.get_cli_matches() {
            Ok(matches) => {
                tokio::task::block_in_place(|| {
                    tauri::async_runtime::block_on(cli::cli_command_interpreter(
                        app.handle(),
                        matches,
                        pool.clone(),
                    ))
                });
            }
            Err(tauri::Error::FailedToExecuteApi(e)) => match e {
                tauri::api::Error::ParseCliArguments(s) => {
                    println!("{}", s);
                }
                e => {
                    error!("Other API Error: {:?}", e);
                }
            },

            Err(e) => {
                info!("No matches: {:?}", e);
            }
        };
        app.handle()
            .plugin(tauri_plugin_single_instance::init(|app, argv, cwd| {
                debug!("{}, {argv:?}, {cwd}", app.package_info().name);
                cli::main_command_response(argv, app.state());
            }));

```

Don't quite remember which but iirc the conventional mac kill mechanism kicks in either when the window launches, or if you're launching one from the shortcut and one from the command line it's possible it doesn't kick in at all.

lars-berger:
> I'll leave the review to Lucas but i wonder if this should be behind a feature flag? I assume the majority of tauri apps can rely on macos' single instance mechanism just fine, which imo should be preferred in that case.
> 
> (Also, if you use tokio just to spawn a task you can use `tauri::async_runtime::spawn` instead. It's basically the same.)

@FabianLars Since MacOS won't launch a 2nd instance if an existing one is open (when not run via CLI), I don't see a need to put this behind a feature flag. In the normal case where a 2nd instance is launched, it'll focus the existing dock app. This means the feature flag would only be for people who don't want to guarantee a single instance, in which case, why would they be using the single instance plugin? 👀 

Getting this merged would mean the plugin is consistent across all platforms (ie. a single instance is guaranteed no matter how the app is launched).

IMO the use case of communicating with an existing instance via CLI is not uncommon. It's my use case as well, and is used by apps like VSCode. Thanks for putting this together @JuliaMerz 🤟

mandaputtra:
I also have the same use case, is there any work around for now? But weirdly the macos will run 2 instances of apps

FabianLars:
Sorry for the long delay here!!

> Since MacOS won't launch a 2nd instance if an existing one is open (when not run via CLI), I don't see a need to put this behind a feature flag. In the normal case where a 2nd instance is launched, it'll focus the existing dock app. This means the feature flag would only be for people who don't want to guarantee a single instance, in which case, why would they be using the single instance plugin? 👀
>
> Getting this merged would mean the plugin is consistent across all platforms (ie. a single instance is guaranteed no matter how the app is launched).
>
> IMO the use case of communicating with an existing instance via CLI is not uncommon. It's my use case as well, and is used by apps like VSCode

I proposed a feature flag because this PR added 2 new dependencies (1 if we follow my first commend) and has quite a bit of code. If a dev doesn't want/need this for their app it'd be relatively wasteful.
Launching apps from the cli may be common, but not _that_ common, especially if an app targets non tech heavy users 🤷 

That said, when i did say this, i completely forgot that the more rust native way to handle this would be to guard the plugin import and init call with a cfg flag... (tldr: ignore that feature flag talk)

So yeah, that means that it's still just about removing tokio. :)

FabianLars:
Actually another thing, i always preferred unix domain sockets over named pipes/fifo files for ipc on linux and macos, does anybody have opinions on that matter?

It would also mean we could get rid of the nix dependency again 🤔

FabianLars:
We went with the std version instead here https://github.com/tauri-apps/plugins-workspace/pull/1035. Thank you for the PR nonetheless!!

PR #751, opened 2023-11-21
feat(fs): improved API

amrbashir:


FabianLars:
It doesn't compile with the watch feature enabled. Just a fairly small change though (the main thing is that resolve_path expects a SafePathBuf) but i won't push it to not risk conflicts if you have local commits left :)

(p.s. opened https://github.com/tauri-apps/plugins-workspace/pull/761 to test more features in ci)

The main api change looks solid to me though (though i do think node's fs api is a bit weird idk), and what i tested so far worked nicely! 🥳 (i'll do some more testing tomorrow)
I do however think that this PR should have 2 approvals (maybe i just don't trust myself enough lol).

amrbashir:
Alright, watcher flag should be fine now, and feel free to push your changes if you need, I don't have any more changes planned.

> though i do think node's fs api is a bit weird idk)

The API is inspired from Deno and Node and the main goal was to make it more familiar to those coming from node.

plugins/fs/src/commands.rs
```diff
@@ -40,360 +56,747 @@ impl Serialize for CommandError {
...
+                .truncate(false)
+                .append(false)
+                .create_new(false);
+        }
+        Some(options) => {
+            #[cfg(unix)]
+            {
+                use std::os::unix::fs::OpenOptionsExt;
+                if let Some(mode) = options.mode {
+                    opts.mode(mode & 0o777);
```
lucasfernog: shouldn't we just set the mode here?

plugins/fs/src/commands.rs
```diff
@@ -40,360 +56,747 @@ impl Serialize for CommandError {
...
+                }
+            }
+
+            opts.read(options.read.unwrap_or(true))
```
lucasfernog: I think they should all default to false (via serde instead of all the logic here); what if I want a file to be write-only for security reasons?

FabianLars:
Why was the `close` export removed? Isn't that required for `open` to make sense? (honestly already forgot everything about this pr so genuine question)

amrbashir:
I guess for security reasons, since `close` accepts any `rid`, I think they should use `.close()` on the returned `FsFile` class

plugins/fs/src/commands.rs
amrbashir: tbh I don't know, this was copied from deno implementation but I guess they have a default mode or something, I will just change that to use the mode directly.

plugins/fs/src/commands.rs
amrbashir: sure

amrbashir: should I also change the default (the `None` case) behavior when options were not provided, to set `read` to `false`?

amrbashir: actually we need to keep `read` to `true`, otherwise opening a file without `read` or `write` will fail.

amrbashir: if someone wants a write-only file, they will need to specify `write: true, read: false`

amrbashir:
btw this PR is blocked until https://github.com/tauri-apps/tauri/pull/8370

lucasfernog (dismissed):

lucasfernog (approved):

PR #758, opened 2023-11-23
chore: Add sponsors section to readme

FabianLars:
Should this be added to all plugin readmes too so it shows up on crates.io / npm too?

cc @nothingismagick


FabianLars:
Actually yes, with the mirrors in place this should be in all readmes. gonna add that in a few

plugins/authenticator/README.md
```diff
@@ -103,6 +103,22 @@ if (counter && counter > 0) {
...
 PRs accepted. Please make sure to read the Contributing Guide before making a pull request.
 
+## Partners
+
+<table>
+  <tbody>
+    <tr>
+      <td align="center" valign="middle">
+        <a href="https://crabnebula.dev" target="_blank">
+          <img src="/.github/sponsors/crabnebula.svg" alt="CrabNebula" width="283">
```
amrbashir: how will this be reflected in the mirror? does it copy the `.github` directory from here to the mirror?

FabianLars: ffs good call, forgot about that

FabianLars: ah damn, this also means the ~~tauri/~~ wry/tao readmes will be broken on crates.io right?

edit: tauri only has it in the repo root, not the crate readme...

amrbashir: it should work in tao and wry, the `.github` directory is included, we have a splash file

FabianLars: phew 😌

FabianLars: okay the published v2 plugins already use a https link so i'll do the same. then it works for the mirrors and the package regs.

amrbashir (dismissed):

FabianLars:
i kinda dislike the auto review dismissal 😅

amrbashir (approved):

FabianLars:
thanks :)

PR #769, opened 2023-11-27
fix(dialog): fix message dialog return with custom buttons, fixes #711

### LINKED ISSUE #711: [dialog] v2 confirm always returns true
tauri alpha.17, dialog plugin 2.0.0-alpha.4, macOS
```ts
import { confirm } from "@tauri-apps/plugin-dialog";
const confirmed = await confirm("Are you sure?");
console.log(confirmed);
```

Current behaviour:
Press Ok => returns true
Press Cancel => returns **true**

Expected behaviour:
Press Ok => returns true
Press Cancel => returns **false**

pashokitsme:
Fixes #711 (And removes some unnecessary lines)

### DIFF
--- a/plugins/dialog/src/desktop.rs
+++ b/plugins/dialog/src/desktop.rs
@@ -53,14 +53,8 @@ impl<R: Runtime> Dialog<R> {
 
 #[cfg(not(target_os = "linux"))]
 macro_rules! run_dialog {
-    ($e:expr, $h: ident) => {{
-        std::thread::spawn(move || {
-            let response = tauri::async_runtime::block_on($e);
-            $h(!matches!(
-                response,
-                rfd::MessageDialogResult::No | rfd::MessageDialogResult::Cancel
-            ));
-        });
+    ($e:expr, $h: expr) => {{
+        std::thread::spawn(move || $h(tauri::async_runtime::block_on($e)));
     }};
 }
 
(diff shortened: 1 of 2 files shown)


plugins/dialog/src/desktop.rs
```diff
@@ -226,5 +208,13 @@ pub fn show_message_dialog<R: Runtime, F: FnOnce(bool) + Send + 'static>(
     dialog: MessageDialogBuilder<R>,
     f: F,
 ) {
+    use rfd::MessageDialogResult as Result;
```
amrbashir: Don't rename to `Result` as it might be confused with std's `Result`

plugins/dialog/src/desktop.rs
```diff
@@ -226,5 +208,13 @@ pub fn show_message_dialog<R: Runtime, F: FnOnce(bool) + Send + 'static>(
...
+    let f = move |res| {
+        f(match res {
+            Result::Ok | Result::Yes => true,
+            Result::Custom(s) => s == OK,
```
amrbashir: This will still fail if the they used a custom ok button label, instead, you should match against `dialog.ok_button_label.unwrap_or_else(|| OK.into)`

amrbashir (changes requested):

amrbashir:
Thanks for the contribution :heart:

pashokitsme:
Oops, you're right. Fixed

pashokitsme:
@amrbashir

amrbashir (dismissed):

amrbashir (approved):

PR #775, opened 2023-11-29
chore: Disable workspace inheritance for [package]

FabianLars:
Currently dependabot fails with errors like these:

grafik

and i'd rather have it figure it out again instead of me dismissing 20 incorrect security alerts.


FabianLars:
I think this can only be tested by merging it (or a dependabot config to make it run on this branch, so either way a commit)

PR #784, opened 2023-11-30
fix(updater): fix appimage extraction

kris-ava:
The updater plugin currently does not work on Linux. This PR fixes two issues:

 * .tar.gz is correctly ungzipped instead of just untared
 * in case the archive is malformed or the AppImage is not present in it - nothing will happen (instead of the current behavior which is removing the current binary)

### DIFF
--- a/plugins/updater/src/error.rs
+++ b/plugins/updater/src/error.rs
@@ -63,7 +63,7 @@ pub enum Error {
     #[error("temp directory is not on the same mount point as the AppImage")]
     TempDirNotOnSameMountPoint,
     #[error("binary for the current target not found in the archive")]
-    BinaryNotFound,
+    BinaryNotFoundInAcrhive,
     #[error(transparent)]
     Http(#[from] http::Error),
 }
--- a/plugins/updater/src/error.rs
+++ b/plugins/updater/src/error.rs
@@ -62,6 +62,8 @@ pub enum Error {
     /// Temp dir is not on same mount mount. This prevents our updater to rename the AppImage to a temp file.
     #[error("temp directory is not on the same mount point as the AppImage")]
     TempDirNotOnSameMountPoint,
+    #[error("binary for the current target not found in the archive")]
+    BinaryNotFound,
     #[error(transparent)]
     Http(#[from] http::Error),
 }
--- a/plugins/updater/src/updater.rs
+++ b/plugins/updater/src/updater.rs
@@ -600,6 +600,7 @@ impl Update {
             ffi::OsStr,
             os::unix::fs::{MetadataExt, PermissionsExt},
         };
+        use flate2::read::GzDecoder;
         let archive = Cursor::new(bytes);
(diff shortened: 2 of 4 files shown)


plugins/updater/src/updater.rs
```diff
@@ -642,7 +644,12 @@ impl Update {
                             }
                         }
                     }
+                    // if we have not returned early we should restore the backup
+                    std::fs::rename(tmp_app_image, &self.extract_path)?;
 
+                    // The convention here is to return Ok(()) if the right installer
+                    // was not found in the package, even though this means the update
+                    // failed.
```
amr-crabnebula, line 652: We should probably make this an error instead, e.g. `return Err(Error::AppimageNotFoundInZip)`

amr-crabnebula (changes requested):

plugins/updater/src/updater.rs
amrbashir: We should probably make this an error instead, e.g. `return Err(Error::AppimageNotFoundInZip)`

amrbashir (changes requested):

plugins/updater/src/error.rs
```diff
@@ -62,6 +62,8 @@ pub enum Error {
     /// Temp dir is not on same mount mount. This prevents our updater to rename the AppImage to a temp file.
     #[error("temp directory is not on the same mount point as the AppImage")]
     TempDirNotOnSameMountPoint,
+    #[error("binary for the current target not found in the archive")]
+    BinaryNotFound,
```
amrbashir:
```suggestion
    BinaryNotFoundInAcrhive,
```

plugins/updater/src/updater.rs
```diff
@@ -642,8 +644,10 @@ impl Update {
                             }
                         }
                     }
+                    // if we have not returned early we should restore the backup
+                    std::fs::rename(tmp_app_image, &self.extract_path)?;
 
-                    return Ok(());
+                    return Err(Error::BinaryNotFound);
```
```suggestion
                    return Err(Error::BinaryNotFoundInAcrhive);
```

amrbashir (dismissed):

amrbashir:
Thanks for your contribution

amrbashir (approved):

plugins/updater/src/error.rs
```diff
@@ -62,6 +62,8 @@ pub enum Error {
     /// Temp dir is not on same mount mount. This prevents our updater to rename the AppImage to a temp file.
     #[error("temp directory is not on the same mount point as the AppImage")]
     TempDirNotOnSameMountPoint,
+    #[error("binary for the current target not found in the archive")]
+    BinaryNotFoundInAcrhive,
```
sinistersnare, line 66: There is a typo in this variant, 'Achrive' should be 'Archive'

FabianLars: damn, thanks for catching that!
https://github.com/tauri-apps/plugins-workspace/pull/812

Issue #791, opened 2023-12-02
[single-instance] expected `Arc<Config>`, found `&Config`


yetone:
```rust
 --> /home/runner/.cargo/git/checkouts/plugins-workspace-fd0259fcb18f51ad/e7c72c9/plugins/single-instance/src/platform_impl/linux.rs:40:30
   |
40 |             let id = dbus_id(app.config());
   |                      ------- ^^^^^^^^^^^^ expected `Arc<Config>`, found `&Config`
   |                      |
   |                      arguments to this function are incorrect
   |
   = note: expected struct `Arc<tauri::Config>`
           found reference `&tauri::Config`
```

PR #792, opened 2023-12-03
fix(single-instance): the incorrect type of dbus_id in linux arch

yetone:
fix https://github.com/tauri-apps/plugins-workspace/issues/791


plugins/single-instance/src/platform_impl/linux.rs
```diff
@@ -89,7 +87,7 @@ pub fn init<R: Runtime>(f: Box<SingleInstanceCallback<R>>) -> TauriPlugin<R> {
 
 pub fn destroy<R: Runtime, M: Manager<R>>(manager: &M) {
     if let Some(connection) = manager.try_state::<ConnectionHandle>() {
-        let dbus_name = format!("org.{}.SingleInstance", dbus_id(manager.config()));
+        let dbus_name = format!("org.{}.SingleInstance", dbus_id(&manager.config()));
```
FabianLars: This `&` shouldn't be necessary, right? (Since config() already returns `&Config`)
Same for line 38.

FabianLars (changes requested):
Thank you! Just one small question (see the other comment). Also, could you please add a changefile similar to https://github.com/tauri-apps/plugins-workspace/blob/v2/.changes/fix-permission-notification.md ?

yetone:
‌‌‌‌All have been repaired.

yetone:
‌‌Oops, maybe I should change it back?

https://github.com/tauri-apps/plugins-workspace/actions/runs/7086070602/job/19284953478?pr=792#step:7:14

FabianLars:
Thank you! Another thing i missed is that i can only merge this PR if all commits are signed. If you could set this up it's ready to go (here's a guide: https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#ssh-commit-signature-verification - i recommend the ssh approach because i think it's easier. A guide for gpg is right above that though)

> ‌‌Oops, maybe I should change it back?

Ahh, my bad. The `config()` function was changed in alpha.18, the repo & lockfile is still on alpha.17. I imagine that other plugins will also break when changing that.

FabianLars:
Okay looks like all the other plugins still work after upgrading (at least locally)

yetone:
‌‌‌‌‌‌ My commits are already signed.

amrbashir (approved):

FabianLars:
Well, this was fixed in another commit, sorry about that :/

PR #796, opened 2023-12-05
Do not use rfd 0.12.1

notriddle:
https://github.com/PolyMeilex/rfd/pull/152
conflicts with the way tauri uses it.

> Yeah, this is a decent solution, not much else we can do about this global state bs on C side.
>
> This will obviously blow up as soon as someone has other code that also uses GTK, but let's ignore that for now, as I want to get rid of GTK backend one day anyway (#66).

Yes, Tauri is other code that also uses GTK, and it does, indeed, blow up. The best way around this is to stop using rfd entirely (they're planning to drop the GTK backend, which will probably break since the XDG Portal API doesn't cover message dialogs).

### DIFF
--- a/.changes/dialog-pin-rfd.md
+++ b/.changes/dialog-pin-rfd.md
@@ -1,0 +1,16 @@
+---
+"dialog": "patch"
+---
+
+Pin the version of `rfd` to 0.12.0.
+
+This is a workaround for [PolyMeilex/rfd#152](https://github.com/PolyMeilex/rfd/pull/152), in which rfd took over responsibility for `gtk_init` by running its own event loop thread. As described in that pull request:
+
+> Yeah, this is a decent solution, not much else we can do about this global state bs on C side.
+>
+> This will obviously blow up as soon as someone has other code that also uses GTK, but let's ignore that for now, as I want to get rid of GTK backend one day anyway (#66).
+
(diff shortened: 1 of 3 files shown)


FabianLars:
Oh damn, good finding! Thank you

FabianLars:
(ignore the failed audit check, not your fault)

Just 2 small-ish things before we can merge this:
1) We require all commits to be signed - here's a guide: https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#ssh-commit-signature-verification - i recommend the ssh approach because i think it's easier. A guide for gpg is right above that though
1.5) Once you set up the code signing you can do `git commit --amend && git push --force` to update this PR
2) Please add a changefile like this one: https://github.com/tauri-apps/plugins-workspace/blob/v2/.changes/dialog-async-message-dialog.md

Thanks again

notriddle:
@FabianLars Okay, that should work.

FabianLars (approved):
Thank you!

(as much as i liked the changefile, i had to trim it down because the changelog generator can't handle large changefile that nice imo)

PolyMeilex:
> XDG Portal API doesn't cover message dialogs

It does, via `zenity`. So the same way it used to work before I introduced RFD to Tauri (`tinyfiledialogs` lib calls `zenity` internally)

(In future I also want to do a fallback to `zenity` for file dialogs in cases when xdp is not present.)

PR #797, opened 2023-12-05
[SQL] Fix queries with SQLite JSON_EXTRACT and number params

mitchelljustin:
By default, JSON numbers (and any other JSON value that's not null) are encoded as strings when bound as a parameter to a SQL query. This seems to be fine for most queries because I assume SQLite coerces these number-strings into INTEGER/REAL when compared to a number column.

However, I'm using a JSON column and JSON_EXTRACT() clauses in my SQL queries, and this seems to select zero rows no matter what number I bind to the query params. It looks like SQLite isn't able to coerce bound parameters when extracting and comparing to values from a JSON column.

To fix this, I explicitly check for JSON numbers and bind them as `f64`s to the query params.

### DIFF
--- a/.changes/fix-number-query-params.md
+++ b/.changes/fix-number-query-params.md
@@ -1,0 +1,5 @@
+---
+"sql": patch
+---
+
+Encode JSON number query params as f64 to fix JSON_EXTRACT queries on SQLite
--- a/plugins/sql/src/plugin.rs
+++ b/plugins/sql/src/plugin.rs
@@ -211,6 +211,8 @@ async fn execute(
             query = query.bind(None::<JsonValue>);
         } else if value.is_string() {
             query = query.bind(value.as_str().unwrap().to_owned())
+        } else if let Some(number) = value.as_number() {
+            query = query.bind(number.as_f64().unwrap_or_default())
         } else {
             query = query.bind(value);
         }
@@ -240,6 +242,8 @@ async fn select(
             query = query.bind(None::<JsonValue>);
         } else if value.is_string() {
             query = query.bind(value.as_str().unwrap().to_owned())
+        } else if let Some(number) = value.as_number() {
+            query = query.bind(number.as_f64().unwrap_or_default())
         } else {
             query = query.bind(value);
         }


FabianLars:
Thanks for contributing!

Just 2 small-ish things before we can merge this:

\- 1) We require all commits to be signed - here's a guide: https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#ssh-commit-signature-verification - i recommend the ssh approach because i think it's easier. A guide for gpg is right above that though
\- 1.5) Once you set up the code signing you can do git commit --amend && git push --force to update this PR
\- 2) Please add a changefile like this one: https://github.com/tauri-apps/plugins-workspace/blob/v2/.changes/dialog-async-message-dialog.md

Also i _think_ this should have targetted v1 as that makes it a bit easier for me to sync the fixes between branches but doesn't matter for this PR, just something for the next ones :)

Thanks again!

mitchelljustin:
> Thanks for contributing!
> 
> Just 2 small-ish things before we can merge this:
> 
> * 1. We require all commits to be signed - here's a guide: [docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#ssh-commit-signature-verification](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#ssh-commit-signature-verification) - i recommend the ssh approach because i think it's easier. A guide for gpg is right above that though
> * 1.5) Once you set up the code signing you can do git commit --amend && git push --force to update this PR
> * 2. Please add a changefile like this one: [`v2`/.changes/dialog-async-message-dialog.md](https://github.com/tauri-apps/plugins-workspace/blob/v2/.changes/dialog-async-message-dialog.md?rgh-link-date=2023-12-05T09%3A40%3A25Z)
> 
> Also i _think_ this should have targetted v1 as that makes it a bit easier for me to sync the fixes between branches but doesn't matter for this PR, just something for the next ones :)
> 
> Thanks again!

Great, thanks for the instructions! I'll get on that.

I targeted v2 because that's the version I'm currently using, but if you sync the changes from v1 -> v2 I have no problem re-targeting the branch to v1.

mitchelljustin:
@FabianLars Signed commit and added changefile.

FabianLars (approved):
Thank you :)

PR #802, opened 2023-12-06
chore: Update Tauri to alpha.18

FabianLars:


FabianLars:
~~maybe it's time to ditch the websocket example too. it's making the audit fail because sveltekit still uses vite 4~~ converted it to vanilla ts on v1 branch

lucasfernog (approved):

PR #824, opened 2023-12-16
feat(http): expose proxy configuration

cijiugechu:
The `fetch` method now exposes a proxy field to allow users to configure the proxy for the reqwest client.
Related issues from original repository: 
- [[feat] Proxy support in fetch function in tauri's http API](https://github.com/tauri-apps/tauri/issues/6467)
- [[feat] How to set proxy settings for the webview](https://github.com/tauri-apps/tauri/issues/4263)

Closes tauri-apps/tauri#6467


plugins/http/src/commands.rs
```diff
@@ -20,6 +20,64 @@ pub struct FetchResponse {
...
+) -> crate::Result<reqwest::ClientBuilder> {
+    let ProxyConfig {
+        all,
+        http,
+        https,
+        basic_auth,
+        no_proxy,
+    } = settings;
+
+    let mut builder = builder;
```
amrbashir: this can be removed and `mut` can be added to the function argument
```rs
fn attach_proxy(
    settings: ProxyConfig,
    mut builder: reqwest::ClientBuilder,
) -> crate::Result<reqwest::ClientBuilder> {}
```

plugins/http/src/commands.rs
```diff
@@ -20,6 +20,64 @@ pub struct FetchResponse {
...
+        builder = builder.proxy(proxy);
+    }
+    if let Some(http_proxy) = http {
+        let proxy = attach_auth_and_exclusion(Proxy::http(http_proxy)?);
+        builder = builder.proxy(proxy);
+    }
+    if let Some(https_proxy) = https {
+        let proxy = attach_auth_and_exclusion(Proxy::https(https_proxy)?);
+        builder = builder.proxy(proxy);
+    }
```
amrbashir: I think we can remove the need for a closure here by doing something similar to this:
```rs
let proxy = if let Some(all) = all {
	Some(Proxy::all(all)?)
  } else if let Some(https) = https {
    Some(Proxy::https(https)?)
  } else if let Some(http) = http {
    Some(Proxy::http(http)?)
  } else {
    None
  } ;

if let Some(proxy) = proxy {
  builder = builder.proxy(proxy);
  if let Some(BasicAuth { username, password }) = &basic_auth {
    proxy = proxy.basic_auth(username, password);
  }
  if let Some(no_proxy) = &no_proxy {
    proxy= proxy.no_proxy(NoProxy::from_string(no_proxy));
  }
}
```

plugins/http/guest-js/index.ts
```diff
@@ -26,6 +26,38 @@
...
+  basicAuth?: {
+    username: string;
+    password: string;
+  };
+  /**
+   * A configuration for filtering out requests that shouldn’t be proxied.
+   * Entries are expected to be comma-separated (whitespace between entries is ignored)
+   */
+  noProxy?: string;
+}
```
amrbashir: This interface doesn't enforce using one of `all`, `https` or `http` fields so `basicAuth` or `noProxy` won't have any effect, and also allows using multiple of them at once where on the backend, only one could be used.

I think we can refactor this type to be something like this to enforce these restrictions:
```ts
type Proxy = {
  all: string | ProxyConfig;
} | {
  https: string | ProxyConfig;
} | {
  http: string | ProxyConfig;
} | {};

interface ProxyConfig {
  url: string;
  basicAuth?: {
    username: string;
    password: string;
  };
  noProxy?: string;
}
```

The `string | ProxyConfig` is for easy of use when you only want to set the URL of the proxy. This will require changes on the Rust side in order to be able to deserialize both cases correctly so probably will be an enum.
```rs
enum ProxyConfig {
 Url(Url),
 Config {
   url: Url,
   basic_auth: Option<String>,
   noProxy: Option<String>
 }
}

```

amrbashir (changes requested):

plugins/http/src/commands.rs
```diff
@@ -29,6 +87,7 @@ pub async fn fetch<R: Runtime>(
     data: Option<Vec<u8>>,
     connect_timeout: Option<u64>,
     max_redirections: Option<usize>,
+    proxy: Option<ProxyConfig>,
```
amrbashir: Let's combine these options into a new struct `ClientConifg` and remove `#[allow(clippy::too_many_arguments)]`

amrbashir (changes requested):

plugins/http/guest-js/index.ts
cijiugechu: `http` and `https` are not mutually exclusive, they can co-exist, I think we can write it this way: 
```ts
export type Proxy =
  | {
      all: string | ProxyConfig;
    }
  | {
      http: string | ProxyConfig;
    }
  | {
      https: string | ProxyConfig;
    }
  | {
      http: string | ProxyConfig;
      https: string | ProxyConfig;
    };
```

plugins/http/src/commands.rs
cijiugechu: In my personal opinion, wouldn't this implementation only allow for a single proxy to exist

amrbashir: ah my bad, I thought you could only add one proxy.

plugins/http/guest-js/index.ts
amrbashir: Yeah, I was under the assumption that only one proxy can exist, then I guess we can just simplify this to 
```ts
Proxy = {
  all?: string | ProxyConfig;
  http?: string | ProxyConfig;
  https?: string | ProxyConfig;
}
```

.changes/http-proxy-config.md
```diff
@@ -0,0 +1,6 @@
+---
+"http": minor
+"http-js": minor
+---
+
+The fetch method now exposes a `proxy` field to configure proxy.
```
amrbashir:
```suggestion
Add `proxy` field to `fetch` options to configure proxy.
```

plugins/http/guest-js/index.ts
```diff
@@ -26,6 +26,45 @@
...
 import { invoke } from "@tauri-apps/api/primitives";
 
+/**
+ * Configuration of a proxy that a Client should pass requests to.
+ *
+ * @since 2.0.0
+ */
+export type Proxy =
+  | {
+      all: string | ProxyConfig;
```
amrbashir: We need to also document these fields

plugins/http/guest-js/index.ts
```diff
@@ -26,6 +26,45 @@
...
+  | {
+      http: string | ProxyConfig;
+    }
+  | {
+      https: string | ProxyConfig;
+    }
+  | {
+      http: string | ProxyConfig;
+      https: string | ProxyConfig;
+    };
```
amrbashir:
```suggestion
export type Proxy = {
  all?: string | ProxyConfig;
  http?: string | ProxyConfig;
  https?: string | ProxyConfig;
};
```

plugins/http/src/commands.rs
```diff
@@ -20,16 +20,122 @@ pub struct FetchResponse {
...
+        http: UrlOrConfig,
+        https: UrlOrConfig,
+    },
+    Http {
+        http: UrlOrConfig,
+    },
+    Https {
+        https: UrlOrConfig,
+    },
+}
```
amrbashir:
```suggestion
pub enum ProxyConfig {
    all: Option<UrlOrConfig>,
    http: Option<UrlOrConfig>,
    https: Option<UrlOrConfig>
}
```

plugins/http/src/commands.rs
```diff
@@ -20,16 +20,122 @@ pub struct FetchResponse {
...
+#[serde(rename_all = "camelCase")]
+#[serde(untagged)]
+pub enum UrlOrConfig {
+    Url(String),
+    Config(DetailedProxyConfig),
+}
+
+#[derive(Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct DetailedProxyConfig {
```
amrbashir: Let's rename this to `ProxyConfig` 
```suggestion
pub struct ProxyConfig {
```
and rename `ProxyConfig` to just `Proxy` wich will conflict with `Proxy` from `reqwest` so just use the qualified path for that, i.e. `reqwest::Proxy`

plugins/http/src/commands.rs
```diff
@@ -20,16 +20,122 @@ pub struct FetchResponse {
...
+    no_proxy: Option<String>,
+}
+
+#[derive(Deserialize)]
+pub struct BasicAuth {
+    username: String,
+    password: String,
+}
+
+fn attach_config(
```
amrbashir:
```suggestion
#[inline]
fn attach_config(
```

amrbashir (changes requested):

time-river:
I find wry has been support proxy, therefore I write a simple patch to support it, it's my pull: 
- [feat(wry): support proxy in wry runtime](https://github.com/tauri-apps/tauri/pull/8441).

Can we merge them to one request?

cijiugechu:
> I find wry has been support proxy, therefore I write a simple patch to support it, it's my pull:
> 
> * [feat(wry): support proxy in wry runtime](https://github.com/tauri-apps/tauri/pull/8441).
> 
> Can we merge them to one request?

It seems these two PRs have different use cases, possibly need to let maintainers make the decision.

FabianLars:
Yeah, 2 different use cases, implementations/targets and repositories. Nothing to merge here :)

Edit: What i meant with that last part is that these 2 PRs should stay seperate and both will be merged into their respective repos.

amrbashir (approved):
Thank you

PR #825, opened 2023-12-16
refactor(fs): use scope from tauri core

lucasfernog:
On Android, when we call canonicalize() on "/data/user/0/appid" (which is the data dir), the result is a "/data/data/appid" path, so we need to adjust our scope for that.

Tauri fix: https://github.com/tauri-apps/tauri/pull/8420


amrbashir (changes requested):
I never noticed before but seems like we have a duplicate code of the fs scope module that already exists in Tauri, so I'd suggest turning this PR into a removal of this scope here and reuse the one from tauri.

lucasfernog:
Needs https://github.com/tauri-apps/tauri/pull/8432

amrbashir (approved):

PR #829, opened 2023-12-18
feat(mobile): add biometric plugin

lucasfernog-crabnebula:
This PR adds a new plugin that allows users to authenticate using the mobile device's biometry (or face ID).


plugins/biometric/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "tauri-plugin-biometric"
+version = "1.0.0"
```
amrbashir, line 3: is there a reason why this is not `2.0.0-alpha-0` or did you forget add a change file with `major` bump?

plugins/biometric/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+edition = { workspace = true }
+authors = { workspace = true }
+license = { workspace = true }
+links = "tauri-plugin-biometric"
+
+[package.metadata.docs.rs]
+features = [ "tauri/dox" ]
```
amrbashir: didn't we remove `tauri/dox` feature?

plugins/biometric/README.md
```diff
@@ -0,0 +1,111 @@
...
+        .plugin(tauri_plugin_biometric::init())
+        .run(tauri::generate_context!())
+        .expect("error while running tauri application");
+}
+```
+
+Afterwards all the plugin's APIs are available through the JavaScript guest bindings:
+
+```javascript
+
```
amrbashir: missing example

amrbashir (changes requested):
Android code LGTM and tested as well, nice Work, I will leave iOS to you as usual.

plugins/biometric/guest-js/index.ts
```diff
@@ -0,0 +1,61 @@
...
+}
+
+export async function checkStatus(): Promise<Status> {
+  return invoke("plugin:biometric|status");
+}
+
+export async function authenticate(
+  reason: string,
+  options?: AuthOptions,
+): Promise<void> {
```
amrbashir, line 72: Shouldn't this return whether it is successful or not?

plugins/biometric/Cargo.toml
lucasfernog-crabnebula: i just forgot to add the change file, then when we release covector turns it into 2.0.0-alpha.0

plugins/biometric/guest-js/index.ts
lucasfernog-crabnebula: it throws instead, added documentation for it

plugins/biometric/Cargo.toml
lucasfernog-crabnebula: oops

amrbashir (dismissed):

amrbashir (approved):

PR #834, opened 2023-12-19
refactor(http): migrate to tauri's new resource table

amrbashir:


FabianLars (dismissed):
I don't have access to a computer atm so i only looked at the code on my phone but it looks alright 🤷

Do we also need changefiles for refactors? There was another plugin change so covector would publish it either way. At this point i doubt i'll ever understand changefiles.

amrbashir:
> Do we also need changefiles for refactors? There was another plugin change so covector would publish it either way. At this point i doubt i'll ever understand changefiles.

If the change affects the user, then we add a change file. This effect could be performance, bug fix, new feature or a change in the public API...etc. which doesn't apply for this case.

Edit: actually this PR needs a change file since it has a breaking change in the `Error` enum

FabianLars (approved):

PR #840, opened 2023-12-20
fix(fs-watch): fix typos in DebouncedEvent type

fitztrev:
1. Fix 2 typos in DebouncedEvent type to match [what is returned from notify](https://github.com/notify-rs/notify/blob/2437896199e58283b4858e40f39d299f3d405e26/notify-debouncer-mini/src/lib.rs#L195-L200)

```diff
-  any
+  Any

-  AnyContinous
+  AnyContinuous
```

2. The `DebounceEventResult` [is a vec](https://github.com/notify-rs/notify/blob/2437896199e58283b4858e40f39d299f3d405e26/notify-debouncer-mini/src/lib.rs#L189C33-L189C33) so changed the type to be an array of those objects

### DIFF
--- a/plugins/fs-watch/guest-js/index.ts
+++ b/plugins/fs-watch/guest-js/index.ts
@@ -35,8 +35,8 @@ type RawEventKind =
   | "other";
 
 export type DebouncedEvent =
-  | { kind: "any"; path: string }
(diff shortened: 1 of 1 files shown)


FabianLars (approved):
Thank you!

PR #845, opened 2023-12-21
feat(clipboard): support `readImage` & `writeImage`

cijiugechu:
related issue:
- [[feat] Clipboard support for RTF (rich text format), HTML and Image](https://github.com/tauri-apps/tauri/issues/7750)


plugins/clipboard-manager/src/mobile.rs
```diff
@@ -39,4 +39,8 @@ impl<R: Runtime> Clipboard<R> {
     pub fn read(&self) -> crate::Result<ClipboardContents> {
         self.0.run_mobile_plugin("read", ()).map_err(Into::into)
     }
+
+    pub fn read_image(&self) -> crate::Result<ClipboardContents> {
+        panic!("unsupported on this platform")
```
amrbashir: Return an Error instead of panic

plugins/clipboard-manager/src/commands.rs
```diff
@@ -16,9 +16,17 @@ pub(crate) async fn write<R: Runtime>(
 }
 
 #[command]
-pub(crate) async fn read<R: Runtime>(
+pub(crate) async fn read_text<R: Runtime>(
```
amrbashir, line 28: let's also change `write` command to `write_text`

plugins/clipboard-manager/src/desktop.rs
```diff
@@ -28,13 +30,33 @@ pub struct Clipboard<R: Runtime> {
 
 impl<R: Runtime> Clipboard<R> {
     pub fn write(&self, kind: ClipKind) -> crate::Result<()> {
```
amrbashir: let's remove this and only expose `write_text` and `write_image` variants.

plugins/clipboard-manager/src/desktop.rs
```diff
@@ -28,13 +30,33 @@ pub struct Clipboard<R: Runtime> {
...
+                    .lock()
+                    .unwrap()
+                    .set_image(image)
+                    .map_err(Into::into)
+            }
+            Err(e) => Err(crate::Error::Clipboard(e.to_string())),
+        }
+    }
+
     pub fn read(&self) -> crate::Result<ClipboardContents> {
```
amrbashir: let's remove this as well

plugins/clipboard-manager/Cargo.toml
```diff
@@ -25,3 +25,4 @@ thiserror = { workspace = true }
 
 [target."cfg(any(target_os = \"macos\", windows, target_os = \"linux\", target_os = \"dragonfly\", target_os = \"freebsd\", target_os = \"openbsd\", target_os = \"netbsd\"))".dependencies]
 arboard = "3"
+image   = "0.24"
```
amrbashir, line 28: I am not too happy with the current state of images handling in tauri overall and before we merge this PR I think it would be much better to have a better representation in tauri core similar to electron's `NativeImage` class. 

cc @lucasfernog

amrbashir (changes requested):

simonhyll (approved):

plugins/clipboard-manager/Cargo.toml
FabianLars: isn't the image crate basically the ecosystem default for that? I'd much rather agree on one of its types (like RgbaImage or DynamicImage idk) across all of our crates than creating our own type(s) (again).

FabianLars: Also something i don't get about arboard, they already depend on the image crate but still created their own image type. imho super annoying...

amrbashir: > isn't the image crate basically the ecosystem default for that? I'd much rather agree on one of its types (like RgbaImage or DynamicImage idk) across all of our crates than creating our own type(s) (again).

yes but image crate is a heavy dependency and you may not need most of it, for that reason, crates like `winit`, `muda` and `tray-icon` opted into a light icon type.

What I meant by my previous comment is that we need something like `NativeImage` that is light like `winit`'s `Icon` type, with a few helper methods to support creating from:
- png / jpeg / ico bytes (based on feature flags enabled)
- png / jpeg / ico files (based on feature flags enabled)
- from internal embedded assets (if possible, not sure how yet)
- raw rbga8
- can have a lighter representation in JS (using tauri's new resource table) so we don't need to send bytes forth and back between JS and Rust (unless user wants to get it in JS)
- unlike the existing `tauri::Icon`, it will be a struct not an Enum

FabianLars: Sounds like the helper functions would negate most of the "lightness" compared to image::RgbaImage honestly (with its default features disabled of course) but i don't have any hard feelings about it, as long as we have _one_ type we can re-use across all repos 🤷

amrbashir: Not really, the helper functions will use the lighter crates `png` and `ico` and will be behind a feature flag as well same as it is now in v1.

lucasfernog:
I wish we had https://github.com/1Password/arboard/pull/103 :(

lucasfernog (dismissed):

plugins/authenticator/src/u2f_crate/protocol.rs
```diff
@@ -95,7 +95,7 @@ impl U2f {
         challenge: Challenge,
         response: RegisterResponse,
     ) -> Result<Registration> {
-        if expiration(challenge.timestamp) > Duration::seconds(300) {
+        if expiration(challenge.timestamp) > Duration::try_seconds(300).unwrap() {
```
amrbashir, line 99:
```suggestion
        // Safe to unwrap since 300 is within the constraints of Duration::try_seconds
        if expiration(challenge.timestamp) > Duration::try_seconds(300).unwrap() {
```

plugins/authenticator/src/u2f_crate/protocol.rs
```diff
@@ -152,7 +152,7 @@ impl U2f {
         sign_resp: SignResponse,
         counter: u32,
     ) -> Result<u32> {
-        if expiration(challenge.timestamp) > Duration::seconds(300) {
+        if expiration(challenge.timestamp) > Duration::try_seconds(300).unwrap() {
```
amrbashir, line 157:
```suggestion
        // Safe to unwrap since 300 is within the constraints of Duration::try_seconds
        if expiration(challenge.timestamp) > Duration::try_seconds(300).unwrap() {
```

plugins/clipboard-manager/Cargo.toml
amrbashir, line 28:
```suggestion
```

plugins/clipboard-manager/guest-js/index.ts
```diff
@@ -49,12 +54,65 @@ async function writeText(
...
+ * ```typescript
+ * import { readImage } from '@tauri-apps/plugin-clipboard-manager';
+ *
+ * const clipboardImage = await readImage();
+ * const blob = new Blob([clipboardImage.bytes], { type: 'image' })
+ * const url = URL.createObjectURL(blob)
+ * ```
+ * @since 2.0.0
+ */
+async function readImage(): Promise<ClipboardImage<Uint8Array>> {
```
amrbashir: this should return an instance of `Image` from `@tauri-apps/image`

amrbashir (changes requested):

lucasfernog:
Just waiting on the next tauri release to push the image changes :)

plugins/clipboard-manager/guest-js/index.ts
lucasfernog: in this case we already have the bytes/width/height and the Image class does note expose an API to directly create an image from those values.. doesn't make much sense to send them back to the rust side...

amrbashir: The `Image` class should be constructed on the Rust side then added to the resources_table, we only return an RresourceId to JS and construct the Js Image class with it.

PR #848, opened 2023-12-22
fix(fs): Fix DebouncedEvent type to match what notify-rs returns

fitztrev:
Same fix as #840 this time to the v2 branch

### DIFF
--- a/.changes/fix-debounced-event-type.md
+++ b/.changes/fix-debounced-event-type.md
@@ -2,4 +2,4 @@
 "fs-js": patch
 ---
 
-Fix `DebouncedEvent` type to match what is returned from notify-rs.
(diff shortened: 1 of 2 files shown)


amrbashir:
Thanks for your contributions, could you add a change file in `.changes` directory?

FabianLars:
Hi, since this was fixed on the v1 branch I typically just merge v1 into v2 to sync it, so this pr would probably just cause merge conflicts (and a duplicate commit) :)

I didn't have the time to do that yet but can do it once I'm home today or tomorrow :)

amrbashir:
This shouldn't introduce a conflict since v1 is under `fs-watch` not `fs` directory so I think it is fine to merge.

amrbashir (approved):

PR #867, opened 2023-12-30
docs(sql): Add migrations section to README

mattiloh:
As a newcomer to both Tauri and Rust, I initially encountered some challenges while attempting to integrate SQLite along with a migration workflow into my first Tauri application. However, upon examining the [plugin's source code](https://github.com/tauri-apps/plugins-workspace/blob/v2/plugins/sql/src/plugin.rs), I was pleased to discover that it included functionality related to database migrations. With assistance from ChatGPT, I was able to gain a better understanding of this feature and subsequently crafted a 'Migrations' section for the README file.

Following this guide, I have successfully setup a migration management system in my application. I think this section is a valuable addition to the documentation, which I believe will be beneficial for other developers embarking on similar projects.

### DIFF
--- a/plugins/sql/README.md
+++ b/plugins/sql/README.md
@@ -105,6 +105,68 @@ const result = await db.execute(
 );
 ```
 
+## Migrations
+
+This plugin supports database migrations, allowing you to manage database schema evolution over time.
+
+### Defining Migrations
+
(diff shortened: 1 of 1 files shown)


FabianLars (approved):
Thanks! I love to see docs prs 😍 

Would you happen to mind opening it against the v1 branch too? (No pressure)

mattiloh:
Sure, no problem. Follow-up PR is open 🙂

PR #873, opened 2024-01-02
fix(dialog): On Android, do not add a `Cancel` button to message dialogs

olivierlemasle:
On Android, do not add a `Cancel` button to message dialogs. This is already the current behavior on desktop (I don't know for iOS devices).

Also, add a message dialog to the example app.


amrbashir (approved):
Thank you, tested and works flawlessly

Issue #625, opened 2023-09-28
[notification] I can't import the Schedule class ？。。


lxfxy:
![image](https://github.com/tauri-apps/plugins-workspace/assets/56082897/c95ec380-9340-418c-b5f8-7141dae02741)

![image](https://github.com/tauri-apps/plugins-workspace/assets/56082897/0498874b-c352-4ccb-8a98-0a1008369ca2)

![image](https://github.com/tauri-apps/plugins-workspace/assets/56082897/f57d694d-e0b0-4e00-b350-6b02c89a045f)

"@tauri-apps/plugin-notification": "2.0.0-alpha.1"

ItsRoy69:
Can someone assign me ? Ican work on this issue.

FabianLars:
Awesome, we appreciate it!

PR #874, opened 2024-01-03
fix(notification): export the missing `Schedule` class

cijiugechu:
close [#625](https://github.com/tauri-apps/plugins-workspace/issues/625).


FabianLars:
Awesome, thank you!
Since the constructor of `Schedule` is private i don't think we need to export ScheduleData right? I think that should stay an internal type. I'm open for discussion though if you disagree :)

FabianLars (approved):
Thanks again :)

PR #885, opened 2024-01-10
feat(fs/watch): migrate to `notify-debouncer-full`

dfaust:
- Adds support for notify-debouncer-full

  This complicates the API some more and I'm not super happy with it, yet. But at least this allows Tauri users to take advantage of the more involved `notify-debouncer-full`.

- Adds a watch section to the file system page of the demo app

  image

I'm a contributor to the `notify` crate, on which the fs watcher is built. So if you have any questions or suggestions, please let me/the rest of the notify team know.
One thing that I noticed is that Tauri might be the first application that really uses the event serialization. At least for me this feature has been more of an afterthought until now. I'm not happy with the way events are serialized, maybe we can change that - I'm thinking of using `#[serde(tag = "kind")]` for enums.

### DIFF
--- a/plugins/fs/guest-js/index.ts
+++ b/plugins/fs/guest-js/index.ts
@@ -1110,20 +1110,59 @@ type WatchEvent = {
  * @since 2.0.0
  */
 type WatchEventKind =
-  | "any "
-  | {
-      access?: unknown;
-    }
-  | {
-      create?: unknown;
-    }
-  | {
-      modify?: unknown;
-    }
-  | {
-      remove?: unknown;
-    }
+  | "any"
+  | { access: WatchEventKindAccess }
+  | { create: WatchEventKindCreate }
+  | { modify: WatchEventKindModify }
+  | { remove: WatchEventKindRemove }
   | "other";
+
+/**
+ * @since 2.0.0
+ */
+type WatchEventKindAccess =
+  | { kind: "any" }
+  | { kind: "close"; mode: "any" | "execute" | "read" | "write" | "other" }
+  | { kind: "open"; mode: "any" | "execute" | "read" | "write" | "other" }
+  | { kind: "other" };
+
+/**
+ * @since 2.0.0
+ */
+type WatchEventKindCreate =
+  | { kind: "any" }
+  | { kind: "file" }
+  | { kind: "folder" }
+  | { kind: "other" };
+
+/**
+ * @since 2.0.0
+ */
+type WatchEventKindModify =
+  | { kind: "any" }
+  | { kind: "data"; mode: "any" | "size" | "content" | "other" }
+  | {
+      kind: "metadata";
+      mode:
+        | "any"
+        | "access-time"
+        | "write-time"
+        | "permissions"
+        | "ownership"
+        | "extended"
+        | "other";
+    }
+  | { kind: "name"; mode: "any" | "to" | "from" | "both" | "other" }
+  | { kind: "other" };
+
+/**
+ * @since 2.0.0
+ */
+type WatchEventKindRemove =
+  | { kind: "any" }
+  | { kind: "file" }
+  | { kind: "folder" }
+  | { kind: "other" };
 
 /**
  * @since 2.0.0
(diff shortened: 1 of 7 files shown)


FabianLars:
Hi, thanks so much for contributing! Do you see any value in keeping the mini implementation around?

Also, before you add more commits:
We require all commits of a pr to be signed - here's a guide: https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#ssh-commit-signature-verification - i recommend the ssh approach because i think it's easier. A guide for gpg is right above that though

And we'll also need a changefile like this one: https://github.com/tauri-apps/plugins-workspace/blob/v2/.changes/dialog-async-message-dialog.md, but it probably makes sense to wait for that in case we end up dropping the mini impl.

amrbashir:
let's just move to the full version and remove mini

dfaust:
> Do you see any value in keeping the mini implementation around?

The full debouncer has more features, but is also more opinionated. Still, I think most people would prefer it.
Then again, as the author of the full debouncer I'm biased :). Maybe @0xpr03, the author of the mini debouncer has an opinion on this?
Removing the mini debouncer would simplify the API and we could always add it back if the need arises.

0xpr03:
From what I can see Tauri is a full OS abstraction to most users. With that in mind I think you want the full debouncer. Debouncer-Mini is just a bare debouncer drop-in for when the old one got removed from notify. So it's for people who want nothing more than a thin abstraction on top and who can live with some surprising behaviour from the filesystem.

Even more: You probably want more features than the full debouncer currently offers. Specifically filesystem detection and automatically choosing backends for paths.

dfaust:
I updated the PR.

- I removed `notify-debouncer-mini`
- Renamed `RawEvent` to `Event` because the full debouncer emits the same event type
- Added a .changes files
- And signed my commits

plugins/fs/guest-js/index.ts
```diff
@@ -1093,42 +1093,76 @@ interface WatchOptions {
...
+  /** Keep track of the file system IDs of all files */
+  trackFileIds?: boolean;
 }
 
 /**
  * @since 2.0.0
  */
-type RawEvent = {
-  type: RawEventKind;
+type Event = {
```
amrbashir: Let's rename this to `WatchEvent` and similar do the same to the other types so it conveys that these events are related to the watcher.

plugins/fs/src/watcher.rs
```diff
@@ -60,10 +62,12 @@ fn watch_raw(on_event: Channel, rx: Receiver<notify::Result<Event>>) {
 
 fn watch_debounced(on_event: Channel, rx: Receiver<DebounceEventResult>) {
     spawn(move || {
-        while let Ok(event) = rx.recv() {
-            if let Ok(event) = event {
-                // TODO: Should errors be emitted too?
-                let _ = on_event.send(&event);
+        while let Ok(events) = rx.recv() {
+            if let Ok(events) = events {
```
amrbashir:
```suggestion
        while let Ok(Ok(events)) = rx.recv() {
```

plugins/fs/guest-js/index.ts
```diff
@@ -1093,42 +1093,76 @@ interface WatchOptions {
 interface DebouncedWatchOptions extends WatchOptions {
   /** Debounce delay */
   delayMs?: number;
+  /** Keep track of the file system IDs of all files */
+  trackFileIds?: boolean;
```
amrbashir: can we keep this in another PR, so this PR would be about migration only and also allow for a better discussion of the option

amrbashir (changes requested):

plugins/fs/guest-js/index.ts
```diff
@@ -1232,8 +1264,7 @@ export type {
   FileInfo,
   WatchOptions,
   DebouncedWatchOptions,
-  DebouncedEvent,
-  RawEvent,
+  WatchEvent,
```
amrbashir, line 1267:
```suggestion
  WatchEvent,
  WatchEventKind,
  WatchEventKindAccess,
  WatchEventKindCreate,
  WatchEventKindModify,
  WatchEventKindRemove,
```

amrbashir (approved):

amrbashir:
Thank you

dfaust:
Thanks for merging, I'll create another PR for the track-file-ids feature then.

PR #891, opened 2024-01-13
feat(updater) support proxy

### LINKED ISSUE #887: [updater][feat] allow set proxy for updater
Allow check & download update file through socks5 or http(s) proxy

zhangsan946:
to close #887

### DIFF
--- a/plugins/updater/guest-js/index.ts
+++ b/plugins/updater/guest-js/index.ts
@@ -15,7 +15,7 @@ interface CheckOptions {
    */
   timeout?: number;
   /**
-   * Request proxy
+   * A proxy url to be used when checking and downloading updates.
    */
   proxy?: string;
   /**
--- a/plugins/updater/guest-js/index.ts
+++ b/plugins/updater/guest-js/index.ts
@@ -15,6 +15,10 @@ interface CheckOptions {
    */
   timeout?: number;
   /**
+   * Request proxy
+   */
+  proxy?: string;
+  /**
    * Target identifier for the running application. This is sent to the backend.
    */
   target?: string;
--- a/plugins/updater/src/updater.rs
+++ b/plugins/updater/src/updater.rs
@@ -94,7 +94,7 @@ pub struct UpdaterBuilder {
     endpoints: Option<Vec<Url>>,
     headers: HeaderMap,
     timeout: Option<Duration>,
-    proxy: Option<String>,
+    proxy: Option<Url>,
     installer_args: Option<Vec<String>>,
 }
 
@@ -162,8 +162,8 @@ impl UpdaterBuilder {
         self
     }
 
-    pub fn proxy(mut self, proxy: impl Into<String>) -> Self {
-        self.proxy.replace(proxy.into());
+    pub fn proxy(mut self, proxy: Url) -> Self {
+        self.proxy.replace(proxy);
         self
     }
 
@@ -225,7 +225,7 @@ pub struct Updater {
     current_version: Version,
     version_comparator: Option<Box<dyn Fn(Version, RemoteRelease) -> bool + Send + Sync>>,
     timeout: Option<Duration>,
-    proxy: Option<String>,
+    proxy: Option<Url>,
     endpoints: Vec<Url>,
     #[allow(dead_code)]
     installer_args: Vec<String>,
@@ -285,7 +285,7 @@ impl Updater {
                 request = request.timeout(timeout);
             }
             if let Some(ref proxy) = self.proxy {
-                let proxy = reqwest::Proxy::all(proxy)?;
+                let proxy = reqwest::Proxy::all(proxy.as_str())?;
                 request = request.proxy(proxy);
             }
             let response = request.build()?.get(url).headers(headers.clone()).send().await;
@@ -375,7 +375,7 @@ pub struct Update {
     /// Request timeout
     pub timeout: Option<Duration>,
     /// Request proxy
-    pub proxy: Option<String>,
+    pub proxy: Option<Url>,
     /// Request headers
     pub headers: HeaderMap,
 }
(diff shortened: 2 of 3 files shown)


plugins/updater/src/updater.rs
```diff
@@ -94,6 +94,7 @@ pub struct UpdaterBuilder {
     endpoints: Option<Vec<Url>>,
     headers: HeaderMap,
     timeout: Option<Duration>,
+    proxy: Option<String>,
```
amrbashir: Let's take a Url instead
```suggestion
    proxy: Option<Url>,
```

plugins/updater/src/updater.rs
```diff
@@ -160,6 +162,11 @@ impl UpdaterBuilder {
         self
     }
 
+    pub fn proxy(mut self, proxy: impl Into<String>) -> Self {
```
amrbashir: ditto
```suggestion
    pub fn proxy(mut self, proxy: Url) -> Self {
```

plugins/updater/src/updater.rs
```diff
@@ -217,6 +225,7 @@ pub struct Updater {
     current_version: Version,
     version_comparator: Option<Box<dyn Fn(Version, RemoteRelease) -> bool + Send + Sync>>,
     timeout: Option<Duration>,
+    proxy: Option<String>,
```
amrbashir: ditto
```suggestion
    proxy: Option<Url>,
```

plugins/updater/src/updater.rs
```diff
@@ -360,6 +374,8 @@ pub struct Update {
     pub signature: String,
     /// Request timeout
     pub timeout: Option<Duration>,
+    /// Request proxy
+    pub proxy: Option<String>,
```
amrbashir: ditto
```suggestion
    pub proxy: Option<Url>,
```

plugins/updater/src/updater.rs
amrbashir: We also need to provide a similar option for the JS API

amrbashir (changes requested):
Thanks for your contributions, just a few changes needed.

plugins/updater/src/updater.rs
zhangsan946: Done

plugins/updater/src/updater.rs
zhangsan946: Done

plugins/updater/src/updater.rs
zhangsan946: Done

plugins/updater/src/updater.rs
zhangsan946: Added

zhangsan946:
@amrbashir please take a check again.

plugins/updater/src/updater.rs
```diff
@@ -271,11 +280,15 @@ impl Updater {
...
                 .parse()?;
 
-            let mut request = Client::new().get(url).headers(headers.clone());
+            let mut request = ClientBuilder::new();
             if let Some(timeout) = self.timeout {
                 request = request.timeout(timeout);
             }
-            let response = request.send().await;
+            if let Some(ref proxy) = self.proxy {
+                let proxy = reqwest::Proxy::all(proxy.as_str())?;
```
amrbashir, line 288: I believe `Proxy::all` can take `Url` directly so we don't need to do `.as_str()`, no?

plugins/updater/src/updater.rs
```diff
@@ -384,13 +400,15 @@ impl Update {
...
-        let mut request = Client::new()
-            .get(self.download_url.clone())
-            .headers(headers);
+        let mut request = ClientBuilder::new();
         if let Some(timeout) = self.timeout {
             request = request.timeout(timeout);
         }
-        let response = request.send().await?;
+        if let Some(ref proxy) = self.proxy {
+            let proxy = reqwest::Proxy::all(proxy.as_str())?;
```
amrbashir, line 413: ditto

plugins/updater/src/updater.rs
vangork: It needs to be `reqwest::Proxy::all(proxy.clone())` or `reqwest::Proxy::all(proxy.as_str())`, I took the later to avoid copy which I think performance may be better

plugins/updater/src/updater.rs
vangork: ditto

plugins/updater/src/updater.rs
zhangsan946: It needs to be either `reqwest::Proxy::all(proxy.clone())` or `reqwest::Proxy::all(proxy.as_str())`. I took the latter to avoid copy which I think performance should be better. Which do you prefer?

plugins/updater/src/updater.rs
zhangsan946: ditto

plugins/updater/src/updater.rs
amrbashir: My bad, I guess you need to do `.clone()` because of `ref` and we can't move it out because it will be used again later.

> which I think performance should be better

it shouldn't matter since `Proxy::all` will take the `&str` and call `Url::parse` on it so let's just stick with your choice.

plugins/updater/guest-js/index.ts
```diff
@@ -14,6 +14,10 @@ interface CheckOptions {
    * Timeout in seconds
    */
   timeout?: number;
+  /**
+   * Request proxy
```
amrbashir:
```suggestion
   * A proxy url to be used when checking and downloading updates.
```

amrbashir (approved):

amrbashir:
Thank you

PR #912, opened 2024-01-20
Add track file ids option to fs watcher

dfaust:
Add `trackFileIds` to the `DebouncedWatchOptions`.

When enabled, the fs watcher gets the filesystem ids of all watched files and keeps them in memory in order to be able to stitch together rename events on Windows and MacOS.

When disabled, users may receive events like:

```
rename /home/user/file.txt
rename /home/user/file-backup.txt
```

When enabled, the events would be merged into:

```
rename /home/user/file.txt to /home/user/file-backup.txt
```

The feature is disabled by default. But we might want to enable it by default, as it is pretty useful.
Also, we may want to automatically disable it on Linux, since inotify emits a cookie for rename events, so `notify-debouncer-full` handles them correctly already.

I didn't add a .changes file because `notify-debouncer-full` is new for the v2 release.

### DIFF
--- a/examples/api/src/views/FileSystem.svelte
+++ b/examples/api/src/views/FileSystem.svelte
@@ -12,6 +12,7 @@
   let watchPath = "";
   let watchDebounceDelay = 0;
   let watchRecursive = false;
+  let watchTrackFileIds = false;
   let unwatchFn;
   let unwatchPath = "";
 
@@ -154,6 +155,7 @@
       let options = {
         recursive: watchRecursive,
         delayMs: parseInt(watchDebounceDelay),
+        trackFileIds: watchTrackFileIds,
       };
       if (options.delayMs === 0) {
         fs.watchImmediate(watchPath, onMessage, options)
@@ -236,6 +238,10 @@
   <div>
     <input type="checkbox" id="watch-recursive" bind:checked={watchRecursive} />
     <label for="watch-recursive">Recursive</label>
+  </div>
+  <div>
+    <input type="checkbox" id="watch-track-file-ids" bind:checked={watchTrackFileIds} />
+    <label for="watch-track-file-ids">Track file IDs (requires a debounce delay to be set)</label>
   </div>
   <br />
   <div>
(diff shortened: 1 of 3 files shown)


FabianLars:
Thanks for the follow-up pr!

> The feature is disabled by default. But we might want to enable it by default, as it is pretty useful.
Also, we may want to automatically disable it on Linux, since inotify emits a cookie for rename events, so notify-debouncer-full handles them correctly already.

Just to double check. macos and windows _with_ the feature enabled would behave the same as linux _without_ it?
And what happens if you do enable it on Linux?

dfaust:
> macos and windows with the feature enabled would behave the same as linux without it?

Correct!

> And what happens if you do enable it on Linux?

It will scan the file ids and keep them in memory, but won't actually change the emitted events.

FabianLars:
Cool, thanks! Does it even makes sense to have it optional in the first place then?
I'm afraid that the platform difference would cause more issues than the value the config would bring, especially for frontend heavy users that aren't used to platform differences.

dfaust:
Good question. I guess 99% of users would want this feature enabled, because it makes things easier.
But it comes at a small cost. Then again, the only scenario in which disabling it would make a difference is if you were watching a huge number of files and weren't interested in rename events.

I wouldn't mind removing the option for now. We can add it back if needed.
Also, there is still some potential to save memory in the `notify-debouncer-full` crate. The file ids are currently stored as a hashmap with full paths for every file.

Btw. I was thinking about the track-file-ids feature not being needed on Linux, and I think will disable it on Linux by default directly in the `notify-debouncer-full` crate.

FabianLars:
Yeah, imo removing the option is fine. The IPC's performance hit probably makes the cost unnoticable anyway 😅 
But i'd like to hear @amrbashir's opinion on that matter too.

amrbashir:
I'd go with consistency on all platforms first and provide the option to disable this, i.e add a feature flag and enable it by default.

dfaust:
This PR has been made obsolete by a change in the `notify-debouncer-full` create (notify-rs/notify#557), so I'm closing it now.
I will create a new one, once a new version of `notify-debouncer-full` has been released.
The new version enables file id tracking by default on Windows and MacOS. It can still be disabled, but I tend towards not exposing that option in Tauri (for now). As long as performance is not critical, it simplifies the usage.

@amrbashir: To be clear, the file id tracking fixes an inconsistency between Windows and MacOS on the one side and Linux on the other.

PR #916, opened 2024-01-25
feat(deep-link): Add deep link support for desktop

### LINKED ISSUE #568: [deep-link] v2 - desktop support
We can start with my v1 plugin and then migrate to Amr's bundler work. We'll probably end up with both to also support portable apps.

FabianLars:
needs https://github.com/tauri-apps/tauri/pull/8680


sebasptsch:
This looks interesting, is this on hold for the moment whilst v2 is finalised?

FabianLars:
it was waiting for the beta release with the big allowlist/ACL changes, yes. Since then i was out sick so i couldn't work on this yet.

This week i've also talked with a community member about the url handling mess in the context of cross platform support (android, ios, macos have an explicit event and are single instance by default, while linux and windows spawn a new instance that gets the url as a cli arg), and i'm not sure how, or if at all to handle it yet, maybe it will be in a follow up pr idk.

amrbashir:
single-instance should not be handled by deep-link, it should be handled by the single-instance plugin

FabianLars:
i did not forget about your stance on this topic :P And that's also not reallyyy what i meant, i agree that the deep link plugin shouldn't change this behavior.

I'll have to find the discussions again to remember what i actually meant though lol. We at least need more docs to show how to handle this, or at least stating the platform differences.

Anyway, no _major_ changes are planned.

pauliesnug:
@FabianLars is this plugin ready for release since deep-links was merged into tauri v2? i'd love macos/linux/windows support added to the current mobile plugin.

FabianLars:
not quite. If it's enough to register the schemes at install time, and not at runtime, then using the latest tauri cli already support that. The plugin only adds apis to un/register schemes at runtime.

pauliesnug:
> not quite. If it's enough to register the schemes at install time, and not at runtime, then using the latest tauri cli already support that. The plugin only adds apis to un/register schemes at runtime.

@FabianLars is there documentation on how to do desktop deep linking with the latest tauri cli?

FabianLars:
No, here's how to configure it: https://github.com/tauri-apps/plugins-workspace/pull/916/files#diff-98b5031760ba756df7078118e3724551064a6b92fe1537a0a19ca5daacdda748 (This already works as long as you don't need mobile and therefore don't include this plugin in your project)

You can get the invoked links on macOS like on iOS: https://beta.tauri.app/features/deep-linking/#usage

On Windows and Linux this will open a new instance with the url in the cli args (accessible in rust via `std::env::args()`). If you want to keep it single instance like on macos you need to use the single-instance plugin.

plugins/deep-link/Cargo.toml
```diff
@@ -16,12 +16,17 @@ targets = [ "x86_64-linux-android" ]
...
 serde = { workspace = true }
 serde_json = { workspace = true }
 tauri = { workspace = true }
+tauri-utils = { workspace = true }
 log = { workspace = true }
 thiserror = { workspace = true }
 url = { workspace = true }
+
+[target."cfg(windows)".dependencies]
+winreg = "0.52"
```
amrbashir: we should use the new [`windows-registry`](https://docs.rs/windows-registry/latest/windows_registry/) crate

plugins/deep-link/README.md
```diff
@@ -128,10 +131,12 @@ Afterwards all the plugin's APIs are available through the JavaScript guest bind
 ```javascript
 import { onOpenUrl } from "@tauri-apps/plugin-deep-link";
 await onOpenUrl((urls) => {
-  console.log('deep link:', urls);
+  console.log("deep link:", urls);
 });
 ```
 
+Note that the Plugin will only emit events on macOS, iOS and Android. On Windows and Linux the OS will spawn a new instance of your app with the URL as a CLI argument. If you want your app to behave on Windows & Linux similar to the other platforms you can use the [single-instance](../single-instance/) plugin.
```
amrbashir, line 138: we should also mention that the URL will be passed as a CLI argument, and maybe showcase an example of using both plugins, maybe take https://www.electronjs.org/docs/latest/tutorial/launch-app-from-url-in-another-app as inspiration

plugins/deep-link/src/lib.rs
```diff
@@ -105,10 +154,164 @@ mod imp {
...
+
+                let output = Command::new("xdg-mime")
+                    .args(["default", &file_name, _protocol.as_ref()])
+                    .output()?;
+
+                Ok(String::from_utf8_lossy(&output.stdout).contains(&file_name))
+            }
+
+            #[cfg(not(any(windows, target_os = "linux")))]
+            Ok(true)
```
amrbashir: should probably be false on unsupported platforms

plugins/deep-link/src/lib.rs
```diff
@@ -105,10 +154,164 @@ mod imp {
...
+            {
+                let hkcu = RegKey::predef(HKEY_CURRENT_USER);
+                let base = Path::new("Software")
+                    .join("Classes")
+                    .join(_protocol.as_ref());
+
+                let exe = tauri::utils::platform::current_exe()?
+                    .display()
+                    .to_string()
+                    .replace("\\\\?\\", "");
```
amrbashir: use `dunce` crate instead of manual removal

plugins/deep-link/src/lib.rs
```diff
@@ -105,10 +154,164 @@ mod imp {
...
+                let (cmd, _) =
+                    hkcu.create_subkey(base.join("shell").join("open").join("command"))?;
+
+                cmd.set_value("", &format!("{} \"%1\"", &exe))?;
+            }
+
+            #[cfg(target_os = "linux")]
+            {
+                let bin = tauri::utils::platform::current_exe()?;
+                // TODO: Research if this conflicts with appimage integrations somehow. If it does, re-add a `-handler` suffix if running in an appimage.
```
amrbashir: is this TODO resolved?

plugins/deep-link/src/lib.rs
```diff
@@ -105,10 +154,164 @@ mod imp {
...
+                let file_name = format!(
+                    "{}.desktop",
+                    tauri::utils::platform::current_exe()?
+                        .file_name()
+                        .unwrap()
+                        .to_string_lossy()
+                );
+
+                let output = Command::new("xdg-mime")
+                    .args(["default", &file_name, _protocol.as_ref()])
```
amrbashir: shouldn't this be `query` instead of `default`?

amrbashir (changes requested):

plugins/deep-link/Cargo.toml
FabianLars: right, totally forgot about that, thanks.

plugins/deep-link/README.md
FabianLars: My idea was to include the single-instance plugin on the website (the url param is mentioned)

plugins/deep-link/src/lib.rs
FabianLars: My idea was that this would make cross platform handling easier like !isRegistered ? register
I see how that could cause problems too but it's similar for false i guess

Maybe a new Error::UnsupportedPlatform or would that be too harsh? At least it would be easy to spot if it returns true/false because of its actual registration status.

plugins/deep-link/src/lib.rs
FabianLars: no

plugins/deep-link/src/lib.rs
FabianLars: maybeee, the pr is too old to remember, i'll check

plugins/deep-link/src/lib.rs
FabianLars: p.s. i can make isRegistered work on macos too so for this api it's only about mobile. But the question about the error remains for mobile, and the other apis.

plugins/deep-link/src/lib.rs
FabianLars: i don't know what the hell i did but the whole linux implementation is wrong. One would imagine that i'd be able to copy paste my own code from the v1 plugin into this _but apparently not_ 😡

faragly:
@FabianLars Is it possible to intercept `deep-link://` by an application running in dev mode on mac OS? I am currently experiencing a problem in dev mode because only the installed application intercepts links, the running `tauri dev` instance does not receive links

FabianLars:
No it's not possible on macos.

The way I did it in the [unofficial plugin](https://github.com/FabianLars/tauri-plugin-deep-link#macos) was to have something similar to the single-instance plugin built into the deep link plugin for _debug_ builds (tauri dev and tauri build --debug) but that still required you to build the app once and install it but then tauri dev worked. The official single-instance plugin does not help here.

I wasn't really comfortable with adding this to the official plugin, especially because many devs struggled with the debug app requirements, but we'll see, maybe we can come up with something better (doubt it tbh).

Either way i consider it out of scope for this PR.

plugins/deep-link/README.md
amrbashir: would probably still make sense to include the example here as well so it can be found on npm and docs.rs as well

plugins/deep-link/src/lib.rs
amrbashir: > Maybe a new Error::UnsupportedPlatform or would that be too harsh? At least it would be easy to spot if it returns true/false because of its actual registration status.

maybe `Option<bool>` ?

FabianLars: That only works if we're really only talking about this api, and not syncing behavior with un/register (which we should consider imo)

plugins/deep-link/README.md
FabianLars: hmm ok

plugins/deep-link/src/lib.rs
amrbashir: if other APIs gonna return an `Error::UnsupportedPlatform` then this one should too

CrabNejonas:
I've been testing this internally and it seems that on macOS this PR does not emit the `deep-link://new-url` event when the app is not already running, is that intentional?

FabianLars:
Yes. I honestly don't even want to think about how to get the timing for that right (it's basically the "how can i use emit in the setup hook" question). You can use getCurrent/get_current api on app start instead.

CrabNejonas:
> Yes. I honestly don't even want to think about how to get the timing for that right (it's basically the "how can i use emit in the setup hook" question). You can use getCurrent/get_current api on app start instead.

ohh okay thx thats a fine solution imo

lucasfernog-crabnebula:
I think it would be valuable to add a Rust event (maybe a closure) for the new-url event, currently it's tricky to get the URLs because the current value is filled after emitting the webview event.

FabianLars:
I'm a bit dense today, could you explain the issue differently somehow? 😅

lucasfernog-crabnebula:
nvm we can just document the new-url event payload format and users deserialize manually

plugins/deep-link/src/lib.rs
FabianLars: i added back the -handler suffix. afaik this is how you're (kinda) supposed to do it, even if it breaks pop os 🤷 

if someone else wanna research this more deeply then they can go ahead, but for me this is good enough.

FabianLars:
i didn't test the linux changes i just pushed but will mark this as ready for review already because yolo (rly tho please review it, and if you test the linux implementation for me that'd be even better 🙃)

assuming everything is correct, and we're happy with the UnsupportedPlatform error, the only thing left to do is adding desktop support to the example and adding the single-instance plugin to the readme as requested by Amr.

FabianLars:
probably a good idea to split it into multiple files per platform cause those cfg flags are urgh

fax1ty:
How to register a schema? 
Like in v1
```
tauri_plugin_deep_link::register("custom-schema", move |request| {
   handle.emit_all("scheme-request-received", request).unwrap();
})
```

Is it not ready yet?

FabianLars:
You can do this: https://github.com/tauri-apps/plugins-workspace/pull/916/files#diff-98b5031760ba756df7078118e3724551064a6b92fe1537a0a19ca5daacdda748 to make the app register them on installation.

Or you use the good old register functions to register them at runtime, yes.
However, they won't accept a handler function anymore, but the plugin will instead emit an event on macos, ios and android and on windows and linux the url will be in `std::env::args[1]`.
The single instance mechanism was also removed from the plugin so you'll have to use the separate single-instance plugin now (only windows and linux).

fax1ty:
@FabianLars Can I use it right now? Like so 
```toml
tauri-plugin-deep-link = { git = "https://github.com/tauri-apps/plugins-workspace", branch = "desktop-deep-links" }
```

I need to migrate Tauri v1 application to v2, but it uses the schema functionality

lucasfernog (approved):

amrbashir (approved):

pauliesnug:
@FabianLars are there docs for this since its been merged now?

FabianLars:
> @FabianLars are there docs for this since its been merged now?

No, i did not expect that this would get merged as-is (see my comments above), yet alone actually released, so i didn't work on docs. I can't give you an ETA on it right now.

PR #919, opened 2024-01-27
type: add OpenDialogReturn generic type

iShibi:
Add generic type `OpenDialogReturn` to conditionally define the return type of the `open` function. It replaces multiple declarations of the `open` function.

### DIFF
--- a/plugins/dialog/guest-js/index.ts
+++ b/plugins/dialog/guest-js/index.ts
@@ -103,8 +103,8 @@ type OpenDialogReturn<T extends OpenDialogOptions> = T["directory"] extends true
     ? string[] | null
     : string | null
   : T["multiple"] extends true
-  ? FileResponse[] | null
-  : FileResponse | null;
+    ? FileResponse[] | null
+    : FileResponse | null;
 
 /**
  * Open a file/directory selection dialog.
--- a/plugins/dialog/guest-js/index.ts
+++ b/plugins/dialog/guest-js/index.ts
@@ -98,18 +98,14 @@ interface ConfirmDialogOptions {
   cancelLabel?: string;
 }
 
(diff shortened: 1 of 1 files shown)


amrbashir:
Is there any benefits over function overloading? 

Also you need to run `pnpm format` in the repo root

iShibi:
> Is there any benefits over function overloading?
> 
> Also you need to run `pnpm format` in the repo root

Conditional Types are preferred over Function Overloading.
https://www.typescriptlang.org/docs/handbook/2/conditional-types.html

image

FabianLars:
is it just me or is the variant based function overloading much easier to read and understand (both in this PR and in the example in your screenshot)?

plugins/dialog/guest-js/index.ts
```diff
@@ -98,18 +98,14 @@ interface ConfirmDialogOptions {
...
-async function open(
-  options?: OpenDialogOptions & { multiple?: true; directory?: false },
-): Promise<null | FileResponse[]>;
-async function open(
-  options?: OpenDialogOptions & { multiple?: false; directory?: true },
-): Promise<null | string>;
-async function open(
-  options?: OpenDialogOptions & { multiple?: true; directory?: true },
-): Promise<null | string[]>;
+type OpenDialogReturn<T extends OpenDialogOptions> = T["directory"] extends true
```
amrbashir, line 101: We need to export this type as well.

amrbashir (changes requested):
I agree with @FabianLars that overloading is easier to read but given that the docs will contain [multiple definitions](https://beta.tauri.app/references/v2/js/dialog/#open), I will go with this generic type solution

amrbashir (approved):

PR #925, opened 2024-02-01
Add TLS configuration

### LINKED ISSUE #694: [websocket] How to ignore certificate errors in websocket?
How to ignore certificate errors in websocket?

Uncaught (in promise) TLS error: native-tls error: A certificate chain processed, but terminated in a root certificate which is not trusted by the trust provider. (os error -2146762487)

### LINKED ISSUE #917: [websocket] Use TLS Client certificate
We are working on a project that requires TLS client certificates during the TLS handshake. There is currently no way to configure anything TLS related for websockets as far as I can tell (somewhat related to #694 as well).

I was wondering if you'd accept a PR that introduces such a configuration. Tungstenite has an API that allows to pass a TLS Connector (`connect_async_tls_with_config`) that can be created using `native-tls::TlsConnector`.

I did some testing and something like this would work:
https://github.com/jnugh/plugins-workspace/commit/037b58c4b5763569f20deec69bd56560de19ba31

A user could just add a `TlsConfig` using manage and it would be used for new WebSocket connections and pass whatever is needed.

I could create a PR for this if you'd like.

jnugh:
As described in #917 there is currently no way to specify parameters for the TLS connections which is required if one would like to (in our case) use a TLS client certificate or maybe even disable certificate checks altogether as in #694).

I'm not sure if the approach on how to supply configuration parameters is a good one (I just added a configuration struct using `manage` that can be changed from other parts of the application).

Let me know if any documentation should be done as well :) 

Closes #917. Closes #694

### DIFF
--- a/plugins/websocket/src/lib.rs
+++ b/plugins/websocket/src/lib.rs
@@ -8,13 +8,13 @@ use tauri::{
 };
 use tokio::{net::TcpStream, sync::Mutex};
 use tokio_tungstenite::{
-    connect_async_with_config,
+    connect_async_tls_with_config,
     tungstenite::{
         client::IntoClientRequest,
         protocol::{CloseFrame as ProtocolCloseFrame, WebSocketConfig},
         Message,
     },
-    MaybeTlsStream, WebSocketStream,
+    Connector, MaybeTlsStream, WebSocketStream,
 };
 
 use std::collections::HashMap;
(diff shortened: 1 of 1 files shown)


FabianLars:
Thanks for contributing and sorry for the delay :)

> I'm not sure if the approach on how to supply configuration parameters is a good one (I just added a configuration struct using manage that can be changed from other parts of the application).

I think it'd be better to just refactor plugin init to a builder similar to https://github.com/tauri-apps/plugins-workspace/blob/7cefa2b0bd0ed56cebc4bc4888dc8bdcde6c138d/plugins/sql/src/plugin.rs#L270-L287 (careful, that's v2 but the syntax should be the same in v1)

FabianLars (approved):

jnugh:
I completely forgot that there was a todo in this PR. Thank you very much for updating and merging the change 🙏🏽

rtiagom:
When can we expect this to be released?

PR #930, opened 2024-02-05
Added Security Policy to All Plugins

### LINKED ISSUE #599: Plugins in this Workspace Should Have a Mandatory SECURITY.md
During looking into the new plugin based structure for v2 I noticed that there is no threat model, security boundary assumptions and way to contact the original author or maintainer in a reasonable secure way.

I would propose to create a template for this file, apply it to existing plugins and once it has matured require it for all officially maintained plugins (aka all plugins in this workspace).

tillmann-crabnebula:
This PR adds a simplistic contact policy to all remaining plugins, which did not have this previously.
The workspace repository itself also contains this policy now.

The threat model of each plugin will be extended once more fine grained permissions are added.

Fixes #599 with at least a contact policy.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,23 @@
+# Security Policy
+
+**Do not report security vulnerabilities through public GitHub issues.**
+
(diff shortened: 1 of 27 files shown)


lucasfernog:
Can you also add the SECURITY.md file to shared/template?

tillmann-crabnebula:
Added to the template as well.
Some concerns could be that others will use us as security contact if they base their plugin on this template 😅 but we can remove it if this happens more than occasionally.

lucasfernog (approved):

PR #945, opened 2024-02-13
refactor!(dialog): change `type` field in JS to `kind`

### LINKED ISSUE #944: Dialog type doesn't work when calling v2 dialog from js side
```js
message('1', { type: 'warning' })
```

image

Legend-Master:
Fix #944 

Before
```js
await invoke("plugin:dialog|confirm", {
	message: "1",
	type_: 'error',
});
```
image

After
```js
await invoke("plugin:dialog|confirm", {
	message: "1",
	type: 'error',
});
```
image


amrbashir (dismissed):

amrbashir:
I think the better choice here is to rename the `type` field to `kind` to match the Rust API.

amrbashir (approved):

Legend-Master:
Yes, if we don't mind breaking the old API, I'm fine with both ways 😀

amrbashir:
Yeah it is better to change now in beta than when stable.

amrbashir:
Thanks for your contribution btw.

PR #951, opened 2024-02-14
ReadMe updated with all plugins details

kunjee17:


FabianLars (approved):
Thank you :)

p.s. i'm not sure it's 100% correct but it's better than the void we had before so let's merge this :)

PR #960, opened 2024-02-17
fix(http): response with status code 204

### LINKED ISSUE #957: [http][v2] When fetch return the 204 status code, new Response() throw an error
If the status code is 204, the Response() constructor does not allow setting the body.

- In chrome, error: `Uncaught TypeError: Failed to construct 'Response': Response with null body status cannot have body`
- In safari, error: `TypeError: Response cannot have a body with the given status`

https://github.com/tauri-apps/plugins-workspace/blob/2f0e30852acf5d84e65b4ae16fd7c1f5d648c55b/plugins/http/guest-js/index.ts#L158-L166

Try
```ts
const body = await invoke<ArrayBuffer>('plugin:http|fetch_read_body', {
    rid: responseRid,
})
const resData = body.byteLength ? new Uint8Array(body) : null
const res = new Response(resData, { 
```
or
```ts
const res = new Response(status === 204 ? null : new Uint8Array(body), {
```
can fix this issue, unsure which one is more suitable

ppbl:
Close #957 

The code has been fine tuned based on #957 because I found that the body parameter of the `Response` constructor supports `ArrayBuffer`, so `new Uint8Array` is no longer needed.


plugins/http/guest-js/index.ts
```diff
@@ -155,11 +155,11 @@ export async function fetch(
     rid,
   });
 
-  const body = await invoke<number[]>("plugin:http|fetch_read_body", {
+  const body = await invoke<ArrayBuffer>("plugin:http|fetch_read_body", {
```
amrbashir: invoke is not guaranteed to return an `ArrayBuffer`, it will only return `ArrayBuffer` if using the v2 IPC protocol through `fetch`, if it used v1 IPC (used on Linux and sometimes on macOS (remote urls)), it will be an array. 
```suggestion
  const body = await invoke<ArrayBuffer | number[]>("plugin:http|fetch_read_body", {
```

and so later we need to check if it is a normal array or array buffer

```js
(body instanceof ArrayBuffer && body.byteLength ) || (body instanceof Array && body.length > 0) ? body : null
```

amrbashir (changes requested):

amrbashir:
you need to run `pnpm build` so that `src/api-iife.js` file is regenerated.

amrbashir (approved):

Issue #965, opened 2024-02-18
[global-shortcut][v2] How to register a shortcut using `tauri_plugin_global_shortcut`


MystiPanda:
During the migration process, I found that the `register` function in the `tauri_plugin_global_shortcut` plug-in does not accept the `handler` parameter. How can I register a shortcut key?

Thanks and looking forward to your reply:)

amrbashir:
I have updated the example, check it out here https://github.com/tauri-apps/plugins-workspace/tree/v2/plugins/global-shortcut#usage

MystiPanda:
emm, with all due respect, this design seems a bit counter-intuitive. Is this part of the api likely to change in the future?

jasonterando:
Hi, I'm on v2.0, beta 0 for@tauri-apps/plugin-global-shortcut.  When running an application with the below main(), I **do** see the "Setting up global shortcut" message printed, but I **do not** see the "Shortcut" println coming up on every hotkey-ish type of key press, and I do not (nor do I see the Ctrl-N println for that matter).  

I'm running a NextJS/React app, if that is relevant.  JavaScript registrations using `register` still work, so I could fall back to that if necessary.

What am I missing?  Is there an additional init() like there is for other plugins?

Thanks!

```
 fn main() {
    tauri::Builder::default()
        .invoke_handler(tauri::generate_handler![open_workbook, save_workbook, run_request, cancel_request, clear_cached_authorization])
        .plugin(tauri_plugin_fs::init())
        .plugin(tauri_plugin_dialog::init())
        .setup(|app| {
            #[cfg(desktop)]
            {
              println!("Setting up global shortcut");
              use tauri_plugin_global_shortcut::{Shortcut, Code, Modifiers};
              let ctrl_n_shortcut = Shortcut::new(Some(Modifiers::CONTROL), Code::KeyN);
              app.handle().plugin(
                tauri_plugin_global_shortcut::Builder::with_handler(move |app, shortcut| {
                    println!("{:?}", shortcut);
                    if shortcut == &ctrl_n_shortcut {
                        println!("Ctrl-N Detected!");
                        app.emit_to(EventTarget::any(), "NEW",  "").unwrap()
                    }
                })
                .build()
              )?;
            }
            Ok(())            
        })
        .run(tauri::generate_context!())
        .expect("error running app");
}
```

FabianLars:
[...]

PR #969, opened 2024-02-19
refactor!(global-shortcut): better APIs DX

### LINKED ISSUE #1012: the "tauri-plugin-global-shortcut" plugin never removes items when unregistering shortcuts.
Hello, I've gone through the code of the "tauri-plugin-global-shortcut" project. 
I noticed that in the `GlobalShortcut` struct, there's no code found within the `unregister` and `unregister_all` methods to remove items stored in the `shortcuts` property. 
Consequently, this could lead to erroneous returns when querying with the `is_registered` method after calling the `unregister` method.

```
    pub fn register_all<T: TryInto<ShortcutWrapper>, S: IntoIterator<Item = T>>(
        &self,
        shortcuts: S,
    ) -> Result<()>
    where
        T::Error: std::error::Error,
    {
        let mut s = Vec::new();
        for shortcut in shortcuts {
            s.push((try_into_shortcut(shortcut)?, None));
        }
        self.register_all_internal(s, ShortcutSource::Rust)
    }
    pub fn unregister<S: TryInto<ShortcutWrapper>>(&self, shortcut: S) -> Result<()>
    where
        S::Error: std::error::Error,
    {
        acquire_manager(&self.manager)?
            .unregister(try_into_shortcut(shortcut)?)
            .map_err(Into::into)
        //Note:
        //The "self.shortcuts" property is not updated here.
    }

    pub fn unregister_all<T: TryInto<ShortcutWrapper>, S: IntoIterator<Item = T>>(
        &self,
        shortcuts: S,
    ) -> Result<()>
    where
        T::Error: std::error::Error,
    {
        let mut s = Vec::new();
        for shortcut in shortcuts {
            s.push(try_into_shortcut(shortcut)?);
        }
        acquire_manager(&self.manager)?
            .unregister_all(&s)
            .map_err(Into::into)
        //Note:
        //The "self.shortcuts" property is not updated here.
    }
```
[https://github.com/tauri-apps/tauri-plugin-global-shortcut/blob/v2/src/lib.rs](url)

amrbashir:
ref: https://github.com/tauri-apps/plugins-workspace/issues/965

### DIFF
--- a/.changes/global-shortcut-refactor.md
+++ b/.changes/global-shortcut-refactor.md
@@ -4,9 +4,8 @@
 
 **Breaking change** Refactored the plugin Rust APIs for better DX and flexibility:
 
-- `GlobalShortcutExt::register` and `GlobalShortcutExt::register_all` now takes a second argument, which is a handler to be called when the shortcut is triggered.
-- Add `GlobalShortcutExt::register_without_handler` and `GlobalShortcutExt::register_all_without_handler` that work the same as old APIs (i.e. register without a handler).
-- Changed `Builder::with_handler` to be a method instead of a static method.
+- `GlobalShortcutExt::register` and `GlobalShortcutExt::register_all` now takes a second argument, which is a optional handler to be called when the shortcut is triggered.
+- Changed `Builder::with_handler` to be a method instead of a static method, it will also be triggered for any and all shortcuts.
   ```rs
   Builder::new().with_handler().build()
   ```
@@ -14,4 +13,4 @@
   ```rs
   Builder::with_handler().build()
   ```
-- if using `Builder::with_handler`, it will be triggered for all shortcuts, even if the shortcut is registered through JS or through `GlobalShortcutExt::register` and `GlobalShortcutExt::register_without_handler`.
+- if using `Builder::with_handler`, it will be triggered for all shortcuts, even if the shortcut is registered through JS or through `GlobalShortcutExt::register`.
--- a/.changes/global-shortcut-refactor.md
+++ b/.changes/global-shortcut-refactor.md
@@ -4,13 +4,6 @@
 
 **Breaking change** Refactored the plugin Rust APIs for better DX and flexibility:
 
-- `GlobalShortcutExt::register` and `GlobalShortcutExt::register_all` now takes a second argument, which is a optional handler to be called when the shortcut is triggered.
-- Changed `Builder::with_handler` to be a method instead of a static method, it will also be triggered for any and all shortcuts.
-  ```rs
-  Builder::new().with_handler().build()
-  ```
-  instead of
-  ```rs
-  Builder::with_handler().build()
-  ```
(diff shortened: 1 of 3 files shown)


.changes/global-shortcut-refactor.md
```diff
@@ -0,0 +1,17 @@
+---
+"global-shortcut": "patch"
+---
+
+**Breaking change** Refactored the plugin Rust APIs for better DX and flexibility:
+
+- `GlobalShortcutExt::register` and `GlobalShortcutExt::register_all` now takes a second argument, which is a handler to be called when the shortcut is triggered.
+- Add `GlobalShortcutExt::register_wihtout_handler` and `GlobalShortcutExt::register_all_wihtout_handler` that work the same as old APIs (i.e. register without a handler).
```
simonhyll:
```suggestion
- Add `GlobalShortcutExt::register_without_handler` and `GlobalShortcutExt::register_all_without_handler` that work the same as old APIs (i.e. register without a handler).
```

simonhyll (changes requested):

lucasfernog:
Do we reall need the without_handler variants? can't users just define an empty closure and then the global listener handles the event?

amrbashir:
I am not happy with it either, the idea is to provide a method to register a shortcut without handler that is appealing to use, without using an empty closure, but a long method name is not good either, maybe just call it `register2`?

lucasfernog:
maybe `register_without_handler` keeps its name `register` and the new API is named `on_shortcut`? so it's clear you intend on providing a callback there

that's even a non breaking change I believe

amrbashir:
> maybe `register_without_handler` keeps its name `register` and the new API is named `on_shortcut`? so it's clear you intend on providing a callback there

I am afraid that might confuse users because `on_` methods throughout tauri, are used to register just the listener.

amrbashir:
Changed to take an `Option` instead

lucasfernog:
> > maybe `register_without_handler` keeps its name `register` and the new API is named `on_shortcut`? so it's clear you intend on providing a callback there
> 
> I am afraid that might confuse users because `on_` methods throughout tauri, are used to register just the listener.

that's what the on_ here will do:

```rust
fn on_shortcut(&self, shortcut: &str, handler: impl Fn()) {}
fn register(&self, shortcut: &str) {}
```

amrbashir:
maybe I am overthinking it, let's roll with that and see user feedback

amrbashir:
> that's what the on_ here will do:
> 
> ```rust
> fn on_shortcut(&self, shortcut: &str, handler: impl Fn()) {}
> fn register(&self, shortcut: &str) {}
> ```

hmm if we go with that approach, there will be inconsistencies between the JS and Rust APIs

lucasfernog:
it's already inconsistent though :joy: JS is just different.. right now i wish we had function overloading in rust :D

amrbashir:
Alright, added that along with a few methods on the `Builder` that should make it more Rusty hopefully.

lucasfernog (approved):

PR #976, opened 2024-02-21
fix(upload): return type on POST is now string

lyager:
A POST to a webserver can not always be expected to be a JSON response.

Success is now determined by the HTTP return code.

Upon success the body content is returned as a string.

### DIFF
--- a/plugins/upload/src/lib.rs
+++ b/plugins/upload/src/lib.rs
@@ -117,7 +117,10 @@ async fn upload<R: Runtime>(
     if response.status().is_success() {
         response.text().await.map_err(Into::into)
     } else {
-        Err(Error::HttpErrorCode(response.status().as_u16()))
+        Err(Error::HttpErrorCode(
+            response.status().as_u16(),
+            response.text().await.unwrap_or_default(),
+        ))
     }
 }
 
--- a/plugins/upload/src/lib.rs
+++ b/plugins/upload/src/lib.rs
@@ -115,9 +115,10 @@ async fn upload<R: Runtime>(
 
     let response = request.send().await?;
     if response.status().is_success() {
-        return response.text().await.map_err(Error::from);
+        response.text().await.map_err(Into::into)
+    } else {
+        Err(Error::HttpErrorCode(response.status().as_u16()))
     }
-    Err(Error::HttpErrorCode(response.status().as_u16()))
 }
 
 fn file_to_body<R: Runtime>(id: u32, window: Window<R>, file: File) -> reqwest::Body {
--- a/plugins/upload/src/lib.rs
+++ b/plugins/upload/src/lib.rs
@@ -96,10 +96,13 @@ async fn upload<R: Runtime>(
 ) -> Result<String> {
     // Read the file
     let file = File::open(file_path).await?;
+    let file_len = file.metadata().await.unwrap().len();
 
     // Create the request and attach the file to the body
     let client = reqwest::Client::new();
-    let mut request = client.post(url).body(file_to_body(id, window, file));
+    let mut request = client.post(url)
+        .header(reqwest::header::CONTENT_LENGTH, file_len)
+        .body(file_to_body(id, window, file));
 
     // Loop trought the headers keys and values
     // and add them to the request object.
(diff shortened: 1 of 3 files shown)


lyager:
This might be more of a suggestion - not all web services receiving a post returns JSON. Propose to leave result processing to the end-user

FabianLars:
Hmm, you're right with that, however we actually ignore the return value on the js side so we may as well return `Result<()>` from the command instead so we can get rid of both res.json() and res.text() and really only use the status.

For that status error i'd prefer if you could add a new Error enum variant instead of using ContentLength

Lastly, it would be much appreciated if you could apply the same to the download command after making those changes :)

P.S. Thank you for contributing!

lyager:
Thanks for the feedback @FabianLars I somehow missed your response. I'll give it another shot hopefully soon

lyager:
@FabianLars Ok, I think that was all of your suggestions :) What do you think?

lyager:
.. my default linter formater apperantly is different that `cargo fmt` - I fixed it. Sorry.

plugins/upload/src/lib.rs
```diff
@@ -108,8 +114,10 @@ async fn upload<R: Runtime>(
     }
 
     let response = request.send().await?;
-
-    response.json().await.map_err(Into::into)
+    if response.status().is_success() {
+        return response.text().await.map_err(Error::from);
+    }
+    Err(Error::HttpErrorCode(response.status().as_u16()))
```
lucasfernog:
```suggestion
    if response.status().is_success() {
        response.text().await.map_err(Into::into)
    } else {
        Err(Error::HttpErrorCode(response.status().as_u16()))
    }
```

FabianLars:
There's still this part open from above:
> Hmm, you're right with that, however we actually ignore the return value on the js side so we may as well return Result<()> from the command instead so we can get rid of both res.json() and res.text() and really only use the status.

So if there's any value in the response body we'll also have to fix the js side to actually forward it to the user.

lyager:
@FabianLars You're right. It should be fixed now.

I also added @lucasfernog suggestion.

I'll do some extensive test, but I think this should finally be it 😅

plugins/upload/src/lib.rs
```diff
@@ -29,6 +29,8 @@ pub enum Error {
     Request(#[from] reqwest::Error),
     #[error("{0}")]
     ContentLength(String),
+    #[error("{0}")]
+    HttpErrorCode(u16),
```
lucasfernog:
```suggestion
    #[error("request failed with status code {0}: {1}")]
    HttpErrorCode(u16, String),
```

plugins/upload/src/lib.rs
```diff
@@ -108,8 +114,11 @@ async fn upload<R: Runtime>(
     }
 
     let response = request.send().await?;
-
-    response.json().await.map_err(Into::into)
+    if response.status().is_success() {
+        response.text().await.map_err(Into::into)
+    } else {
+        Err(Error::HttpErrorCode(response.status().as_u16()))
```
lucasfernog:
```suggestion
        Err(Error::HttpErrorCode(
            response.status().as_u16(),
            response.text().await.unwrap_or_default(),
        ))
```

lucasfernog (dismissed):
We also need a change file (a markdown file under the .changes folder) like this:
```
---
"upload": patch
"upload-js": patch
---

Return the upload response as a string and error out if the status code is not within 200-299.
```

lucasfernog:
Tried to push changes to your repo but i'm not allowed to :(

lyager:
@lucasfernog Thank you for your help, there are clearly some corners in the repo that are unknown to me. I've accepted you proposals in the Github UI.

lucasfernog:
@lyager please also push the change file i mentioned above :)

lyager:
Thanks for the patience @lucasfernog , I've rebased and added the covector file you mentioned.

lucasfernog:
idk why but the diff is crazy right now :( something went wrong with your rebase

lyager:
@lucasfernog I've rebased against v1, now only 3 files have changed.. Sorry about that, and thank you for your patience.

FabianLars (approved):

PR #977, opened 2024-02-21
feat(clipboard): Implement HTML and clear functionality

TukanDev:
This PR introduces ability to set and read HTML to/from clipboard, including implementation to trigger clipboard clear.
Currently the added features are locked behind `#[cfg(desktop)]` as i have no ability to test them for mobile targets.


plugins/clipboard-manager/README.md
```diff
@@ -60,9 +60,15 @@ fn main() {
...
-import { writeText, readText } from "@tauri-apps/plugin-clipboard-manager";
+import { writeText, readText, writeHtml, readHtml, clear } from "@tauri-apps/plugin-clipboard-manager";
 await writeText("Tauri is awesome!");
 assert(await readText(), "Tauri is awesome!");
+
+// 2nd parameter is alternative text that is treated as plain text
+await writeHtml('<h1>Tauri is awesome!</h1>', '<h1>Tauri is awesome!</h1>')
+assert(await readHtml(), "<h1>Tauri is awesome!</h1>");
+
+await clear()
```
amrbashir: let's keep the example here simple

plugins/clipboard-manager/guest-js/index.ts
```diff
@@ -11,6 +11,7 @@
 import { invoke } from "@tauri-apps/api/core";
 
 type ClipResponse = Record<"plainText", { text: string }>;
+type ClipHtmlResponse = Record<"html", { html: string, alt_html: string }>;
```
amrbashir: The Rust enum has `rename_all = "camelCase"`
```suggestion
type ClipHtmlResponse = Record<"html", { html: string, altHtml: string }>;
```

plugins/clipboard-manager/guest-js/index.ts
```diff
@@ -53,4 +54,60 @@ async function readText(): Promise<string> {
...
+): Promise<void> {
+  return invoke("plugin:clipboard-manager|write_html", {
+    data: {
+      html: {
+        html,
+        alt_html
+      },
+    },
+  });
+}
```
amrbashir: we should make this optional, also use camelCase for JS APIs
```suggestion
async function writeHtml(
    html: string,
    altHtml?: string
): Promise<void> {
  return invoke("plugin:clipboard-manager|write_html", {
    data: {
      html: {
        html,
        altHtml
      },
    },
  });
}
```

plugins/clipboard-manager/guest-js/index.ts
```diff
@@ -53,4 +54,60 @@ async function readText(): Promise<string> {
...
+/**
+ * Gets the clipboard content as HTML text.
+ * @example
+ * ```typescript
+ * import { readHtml } from '@tauri-apps/plugin-clipboard-manager';
+ * const clipboardHtml = await readHtml();
+ * ```
+ * @since 2.0.0
+ */
+async function readHtml(): Promise<object> {
```
amrbashir: We need a better return type here, seems like it is returning a string

plugins/clipboard-manager/guest-js/index.ts
```diff
@@ -53,4 +54,60 @@ async function readText(): Promise<string> {
...
+  return kind.html
+}
+
+/**
+ * Gets the clipboard content as HTML text.
+ * @example
+ * ```typescript
+ * import { readHtml } from '@tauri-apps/plugin-clipboard-manager';
+ * const clipboardHtml = await readHtml();
+ * ```
```
amrbashir: example is incorrect

plugins/clipboard-manager/src/desktop.rs
```diff
@@ -44,4 +48,35 @@ impl<R: Runtime> Clipboard<R> {
...
+            }
+            Err(e) => Err(crate::Error::Clipboard(e.to_string())),
+        }
+    }
+
+    pub fn clear(&self) -> crate::Result<()> {
+        match &self.clipboard {
+            Ok(clipboard) => {
+                clipboard.lock().unwrap().clear().unwrap();
+                Ok(())
```
amrbashir: We can't unwrap here, need to propagate the error
```suggestion
                clipboard.lock().unwrap().clear().map_err(Into::into)
```

amrbashir (changes requested):
Thanks for your contributions, left a few comments

amrbashir (changes requested):
Thank you, left a few comments

plugins/clipboard-manager/README.md
TukanDev: I will revert README  example to previous one, thought it would be good to provide example showing all possible options.

plugins/clipboard-manager/guest-js/index.ts
TukanDev: In this case i opted to returning entire object as i wanted to provide entire result back, but yea i agree as there is no way in arboard to determine if set_html() fallback is used instead of actual html alt will always be null, since arboard also does not have get_html() as it has for plaintext and images.

plugins/clipboard-manager/guest-js/index.ts
TukanDev: Any pointers to improve it would be good as i honestly cant determine how to improve it, perhaps adding something like assert() or console.log()

plugins/clipboard-manager/guest-js/index.ts
amrbashir: hmmm, we could instead remove this function and rely on `get_text`, I think it is enough

TukanDev: Will remove it right now

plugins/clipboard-manager/guest-js/index.ts
amrbashir: Sorry, I meant the documentation is wrong, this is the clear function and the documentation for readHTML function

TukanDev: marking this as resolved, determined it can be removed.

plugins/clipboard-manager/guest-js/index.ts
TukanDev: Pushed a change stripping everything related to read_html from both sides. Will mark this as resolved.

plugins/clipboard-manager/guest-js/index.ts
```diff
@@ -11,6 +11,8 @@
 import { invoke } from "@tauri-apps/api/core";
 
 type ClipResponse = Record<"plainText", { text: string }>;
+// Commented this type out as it can come in handy to stay
+//type ClipHtmlResponse = Record<"html", { html: string, altHtml: string }>;
```
amrbashir: Let's remove, we can add back if needed

plugins/clipboard-manager/guest-js/index.ts
```diff
@@ -53,4 +55,46 @@ async function readText(): Promise<string> {
...
 
-export { writeText, readText };
+/**
+ * Writes HTML or fallbacks to write provided plain text to the clipboard.
+ * @example
+ * ```typescript
+ * import { writeHtml, readHtml } from '@tauri-apps/plugin-clipboard-manager';
+ * await writeHtml('<h1>Tauri is awesome!</h1>', 'plaintext');
+ * await writeHtml('<h1>Tauri is awesome!</h1>', '<h1>Tauri is awesome</h1>'); // Will write "<h1>Tauri is awesome</h1>" as plain text
+ * assert(await readHtml(), '<h1>Tauri is awesome!</h1>');
```
amrbashir: we should use readText() here 
```suggestion
 * assert(await readText(), '<h1>Tauri is awesome!</h1>');
```

plugins/clipboard-manager/guest-js/index.ts
```diff
@@ -53,4 +55,46 @@ async function readText(): Promise<string> {
...
+  });
+}
+
+/**
+ * Gets the clipboard content as HTML text.
+ * @example
+ * ```typescript
+ * import { readHtml } from '@tauri-apps/plugin-clipboard-manager';
+ * const clipboardHtml = await readHtml();
+ * ```
```
amrbashir: This documentation is still wrong

plugins/clipboard-manager/src/models.rs
```diff
@@ -8,10 +8,12 @@ use serde::{Deserialize, Serialize};
...
 pub enum ClipKind {
     PlainText { label: Option<String>, text: String },
+    Html { html: String, alt_html: Option<String> }
 }
 
 #[derive(Debug, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub enum ClipboardContents {
     PlainText { text: String },
+    Html { html: String, alt_html: Option<String> }
```
amrbashir: This is not needed, let's remove

amrbashir (changes requested):

plugins/clipboard-manager/guest-js/index.ts
TukanDev: Resolved

amrbashir:
Looks like you need to run `pnpm build` and `cargo fmt` to fix formatting and also remove the extra undeeded `read_html` function

plugins/clipboard-manager/src/mobile.rs
```diff
@@ -39,4 +39,17 @@ impl<R: Runtime> Clipboard<R> {
...
+        Err(crate::Error::Clipboard(
+            "Unsupported on this platform".to_string(),
+        ))
+    }
+
+    pub fn read_html(&self) -> crate::Result<ClipboardContents> {
+        Err(crate::Error::Clipboard(
+            "Unsupported on this platform".to_string(),
+        ))
+    }
```
amrbashir: This can be removed

plugins/clipboard-manager/src/mobile.rs
```diff
@@ -39,4 +39,17 @@ impl<R: Runtime> Clipboard<R> {
     pub fn read(&self) -> crate::Result<ClipboardContents> {
         self.0.run_mobile_plugin("read", ()).map_err(Into::into)
     }
+
+    // Treat HTML as unsupported on mobile until tested
+    pub fn write_html(&self) -> crate::Result<ClipboardContents> {
```
amrbashir: To fix mobile CI failing
```suggestion
    pub fn write_html(&self, _kind: ClipKind) -> crate::Result<()> {
```

plugins/clipboard-manager/src/mobile.rs
```diff
@@ -39,4 +39,17 @@ impl<R: Runtime> Clipboard<R> {
     pub fn read(&self) -> crate::Result<ClipboardContents> {
         self.0.run_mobile_plugin("read", ()).map_err(Into::into)
     }
+
+    // Treat HTML as unsupported on mobile until tested
+    pub fn write_html(&self, _kind: ClipKind) -> crate::Result<ClipboardContents> {
```
amrbashir: You also need to change the return type too

```suggestion
    pub fn write_html(&self, _kind: ClipKind) -> crate::Result<()> {
```

plugins/clipboard-manager/src/mobile.rs
TukanDev: Removed

amrbashir (dismissed):

amrbashir (dismissed):

amrbashir (approved):

PR #978, opened 2024-02-22
feat(dialog): allow setting `canCreateDirectories` on macOS

### LINKED ISSUE #949: [feat] "New Folder" option for Mac file dialogs
### Describe the problem

I'm not able to show the native "New Folder" button in Mac file dialogs to let users create a new folder directly in the dialog according to the [docs](https://tauri.studio/v1/api/js/interfaces/dialog.OpenDialogOptions).

I'm building an app where the user has to choose a location for their "project". The project location likely doesn't exist yet, so they'll have to create it. Because they can't create the location inline in the file dialog, they have to jump out of the app, create the directory, come back to the app, and then select it.

In Electron, you can pass the `createDirectory` flag to show a native "New Folder" button in the Mac file dialog ([link to relevant docs](https://www.electronjs.org/docs/latest/api/dialog)).

Seems like this is the cleanest solution, and Tauri could expose this option.

### Alternatives considered

User can always go outside the app to create the directory.

### Additional context

I understand Tauri uses [rfd](https://github.com/PolyMeilex/rfd) for file dialogs. Maybe this needs to be implemented there first? It seems it's not supported at the moment.

Love what you folks are doing with Tauri! 🎉

amrbashir:
closes #949


plugins/dialog/src/commands.rs
```diff
@@ -51,6 +51,8 @@ pub struct OpenDialogOptions {
     #[serde(default)]
     #[cfg_attr(mobile, allow(dead_code))]
     recursive: bool,
+    ///  Whether to allow creating directories in the dialog **macOS Only**
```
FabianLars:
```suggestion
    /// Whether to allow creating directories in the dialog **macOS Only**
```

plugins/dialog/src/commands.rs
```diff
@@ -65,6 +67,8 @@ pub struct SaveDialogOptions {
     filters: Vec<DialogFilter>,
     /// The initial path of the dialog.
     default_path: Option<PathBuf>,
+    ///  Whether to allow creating directories in the dialog **macOS Only**
```
FabianLars:
```suggestion
    /// Whether to allow creating directories in the dialog **macOS Only**
```

plugins/dialog/src/lib.rs
```diff
@@ -345,6 +347,12 @@ impl<R: Runtime> FileDialogBuilder<R> {
         self
     }
 
+    /// Set whether dialog should all creating new directories in the dialog  **macOS only**.
```
FabianLars:
```suggestion
    /// Set whether it should be possible to create new directories in the dialog. Enabled by default. **macOS only**.
```

plugins/dialog/guest-js/index.ts
```diff
@@ -73,6 +75,8 @@ interface SaveDialogOptions {
    * If it's not an existing directory, the file name will be set to the dialog's file name input and the dialog will be set to the parent folder.
    */
   defaultPath?: string;
+  /** Whether to allow creating directories in the dialog **macOS Only** */
```
FabianLars:
```suggestion
  /** Whether to allow creating directories in the dialog. Enabled by default. **macOS Only** */
```

plugins/dialog/guest-js/index.ts
```diff
@@ -55,6 +55,8 @@ interface OpenDialogOptions {
    * Defines whether subdirectories will be allowed on the scope or not.
    */
   recursive?: boolean;
+  /** Whether to allow creating directories in the dialog **macOS Only** */
```
FabianLars:
```suggestion
  /** Whether to allow creating directories in the dialog. Enabled by default. **macOS Only** */
```

FabianLars (approved):

PR #983, opened 2024-02-23
fix(http): allow `User-Agent` header to be set

### LINKED ISSUE #966: [http][v2] not able to pass `User-Agent` in the fetch header
It seems that the headers param would accept any key value pair but simply ignored `User-Agent`
```
const response = await fetch("some_url", {
    headers: {
        "User-Agent": "some_string",
    },
    method: "get",
});
```
Afterwards, the user-agent would always be set to `tauri` by:
```
if !headers.contains_key(header::USER_AGENT.as_str()) {
    request = request.header(header::USER_AGENT, HeaderValue::from_static("tauri"));
}
```

amrbashir:
closes #966


FabianLars (approved):
I wonder if the forbidden headers killed some usecases of the plugin, mainly the Origin one? 🤔

amrbashir:
All of these were forbidden before by the browser so we aren't breaking anything

PR #993, opened 2024-02-25
feat(deep-link): Add more AssociatedDomain attributes

m00nwtchr:
I needed to use the `pathPattern` attribute, so I added it.

Also added `scheme` to allow specifying `https` only. (+ other possible schemes)


plugins/deep-link/src/config.rs
```diff
@@ -10,10 +10,20 @@ use serde::{Deserialize, Deserializer};
 
 #[derive(Deserialize)]
 pub struct AssociatedDomain {
+    #[serde(default)]
```
lucasfernog: shouldn't we default this to at least `https` (and maybe `http` too at least on debug)?

FabianLars:
will look into ios this week-ish (even though it doesn't reallyyy/properly support custom schemes anymore)

FabianLars:
nevermind, forgot that we don't have a way to edit the ios info.plist from inside a plugin yet(?)

FabianLars (approved):

PR #996, opened 2024-02-25
docs: improve `tauri-plugin-store` readme and code example

Rigidity:
This brings `tauri-plugin-store`'s readme up to date with the the latest v2 beta, improves the wording of a few things, and adds a note about preventing crashes while loading from the Rust side of things.

### DIFF
--- a/plugins/store/README.md
+++ b/plugins/store/README.md
@@ -105,9 +105,7 @@ fn main() {
             let mut store = StoreBuilder::new("app_data.bin").build(app.handle().clone());
 
             // Attempt to load the store, if it's saved already.
-            if let Err(error) = store.load() {
-                println!("Failed to load store from disk: {error}.");
-            }
+            store.load().expect("Failed to load store from disk");
 
             // Note that values must be serde_json::Value instances,
             // otherwise, they will not be compatible with the JavaScript bindings.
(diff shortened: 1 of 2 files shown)


plugins/store/README.md
```diff
@@ -69,19 +69,30 @@ await store.set("some-key", { value: 5 });
...
-1. The application is closed gracefully (plugin automatically saves)
-2. The store is manually saved (using `store.save()`)
+You can manually save a store with:
+
+```javascript
+await store.save();
+```
+
+Stores are loaded automatically on creation when using the JavaScript bindings.  
+However, you can also load them manually later like so:
```
amrbashir: Stores are not loaded on creation but rather when using any method on it.

plugins/store/README.md
```diff
@@ -91,23 +102,70 @@ fn main() {
...
+
+```rust
+if let Err(error) = store.load() {
+    if let tauri_plugin_store::Error::NotFound(path) = error {
+        println!("File not found for store: {path:?}");
+    } else {
+        return Err(error)?;
+    }
+}
+```
```
amrbashir: We could probably skip this part, don't think it is that useful

amrbashir (changes requested):

Rigidity:
I'll finish this up today - sorry for the delay. Been busy with work

simonhyll:
I'd say this is an improvement from a docs point of view and we have lots of users begging for better `store` docs. Imo we should merge this and potentially improve upon it later on, as long as the information is at least sufficiently accurate.

simonhyll (approved):

amrbashir (approved):

PR #1001, opened 2024-02-28
Bump tauri to v2 beta.7

ahqsoftwares:
The following pr upgrades the lugis to support tauri beta.7


amrbashir:
@ahqsoftwares could you also add a change file so we can release it as well?

PR #1004, opened 2024-02-29
feat: add content-length on POST

lyager:
Not all embedded devices are acceptable to receiving unspecified amounts of data. Appending the content-length up front helps this devices succeed.


lyager:
Closed.. Duplicate

PR #1006, opened 2024-02-29
feat(http): Enable cookie store

m00nwtchr:


m00nwtchr:
This works decently for storing cookies within an app session, any ideas on how I could persist them though?

m00nwtchr:
So this works for storing cookies for the current app session, would using the `store` plugin to persist them work? (And should I try to integrate that into the `http` plugin?)

FabianLars:
Sorry for the long silence here. I think this PR went under the radar and was superseeded by https://github.com/tauri-apps/plugins-workspace/pull/1978 :/

PR #1011, opened 2024-03-02
feat(store): Add android & iOS support

elkhoudiry:
Add support for android & iOS in a sandboxed and application's specific locations that requires no permissions


plugins/store/android/build.gradle.kts
```diff
@@ -0,0 +1,43 @@
...
+    id("com.android.library")
+    id("org.jetbrains.kotlin.android")
+}
+
+android {
+    namespace = "app.tauri.store"
+    compileSdk = 33
+
+    defaultConfig {
+        minSdk = 24
```
amrbashir: we could probably target 21 or 20, right?

plugins/store/android/src/androidTest/java/ExampleInstrumentedTest.kt
```diff
@@ -0,0 +1,28 @@
+// Copyright 2019-2023 Tauri Programme within The Commons Conservancy
```
amrbashir: let's remove these test files along with dependencies related to them, they are not useful in their current state

plugins/store/android/src/main/java/StorePlugin.kt
```diff
@@ -0,0 +1,52 @@
...
+    fun load(invoke: Invoke) {
+        try {
+            val path = invoke.parseArgs(String::class.java)
+            val file = File(activity.applicationContext.getExternalFilesDir(null), path)
+
+            if (!file.exists()) {
+                file.parentFile?.mkdirs()
+                file.createNewFile()
+                file.writeText("{}")
+            }
```
amrbashir: if file doesn't exist we should return with an error instead of creating the file

plugins/store/ios/Sources/StorePlugin.swift
```diff
@@ -0,0 +1,196 @@
...
+                    appropriateFor: nil,
+                    create: true
+                )
+                .appendingPathComponent(path)
+            if !FileManager.default.fileExists(atPath: fileURL.path) {
+                let x = try FileManager.default.createFile(
+                    atPath: fileURL.path,
+                    contents: "{}".data(using: String.Encoding.utf8)
+                )
+            }
```
amrbashir: same as android, if file doesn't exist, return with an error

plugins/store/ios/Tests/PluginTests/PluginTests.swift
```diff
@@ -0,0 +1,12 @@
+// Copyright 2019-2023 Tauri Programme within The Commons Conservancy
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-License-Identifier: MIT
+
+import XCTest
+@testable import StorePlugin
+
```
amrbashir: let's remove these test files as awell

plugins/store/src/mobile.rs
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    pub fn load(&mut self) -> Result<()> {
+        let result: Value = self
+            .mobile_plugin_handle
+            .as_ref()
+            .ok_or_else(|| crate::error::Error::MobilePluginHandleUnInitialized)?
+            .run_mobile_plugin("load", self.path.to_string_lossy().to_string())?;
+
+        let map = serde_json::from_value::<HashMap<String, Value>>(result).unwrap();
```
amrbashir:
```suggestion
        let map = serde_json::from_value::<HashMap<String, Value>>(result)?;
```

amrbashir (changes requested):

amrbashir:
Thanks for contribution

elkhoudiry:
Thank you, all comments should be resolved now

plugins/store/ios/Sources/StorePlugin.swift
```diff
@@ -0,0 +1,217 @@
...
+func initPlugin() -> Plugin {
+    return StorePlugin()
+}
+
+private extension URL {
+    func appendPath(path: String) -> URL {
+        if #available(iOS 16.0, *) {
+            return self.appending(path: path)
+        } else {
+            return self.appendingPathComponent(path)
```
amrbashir: We could avoid unnecessary I/O operation to check if appended component is a directory or not by specifying a hint https://developer.apple.com/documentation/foundation/url/3988447-appending so we should change this extension method to take a second parameter (maybe optional) to specify whether we are adding a directory component or not.

plugins/store/ios/Sources/StorePlugin.swift
```diff
@@ -0,0 +1,217 @@
...
+                url = url.appendPath(path: components[i])
+            }
+            
+            if components.count > 1 {
+                try FileManager.default.createDirectory(at: url, withIntermediateDirectories: true)
+            }
+            
+            url = url.appendPath(path: components.last!)
+            
+            return url
```
amrbashir, line 83: can't this be simplified to

```suggestion
            let components = path.split(separator: "/").map { element in String(element) }
            
            for i in 0..<components.count {
                if i == components.count - 1 {
                    try FileManager.default.createDirectory(at: url, withIntermediateDirectories: true)
                }
            
                url = url.appendPath(path: components[i])
            }
            
            return url
```

amrbashir (changes requested):

plugins/store/ios/Sources/StorePlugin.swift
```diff
@@ -0,0 +1,217 @@
...
+            invoke.resolve()
+        } catch {
+            invoke.reject(error.localizedDescription)
+        }
+    }
+    
+    @objc public func load(_ invoke: Invoke) throws {
+        do {
+            let path = try invoke.parseArgs(String.self)
+            let fileURL = getUrlFromPath(path: path)
```
amrbashir: this will create intermediate directories on the load function where it shouldn't, maybe `getUrlFromPath` should be changed to take a second parameter to indicate when to create directories?

amrbashir (dismissed):

amrbashir (approved):

Legend-Master:
`with_store` and `StoreCollection` are now private, is this intended?

PR #1019, opened 2024-03-04
fix: single instance doesn't shutdown immediately

### LINKED ISSUE #1017: [single-instance] doesn't shutdown immediately on Windows
When using single instance plugin on Windows, the second instance doesn't shutdown immediately

Windows implementation uses `app.exit` while the Linux one uses `std::process::exit`, `app.exit` only sends a shutdown signal, and the app will still be running through all the following setup functions, e.g. create the window, create the system tray...

This was changed to this in 4d903f3533f701441e06e83c4b666f1eef035d34 for cleaning up system tray, but currently it doesn't even trigger exit event for some reason (even if it actually triggers it, it's still problematic to have flashing windows)

In my opinion, we should just use `std::process::exit` here (maybe run with `cleanup_before_exit` to clean up system tray) (simple fix but may not be perfect) or we'll need a way to shutdown the app cleanly in setup step from Tauri

https://github.com/tauri-apps/plugins-workspace/blob/99bea2559c2c0648c2519c50a18cd124dacef57b/plugins/single-instance/src/platform_impl/windows.rs#L65

https://github.com/tauri-apps/plugins-workspace/blob/99bea2559c2c0648c2519c50a18cd124dacef57b/plugins/single-instance/src/platform_impl/linux.rs#L73

Legend-Master:
Fix #1017

### DIFF
--- a/.changes/fix-shutdown-timing.md
+++ b/.changes/fix-shutdown-timing.md
@@ -1,0 +1,5 @@
+---
+'single-instance': patch
+---
+
+Fix doesn't shutdown immediately.
--- a/plugins/single-instance/src/platform_impl/linux.rs
+++ b/plugins/single-instance/src/platform_impl/linux.rs
@@ -70,7 +70,8 @@ pub fn init<R: Runtime>(f: Box<SingleInstanceCallback<R>>) -> TauriPlugin<R> {
                             ),
                         );
                     }
-                    std::process::exit(0)
+                    app.cleanup_before_exit();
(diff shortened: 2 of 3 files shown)


FabianLars:
Thank you! Could you also add a changefile like this https://github.com/tauri-apps/plugins-workspace/blob/v2/.changes/fix-zbus-import.md? Then it'd be good to go :)

FabianLars (approved):
thank you :)

PR #1029, opened 2024-03-05
fix: fix `plugins/fs` infinite loop on autoregen

otamam818:
The auto-regen from `plugins/fs/build.rs` would keep generating on compile time. Upon check-up, it made more sense to just make sure it doesn't regenerate files redundantly, so I implemented a check to see if the files are redundant or not.

Please let me know if I missed something


FabianLars:
Thanks for contributing!

To merge this we'll need a changefile similar to https://github.com/tauri-apps/plugins-workspace/blob/v2/.changes/fix-zbus-import.md

Also, the if statement is a bit incorrect. As it is, it won't write the file if it doesn't exist already.

otamam818:
> Also, the if statement is a bit incorrect. As it is, it won't write the file if it doesn't exist already.

Thanks @FabianLars for pointing that out, I appreciate it. I see it now. Give me a moment, I'll update this PR and address that

> To merge this we'll need a changefile similar to https://github.com/tauri-apps/plugins-workspace/blob/v2/.changes/fix-zbus-import.md

I'll also add a file to `.changes` called `file-autogen-fix.md`

plugins/fs/build.rs
```diff
@@ -154,8 +154,27 @@ permissions = [
...
+                // The file has changed, it should be updated
+                std::fs::write(&target_file, toml)
+                    .unwrap_or_else(|e| panic!("unable to autogenerate ${upper}: {e}"));
+            }
+        } else {
+            // If it doesn't exist we can just create a new file and write to
+            // it
+            std::fs::write(&target_file, toml)
+                .unwrap_or_else(|e| panic!("unable to autogenerate ${upper}: {e}"));
+        }
```
lucasfernog:
```suggestion
        let permission_path = base_dirs.join(format!("{lower}.toml"));
        if toml != std::fs::read_to_string(&permission_path).unwrap_or_default() {
            std::fs::write(permission_path, toml)
                .unwrap_or_else(|e| panic!("unable to autogenerate ${lower}: {e}"));
        }
```

lucasfernog (changes requested):
this is the way we handle this on the tauri side

amrbashir (approved):

PR #1033, opened 2024-03-06
fix(dialog): Remove Linux specific code for rfd 0.14 compatibility

### LINKED ISSUE #956: [dialog][v2][linux] Any blocking dialog does not display properly when running on Linux
## description

The code example below is from the Rust side, but it hangs the same way from the frontend because the blocking dialog is invoked via Command. However, this seems to be a problem that `invoke_with_priority` may not be able to aquire `MainContext`, so it may work in some cases.

Note: I think tauri v1 also has the same problem.

- ref: https://github.com/tauri-apps/plugins-workspace/blob/531123cad0a381eb3e1b9cd98b2901ed4b6aa35f/plugins/dialog/src/desktop.rs#L61-L69

### What I did

```console
$ npm create tauri-app@latest -- --alpha
warning: The `--alpha` option is now an alias for `--beta` and may be removed in the future.
✔ Project name · tauri-v2-exp2
✔ Choose which language to use for your frontend · TypeScript / JavaScript - (pnpm, yarn, npm, bun)
✔ Choose your package manager · npm
✔ Choose your UI template · React - (https://react.dev/)
✔ Choose your UI flavor · TypeScript
✔ Would you like to setup the project for mobile as well? · no

Template created! To get started run:
  cd tauri-v2-exp2
  npm install
  npm run tauri dev

$ cd tauri-v2-exp2
$ npm i
$ npm run tauri add dialog
$ vim src-tauri/src/main.rs
$ npm run tauri dev
```

```rust
// src-tauri/src/main.rs
use tauri_plugin_dialog::DialogExt;

#[tauri::command]
fn greet<R: tauri::Runtime>(app: tauri::AppHandle<R>, name: &str) -> String {
    // The dialog does not appear and this program is stuck on next line.
    app.dialog().message("hello").blocking_show();

    format!("Hello, {}! You've been greeted from Rust!", name)
}

fn main() {
    tauri::Builder::default()
        .plugin(tauri_plugin_dialog::init())
        .plugin(tauri_plugin_shell::init())
        .invoke_handler(tauri::generate_handler![greet])
}
```

FabianLars:
fixes #956

After https://github.com/PolyMeilex/rfd/pull/152 it always freezes the app since we already have a gtk main loop running and try to run the dialog on it too. Since gtk allows to have multiple gtk event loops in the same app, as long as they run in seperate threads, i removed our special handling and let rfd spawn its own loop (see pr linked above). From my testing this works fine and allows us to keep upgrading rfd instead of locking it to 0.12(?).

### DIFF
--- a/.changes/dialog-linux-freeze.md
+++ b/.changes/dialog-linux-freeze.md
@@ -1,0 +1,5 @@
+---
+"dialog": "patch"
+---
+
+Fixed an issue where dialogs would not spawn but instead freeze the whole app.
--- a/plugins/dialog/src/desktop.rs
+++ b/plugins/dialog/src/desktop.rs
@@ -11,22 +11,13 @@
 use std::path::PathBuf;
 
 use raw_window_handle::{HasWindowHandle, RawWindowHandle};
+use rfd::{AsyncFileDialog, AsyncMessageDialog};
 use serde::de::DeserializeOwned;
 use tauri::{plugin::PluginApi, AppHandle, Runtime};
 
 use crate::{models::*, FileDialogBuilder, MessageDialogBuilder};
 
 const OK: &str = "Ok";
-
-#[cfg(target_os = "linux")]
-type FileDialog = rfd::FileDialog;
-#[cfg(not(target_os = "linux"))]
-type FileDialog = rfd::AsyncFileDialog;
-
-#[cfg(target_os = "linux")]
-type MessageDialog = rfd::MessageDialog;
-#[cfg(not(target_os = "linux"))]
-type MessageDialog = rfd::AsyncMessageDialog;
 
 pub fn init<R: Runtime, C: DeserializeOwned>(
     app: &AppHandle<R>,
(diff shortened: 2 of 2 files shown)


amrbashir (approved):

valadaptive:
Hi,

I just stumbled upon this. It seems my PR to rfd has created a lot of unintended effects for those who use it, and I'm curious where exactly GTK documents that it's safe to run multiple event loops in the same app.

The quote in https://github.com/PolyMeilex/rfd/pull/152 comes from documentation that's no longer online, but I can't seem to find any other documentation that says that GTK *can* be run on multiple threads as long as they don't interact. If you could point me to such documentation, I may be able to create another PR to rfd.

FabianLars:
I can't remember if i based this on gtk docs or just on forum discussions ([example](https://discourse.gnome.org/t/glib-multiple-gmaincontexts/12872), just for glib though) and our own testing tbh :/
I do remember reading though [gmaincontext](https://developer.gnome.org/documentation/tutorials/main-contexts.html) / [glib loop](https://docs.gtk.org/glib/main-loop.html) docs too (which of course does not guarantee that gtk behaves the same), but those aren't really explicit about it either, for example:
> To allow multiple independent sets of sources to be handled in different threads, each source is associated with a GMainContext.

In the 3 months since this was merged this did not cause a single issue that we're aware of so we're fairly confident.

PR #1035, opened 2024-03-06
feat(single-instance): add MacOs unix domain socket impl

WitoDelnat:
Hi! My Tauri project needs single-instance support on macOS as it'll mainly launch through the CLI. Last week during my research I ran into [this PR](https://github.com/tauri-apps/plugins-workspace/pull/749). Since back then little movement had happened, I initially copied the solution to tweak it to v2, but FIFO files were giving some problems for my (rather complex) use case. I decided to rewrite it using unix domain sockets.

I now read [this comment](https://github.com/tauri-apps/plugins-workspace/pull/749#issuecomment-1969964906) and wanted to share my solution as a possible alternative. You can see it in action in the video below.

https://github.com/tauri-apps/plugins-workspace/assets/7761005/52b57523-3675-47d5-8bd9-1d704d3c9bee

### DIFF
--- a/plugins/single-instance/examples/vanilla/src-tauri/src/main.rs
+++ b/plugins/single-instance/examples/vanilla/src-tauri/src/main.rs
@@ -9,6 +9,7 @@
 
 fn main() {
   tauri::Builder::default()
+    .plugin(tauri_plugin_cli::init())
     .plugin(tauri_plugin_single_instance::init(|app, argv, cwd| {
       println!("{}, {argv:?}, {cwd}", app.package_info().name);
     }))
--- a/plugins/single-instance/src/platform_impl/macos.rs
+++ b/plugins/single-instance/src/platform_impl/macos.rs
@@ -4,13 +4,164 @@
 
 #![cfg(target_os = "macos")]
 
+use std::{
+    io::{BufWriter, Read, Write},
+    os::unix::net::{UnixListener, UnixStream},
+    path::PathBuf,
+};
+
 use crate::SingleInstanceCallback;
 use tauri::{
     plugin::{self, TauriPlugin},
-    Manager, Runtime,
+    AppHandle, Config, Manager, RunEvent, Runtime,
 };
-pub fn init<R: Runtime>(_f: Box<SingleInstanceCallback<R>>) -> TauriPlugin<R> {
-    plugin::Builder::new("single-instance").build()
+
+enum ErrorKind {
+    Retry,
+    Abort,
 }
 
-pub fn destroy<R: Runtime, M: Manager<R>>(_manager: &M) {}
+pub fn init<R: Runtime>(cb: Box<SingleInstanceCallback<R>>) -> TauriPlugin<R> {
+    plugin::Builder::new("single-instance")
+        .setup(|app, _api| {
+            start(app, cb, false);
+            Ok(())
+        })
+        .on_event(|app, event| {
+            if let RunEvent::Exit = event {
+                destroy(app);
+            }
+        })
+        .build()
+}
+
+pub fn destroy<R: Runtime, M: Manager<R>>(manager: &M) {
+    let path = socket_path(manager.config());
+    socket_cleanup(path);
+}
+
+fn start<R: Runtime>(app: &AppHandle<R>, cb: Box<SingleInstanceCallback<R>>, retry: bool) {
+    let socket = socket_path(app.config());
+
+    match UnixListener::bind(&socket) {
+        Ok(listener) => {
+            // This process is the singleton.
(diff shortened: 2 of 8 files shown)


plugins/single-instance/src/platform_impl/macos.rs
```diff
@@ -4,13 +4,164 @@
...
+                }
+            }
+        }
+    }
+}
+
+fn socket_path(config: &Config) -> PathBuf {
+    let identifier = config.identifier.replace(['.', '-'].as_ref(), "_");
+    // Use /tmp as socket path must be shorter than 100 chars.
+    PathBuf::from(format!("/tmp/{}_si.sock", identifier))
```
WitoDelnat: Some input here would be appreciated on the usage of the `/tmp` directory. The character limit on socket files prevents usage of Tauri's suggested directories. I also tried `/var/run` as that's where docker.sock is located but that also gives errors.

plugins/single-instance/examples/vanilla/src-tauri/src/main.rs
```diff
@@ -9,6 +9,7 @@
 
 fn main() {
   tauri::Builder::default()
+    .plugin(tauri_plugin_cli::init())
```
WitoDelnat, line 12: I believe we'll likely want to remove this last commit which modified some files in the example. Including the CLI was the main (only?) way to test this on MacOS and I wanted to make it easy for you folks to try it out yourself. I can remove it when you are ready for it, let me know about how you'd like to tackle this.

plugins/single-instance/src/platform_impl/macos.rs
FabianLars: /tmp is fine afaik, that's what i used in other tauri projects too and it didn't cause any issues.

plugins/single-instance/examples/vanilla/src-tauri/src/main.rs
FabianLars: as far as i know this is not required to start the app from the console. If it is then imo it's fine to keep the cli plugin in the example.

plugins/single-instance/src/platform_impl/macos.rs
```diff
@@ -4,13 +4,164 @@
...
+            if let Err(_e) = bf.flush() {
+                return Err(ErrorKind::Abort);
+            }
+
+            drop(bf);
+            Ok(())
+        }
+        Err(e) => {
+            match e.raw_os_error() {
+                Some(61) => {
```
FabianLars: is there any advantage over just checking if it's `ErrorKind::ConnectionRefused`?

plugins/single-instance/src/platform_impl/macos.rs
```diff
@@ -4,13 +4,164 @@
...
+        Ok(stream) => {
+            let mut bf = BufWriter::new(&stream);
+            let args_joined = std::env::args().collect::<Vec<String>>().join("\0");
+
+            if let Err(_e) = bf.write_all(args_joined.as_bytes()) {
+                return Err(ErrorKind::Abort);
+            }
+            if let Err(_e) = bf.flush() {
+                return Err(ErrorKind::Abort);
+            }
```
FabianLars: hmm, i feel like `.map_err(|_| ErrorKind::Abort)?` may look a bit cleaner. What do you think?

FabianLars:
Thanks for contributing :)

Did you ever try to change the order to first try to bin the UnixStream? I've been doing that and couldn't spot any issues though my use of unix sockets on macos wasn't tested that much... It would simplify the implementation quite a bit as you could get rid of the whole retry logic.

WitoDelnat:
Thanks for the fast feedback! I agree with your suggestion to simplify. I'll try it out to see if it works equally well. I'm a bit occupied right now but I'll look into addressing your input somewhere over the coming week.

FabianLars:
Thank you! On first glance (i'll do a proper review later, busy day) the only thing left to do is to replace println/eprintln with log::debug since most of them are part of normal operation. Maybe use log::warn and/or log::error for those that aren't, like the `single_instance failed to listen to other processes - launching normally: {}` one.

Oh and we'll also need a changefile like https://github.com/tauri-apps/plugins-workspace/blob/v2/.changes/fix-shutdown-timing.md :)

WitoDelnat:
Hey again! I just found some time to wrap up this PR. I pushed the log changes and added a changefile. Please check it out and let me know if I can somehow help.

ps: I'm uncertain if it should be minor or patch. All other changefiles are patches so I just kept in line with that. Let me know if you'd like to see it differently and I can update it.

FabianLars (approved):
thank you!!

PR #1036, opened 2024-03-06
tauri v2.0.0-beta.9 http compatibilty

Dreaming-Codes:


amrbashir:
Thank you, this is being fixed for all plugins in #1037

PR #1039, opened 2024-03-07
ci: Move cargo target dir to /mnt

FabianLars:
Currently our publishing workflow fails _twice_ because it runs out of disk space. https://github.com/tauri-apps/plugins-workspace/actions/runs/8182433330/job/22373757645#step:8:8499

A month ago github changed the azure vms for public repos which resulted in a size decrease of the root partition by ~10gb (though i kinda doubt this is the sole reason because it doesn't fail once, but twice). At the same time the /mnt partition was increased to a whopping 75gb, over 60gb of it unused, since the root partition only has ~20gb free space we'll let rust save its artifact into /mnt hoping that we're actually dealing with space issues and not something else...


FabianLars:
well, let's just try it out with a single change release to see if covector can handle it (though i think it doesn't even care)

PR #1040, opened 2024-03-07
fix(store): `with_store` and `StoreCollection` become private in #1011

Legend-Master:
Fix `with_store` and `StoreCollection` changed to private in #1011


FabianLars (approved):
Thank you! I'll trigger the release pipeline asap.

PR #1051, opened 2024-03-08
fix(updater): fallback to passive mode & fix `installerArgs` deserialzation

### LINKED ISSUE #1015: [updater] install mode is not passive by default if windows field is not declared
`Config::windows` is default to `None` instead of `Some(WindowsConfig::default())`

https://github.com/tauri-apps/plugins-workspace/blob/99bea2559c2c0648c2519c50a18cd124dacef57b/plugins/updater/src/config.rs#L84

https://github.com/tauri-apps/plugins-workspace/blob/99bea2559c2c0648c2519c50a18cd124dacef57b/plugins/updater/src/updater.rs#L557-L574

Legend-Master:
Fix #1015

Also fixes `installerArgs` not working in tauri config files

### DIFF
--- a/plugins/updater/src/config.rs
+++ b/plugins/updater/src/config.rs
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-License-Identifier: MIT
 
-use std::{ffi::OsString, fmt::Display};
+use std::fmt::Display;
 
 use serde::{Deserialize, Deserializer};
 use url::Url;
@@ -64,12 +64,8 @@ impl Default for WindowsUpdateInstallMode {
 #[serde(rename_all = "camelCase")]
 pub struct WindowsConfig {
     /// Additional arguments given to the NSIS or WiX installer.
-    #[serde(
-        default,
-        alias = "installer-args",
-        deserialize_with = "deserialize_os_string"
-    )]
-    pub installer_args: Vec<OsString>,
+    #[serde(default, alias = "installer-args")]
+    pub installer_args: Vec<String>,
     /// Updating mode, see [`WindowsUpdateInstallMode`] for more info.
     #[serde(default, alias = "install-mode")]
     pub install_mode: WindowsUpdateInstallMode,
@@ -85,8 +81,7 @@ pub struct Config {
     /// Signature public key.
     pub pubkey: String,
     /// The Windows configuration for the updater.
-    #[serde(default)]
-    pub windows: WindowsConfig,
+    pub windows: Option<WindowsConfig>,
 }
 
 /// A URL to an updater server.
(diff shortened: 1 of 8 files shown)


plugins/updater/src/config.rs
```diff
@@ -113,3 +118,13 @@ impl<'de> Deserialize<'de> for UpdaterEndpoint {
...
+
+fn deserialize_os_string<'de, D>(deserializer: D) -> Result<Vec<OsString>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    Ok(Vec::<String>::deserialize(deserializer)?
+        .into_iter()
+        .map(OsString::from)
+        .collect::<Vec<_>>())
+}
```
amrbashir: This is effectively the same as deserializing a `String`, so let's just change `installer_args` field type to `Vec<String>` and remove the custom deserialization.

plugins/updater/src/config.rs
```diff
@@ -81,7 +85,8 @@ pub struct Config {
     /// Signature public key.
     pub pubkey: String,
     /// The Windows configuration for the updater.
-    pub windows: Option<WindowsConfig>,
+    #[serde(default)]
+    pub windows: WindowsConfig,
```
amrbashir: I would rather keep this as `None` and just apply a default installation mode if it was `None`

amrbashir (changes requested):

plugins/updater/src/config.rs
Legend-Master: If we change this to `String`, do we want to also change the one from Builder to also be `String`?

plugins/updater/src/config.rs
Legend-Master: Ok, I just thought this is cleaner for the consumers so that they can handle 1 less `Option`

plugins/updater/src/config.rs
Legend-Master: Also, I would like to open another PR to migrate current powershell implementation to `ShellExecute` to avoid flushing powershell window, in that case I think `String` type makes more sense as we will convert it though `HSTRING::from` anyway

amrbashir: > If we change this to `String`, do we want to also change the one from Builder to also be `String`?

yeah let's do it



> Also, I would like to open another PR to migrate current powershell implementation to `ShellExecute` to avoid flushing powershell window, in that case I think `String` type makes more sense as we will convert it though `HSTRING::from` anyway

`HSTRING` can convert from `OsString` as well but that's beside the point, `ShellExecute` sounds good to me.

plugins/updater/src/config.rs
amrbashir: I prefer to write `None` that `WindowsConfig::default` when using the `Config` struct from rust.

Legend-Master:
Not sure why it can't pick up `installer_args` if `windows` field is `Option<WindowsConfig>`, `WindowsConfig` works though

Legend-Master:
Strange, `cargo test --bin app-updater` on my machine succeed

Legend-Master:
Found the problem, this part is still handled from tauri itself, not sure why changing `installer_args` to be not an `Option` fixes the problem

https://github.com/tauri-apps/tauri/blob/45384ab3aa44687f25323ebf367058270ec3a4d3/tooling/cli/src/interface/rust.rs#L780-L783

amrbashir (approved):
Sorry for the later reply, I have decided to revert back to `OsString` and go with your original solution, deserialize from string first.

Thanks for contribution

PR #1053, opened 2024-03-08
feat(fs): enhance scope config to accept string as well

amrbashir:


plugins/fs/src/lib.rs
```diff
@@ -48,25 +43,6 @@ impl<R: Runtime, T: Manager<R>> FsExt<R> for T {
...
-    ) -> std::result::Result<Self, Self::Error> {
-        #[derive(Deserialize)]
-        struct EntryRaw {
-            path: PathBuf,
-        }
-
-        let entry = serde_json::from_value::<EntryRaw>(raw.into())?;
-
-        Ok(Self {
-            path: app.path().parse(entry.path)?,
```
lucasfernog, line 65: you removed this bit :) gonna push it again

lucasfernog (approved):

plugins/fs/src/lib.rs
amrbashir: thanks, wasn't sure why deserialization was done this way and didn't notice the resolve part.

PR #1054, opened 2024-03-08
refactor!(updater): migrate run updater using powershell to `ShellExecute`

Legend-Master:
Fix flashing powershell window on update

### DIFF
--- a/plugins/updater/Cargo.toml
+++ b/plugins/updater/Cargo.toml
@@ -9,11 +9,11 @@ rust-version = { workspace = true }
 links = "tauri-plugin-updater"
 
 [package.metadata.docs.rs]
-rustc-args = [ "--cfg", "docsrs" ]
-rustdoc-args = [ "--cfg", "docsrs" ]
+rustc-args = ["--cfg", "docsrs"]
+rustdoc-args = ["--cfg", "docsrs"]
 
 [build-dependencies]
-tauri-plugin = { workspace = true, features = [ "build" ] }
+tauri-plugin = { workspace = true, features = ["build"] }
 
 [dependencies]
 tauri = { workspace = true }
@@ -21,15 +21,18 @@ serde = { workspace = true }
 serde_json = { workspace = true }
 thiserror = { workspace = true }
 tokio = "1"
-reqwest = { version = "0.11", default-features = false, features = [ "json", "stream" ] }
+reqwest = { version = "0.11", default-features = false, features = [
+    "json",
+    "stream",
+] }
 url = { workspace = true }
 http = "0.2"
 dirs-next = "2"
 minisign-verify = "0.2"
-time = { version = "0.3", features = [ "parsing", "formatting" ] }
+time = { version = "0.3", features = ["parsing", "formatting"] }
 base64 = "0.21"
 percent-encoding = "2"
-semver = { version = "1", features = [ "serde" ] }
+semver = { version = "1", features = ["serde"] }
 futures-util = "0.3"
 tempfile = "3"
 zip = "0.6"
@@ -37,6 +40,11 @@ tar = "0.4"
 
 [target."cfg(target_os = \"windows\")".dependencies]
 zip = { version = "0.6", default-features = false }
+windows = { version = "0.54", features = [
+    "Win32_System_WinRT",
+    "Win32_UI_Shell",
+    "Win32_UI_WindowsAndMessaging",
+] }
 
 [target."cfg(any(target_os = \"macos\", target_os = \"linux\"))".dependencies]
 flate2 = "1.0.27"
@@ -45,6 +53,6 @@ flate2 = "1.0.27"
 mockito = "0.31"
 
 [features]
-native-tls = [ "reqwest/native-tls" ]
-native-tls-vendored = [ "reqwest/native-tls-vendored" ]
-rustls-tls = [ "reqwest/rustls-tls" ]
+native-tls = ["reqwest/native-tls"]
+native-tls-vendored = ["reqwest/native-tls-vendored"]
+rustls-tls = ["reqwest/rustls-tls"]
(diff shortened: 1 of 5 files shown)


Legend-Master:
Note: I have not tested this yet

plugins/updater/Cargo.toml
```diff
@@ -9,34 +9,42 @@ rust-version = { workspace = true }
 links = "tauri-plugin-updater"
 
 [package.metadata.docs.rs]
-rustc-args = [ "--cfg", "docsrs" ]
-rustdoc-args = [ "--cfg", "docsrs" ]
+rustc-args = ["--cfg", "docsrs"]
+rustdoc-args = ["--cfg", "docsrs"]
```
amrbashir: let's revert all of these formatting changes

plugins/updater/Cargo.toml
```diff
@@ -9,34 +9,42 @@ rust-version = { workspace = true }
...
-semver = { version = "1", features = [ "serde" ] }
+semver = { version = "1", features = ["serde"] }
 futures-util = "0.3"
 tempfile = "3"
 zip = "0.6"
 tar = "0.4"
 
 [target."cfg(target_os = \"windows\")".dependencies]
 zip = { version = "0.6", default-features = false }
+windows = { version = "0.54", features = [
```
amrbashir: I think we should use `windows-sys` instead for faster compile times and easier maintenance since it is not frequently updated.

plugins/updater/Cargo.toml
```diff
@@ -9,34 +9,42 @@ rust-version = { workspace = true }
...
+    "Win32_System_WinRT",
```
amrbashir: this feature shouldn't be needed

plugins/updater/src/updater.rs
```diff
@@ -514,6 +514,13 @@ impl Update {
     fn install_inner(&self, bytes: Vec<u8>) -> Result<()> {
         use std::{fs, process::Command};
 
+        use windows::{
+            core::{w, HSTRING},
+            Win32::UI::{Shell::ShellExecuteW, WindowsAndMessaging::SW_SHOW},
+        };
+
+        use crate::config::WindowsUpdateInstallMode;
```
amrbashir: unused import

plugins/updater/src/updater.rs
```diff
@@ -534,118 +541,50 @@ impl Update {
...
-                std::process::exit(0);
+                );
+                installer_args.push(OsStr::new("/promptrestart"));
+            } else {
+                continue;
+            }
+            let file = HSTRING::from(found_path.as_os_str());
+            let parameters = HSTRING::from(installer_args.join(OsStr::new(" ")));
+            let ret =
+                unsafe { ShellExecuteW(None, w!("runas"), &file, &parameters, None, SW_SHOW) };
```
amrbashir: should probably use `open` instead of `runas`

amrbashir (changes requested):
Thank you, left a few comments for now and will test later

plugins/updater/Cargo.toml
Legend-Master: `HSTRING` is from `Win32_System_WinRT` feature

plugins/updater/src/updater.rs
Legend-Master: It was `runas` in the powershell version, just migrated as is, I'm not sure should we run it elevated by default or not

plugins/updater/Cargo.toml
Legend-Master: I'm not very familiar with those type conversions in `windows-sys`, how do we convert `OsStr` to `HSTRING` here?

I did `HSTRING::from(found_path.as_os_str().as_encoded_bytes().as_ptr() as _)`, is this correct?

Legend-Master:
I doubt I did it right on the `windows-sys` version, felt like I was just cheating on the compiler 🤣

plugins/updater/Cargo.toml
amrbashir: `HSTRING` should be in `windows::core` module and not require any feature flags

plugins/updater/src/updater.rs
amrbashir: we were using `Start-Process` without `-Verb` which defaults to `open` iirc

plugins/updater/Cargo.toml
Legend-Master: I found it on https://microsoft.github.io/windows-rs/features/#/0.53.0, didn't know `windows::core::HSTRING` was there

image

amrbashir: that's probably just a re-export, anyways since this PR has changed to use `windows-sys`, it shouldn't matter.

plugins/updater/src/updater.rs
Legend-Master: Oops, you're right, there's no `-Verb runas` before, probably mixed it with something I saw else where

plugins/updater/src/updater.rs
```diff
@@ -534,118 +540,63 @@ impl Update {
...
-                        .spawn();
-                }
-
-                std::process::exit(0);
+                );
+                installer_args.push(OsStr::new("/promptrestart"));
+            } else {
+                continue;
+            }
+            let file = HSTRING::from(found_path.as_os_str().as_encoded_bytes().as_ptr() as _);
```
amrbashir: instead of as_encoded_bytes, let's use this utility function
```rs
pub fn encode_wide(string: impl AsRef<OsStr>) -> Vec<u16> {
    string.as_ref().encode_wide().chain(std::iter::once(0)).collect()
}
```

amrbashir: `HSTRING` shouldn't be needed here, we can do:
```rs
let file = encode_wide(found_path);
```
and then later use it in `ShellExecuteW` like this:
```rs
                ShellExecuteW(
                    0,
                    w!("open"),
                    file.as_ptr(),
                    parameters as _,
                    std::ptr::null(),
                    SW_SHOW,
                )
```
same should be done for parameters

amrbashir (changes requested):

amrbashir (approved):

PR #1059, opened 2024-03-09
fix(http): inconsistencies on urlpattern usage on scope

### LINKED ISSUE #1061: [http][v2] acl regression for url containing params
For url with params like: `https//www.domain.com/page?key1=value1&key2=value2`, even the acl has been set to allow all:
```
{
  "identifier": "http:default",
  "allow": [
    {"url": "https://*"}
  ]
}
```
It won't be allowed with error of `url not allowed on the configured scope`. It only affects the latest version of [http v2.0.0-beta.2](https://github.com/tauri-apps/plugins-workspace/releases/tag/http-v2.0.0-beta.2). Earlier version doesn't have this issue.

lucasfernog:
Changes the HTTP scope URL parsing to match the behaivor of the JavaScript implementation of URLPattern.
A pattern like `http://localhost` should automatically allow any subpath, query params and hash.
To define a custom query param match rule, you must escape the `?` character e.g. `http://localhost/path\\?q=*&x=*`

### DIFF
--- a/.changes/fix-http-scope-url-match.md
+++ b/.changes/fix-http-scope-url-match.md
@@ -1,0 +1,5 @@
+---
+"http": patch
+---
+
+Fixes scope not allowing subpaths, query parameters and hash when those values are empty.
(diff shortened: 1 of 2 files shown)


vangork:
Shall we add some test cases for urls containing query parameters and hash?

amrbashir (dismissed):

FabianLars:
> Shall we add some test cases for urls containing query parameters and hash?

Probably a good idea :)

PR #1065, opened 2024-03-12
fix(positioner): (v2) replace `Window` with `WebviewWindow` for Tauri 2.0

pewsheen:
We migrated `Window` to `WebviewWindow` in Tauri 2.0, need to update the API as well.

### DIFF
--- a/.changes/fix-positioner-webview-window.md
+++ b/.changes/fix-positioner-webview-window.md
@@ -1,0 +1,5 @@
+---
+"positioner": patch
+---
+
+Migrate `Window` to `WebviewWindow` for Tarui 2.0.
--- a/.changes/fix-positioner-webview-window.md
+++ b/.changes/fix-positioner-webview-window.md
@@ -2,4 +2,4 @@
 "positioner": patch
 ---
 
-Migrate `Window` to `WebviewWindow` for Tarui 2.0.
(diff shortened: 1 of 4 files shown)


FabianLars:
hmm, so something's wrong with our CI 🤔 It should have catched that in the PRs where tauri was updated...

FabianLars:
Also, maybe we should keep the WindowExt name and implement it for WebviewWindow and Window once it's stable? 🤔
cc @lucasfernog @amrbashir

amrbashir:
> Also, maybe we should keep the WindowExt name and implement it for WebviewWindow and Window once it's stable? 

I agree, this should be the way

pewsheen:
> Also, maybe we should keep the WindowExt name and implement it for WebviewWindow and Window once it's stable? 🤔 cc @lucasfernog @amrbashir

Sounds good

lucasfernog:
No need to update the command here, it still receives a valid `Window` instance (which is internally referenced by WebviewWindow).. the reason the Window and Webview structs are still usable without the unstable feature flag is that it gets complicated to manage that (specially for callback closures).

pewsheen:
Maybe just let users enable `unstable` feature and we can keep everything the same?

pewsheen:
It looks no need to change anything.

User can still get `Window` from the command or `WebviewWindow` by `webview_window.as_ref().window()`

PR #1070, opened 2024-03-12
fix(updater): run cleanup before exit on Windows

### LINKED ISSUE #1014: [updater] doesn't shutdown app gracefully on Windows
This skips app clean up code, for example kill sidecar / save windows states / save store

https://github.com/tauri-apps/plugins-workspace/blob/99bea2559c2c0648c2519c50a18cd124dacef57b/plugins/updater/src/updater.rs#L589

Legend-Master:
Fix #1014

### DIFF
--- a/.changes/fix-updater-cleanup-md
+++ b/.changes/fix-updater-cleanup-md
@@ -2,6 +2,4 @@
 "updater": patch
 ---
 
-Fix updater doesn't shutdown the app gracefully on update
-
-**Breaking change** `Update::download_and_install` now takes `AppHandle` as its first parameter
+Add a `on_before_exit` hook for cleanup before spawning the updater on Windows, defaults to `app.cleanup_before_exit`
--- a/.changes/fix-updater-cleanup-md
+++ b/.changes/fix-updater-cleanup-md
@@ -1,0 +1,7 @@
+---
+"updater": patch
+---
+
+Fix updater doesn't shutdown the app gracefully on update
+
+**Breaking change** `Update::download_and_install` now takes `AppHandle` as its first parameter
--- a/.changes/fix-updater-cleanup-md
+++ b/.changes/fix-updater-cleanup-md
@@ -2,4 +2,4 @@
 "updater": patch
 ---
 
-Add a `on_before_exit` hook for cleanup before spawning the updater on Windows, defaults to `app.cleanup_before_exit`
+Add a `on_before_exit` hook for cleanup before spawning the updater on Windows, defaults to `app.cleanup_before_exit` when used through `UpdaterExt`
--- a/plugins/updater/src/updater.rs
+++ b/plugins/updater/src/updater.rs
@@ -89,7 +89,7 @@ impl RemoteRelease {
     }
 }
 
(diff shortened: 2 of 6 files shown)


Legend-Master:
The downside about this is that we will have to assume `ShellExecute` will run successfully, since we need to clean up before the installer gets executed (installer will kill all running instances)

Legend-Master:
And, I'm not familiar with how updaters shutdown the app on other platforms, is this just a Windows problem or we need to do something for other platforms as well?

amrbashir:
> And, I'm not familiar with how updaters shutdown the app on other platforms, is this just a Windows problem or we need to do something for other platforms as well?

correct, other platforms update without restarting and wait for users to relaunch.

amrbashir (changes requested):
I don't quite like the additional argument, I think a better way to handle this would be to take a generic callback to be executed on before restart, here is some psuedo-code of what I mean:
```rs
let builder = UpdaterBuilder::new()
  .with_before_exit(|| {
    app.cleanup_before_exist();
  })
  .build();
```

and we can provide this callback automatically in the `UpdaterExt` trait implementation.

Legend-Master:
A bit off topic for this PR, do we want to expose `download` and `install` directly, so that the app can download the update in background and prompt the user once it's done (like the way vscode did it)?

amrbashir:
they are exposed https://docs.rs/tauri-plugin-updater/latest/tauri_plugin_updater/struct.Update.html#method.download

Legend-Master:
> I think a better way to handle this would be to take a generic callback to be executed on before restart

I'll take a look

> they are exposed https://docs.rs/tauri-plugin-updater/latest/tauri_plugin_updater/struct.Update.html#method.download

I mean the JavaScript side

amrbashir:
> I mean the JavaScript side

yeah sure, was hoping to do it at some point

.changes/fix-updater-cleanup-md
```diff
@@ -0,0 +1,5 @@
+---
+"updater": patch
+---
+
+Add a `on_before_exit` hook for cleanup before spawning the updater on Windows, defaults to `app.cleanup_before_exit`
```
amrbashir:
```suggestion
Add a `on_before_exit` hook for cleanup before spawning the updater on Windows, defaults to `app.cleanup_before_exit` when used through `UpdaterExt`
```

amrbashir (dismissed):
Thank you

amrbashir (dismissed):

plugins/updater/src/updater.rs
```diff
@@ -375,9 +388,10 @@ impl Updater {
     }
 }
 
-#[derive(Debug, Clone)]
+#[derive(Clone)]
 pub struct Update {
     config: Config,
+    on_before_exit: Option<OnBeforeExit>,
```
amrbashir, line 395:
```suggestion
    #[allow(unused)]
    on_before_exit: Option<OnBeforeExit>,
```

amrbashir (dismissed):

amrbashir (approved):

PR #1078, opened 2024-03-15
refactor(window-state)!: use json instead of bincode

thewh1teagle:
Add option to store window state in JSON readable format using `serde_json`

### DIFF
--- a/.changes/window-state-json.md
+++ b/.changes/window-state-json.md
@@ -1,0 +1,5 @@
+---
+"window-state": patch
+---
+
+**Breaking change**: Changed the format of the state file from bincode to json. Also changed the filename to from `.window-state` to `.window-state.json`.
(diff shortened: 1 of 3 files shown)


amrbashir (changes requested):
sorry for the very late reply, I think we should just use json only and remove bincode

thewh1teagle:
> sorry for the very late reply, I think we should just use json only and remove bincode

I think so too, just need to check that it's not a breaking change (what happens if the plugin can't read the old bincode file)
Also I think that it will be good to serialize it with indentation (pretty and readable) 

Update:
changed to use serde_json with and pretty serialized. also tested on fresh tauri app, and it works even when the previous window-state file isn't readable, it just rewrite it.

amrbashir (dismissed):

amrbashir (approved):
Thank you

Issue #516, opened 2023-07-26
[window-state] Set the name of the saved config file


luucasrb:
## Description

Instead of using the default `.window-state`, I want to set the name of the config file as a parameter to the `tauri_plugin_window_state::Builder::default().build()`, and be able to restore the same file when needed.

## Reference

The feature is similar to the [electron-store#name](https://github.com/sindresorhus/electron-store#name).

amrbashir:
I am a bit curious why you need to change the file name? since all state are managed by the plugin, so you can't modify it on your own

luucasrb:
> I am a bit curious why you need to change the file name? since all state are managed by the plugin, so you can't modify it on your own

I have two different frontend applications under the same Host and different ports that runs with the same Tauri code.

I want to be able to store different states for each application in a way that the state of the app1 don't intefere in the app2.

amrbashir:
that should already be covered by the [`identifier`](https://tauri.app/v1/api/config/#bundleconfig.identifier) field which should be unique

luucasrb:
> that should already be covered by the [`identifier`](https://tauri.app/v1/api/config/#bundleconfig.identifier) field which should be unique

Based on my understanding of the [identifier](https://tauri.app/v1/api/config/#bundleconfig.identifier) docs, it should be unique for the application domain name, right? In my use case, the domain is the same but the application port is different. 

- my.application.com:8000
- my.application.com:8001

Or should I simply set an random identifier for each application?

amrbashir:
the `identifier` should be unique for each app, you can reuse your domain but it has to at least be different in one character, so it should be something like this:
- com.domain.calculator
- com.domain.clock
- com.domain.dialer
and so on...

and I don't think `:` is allowed in the identifiers, so I think it gets stripped and thus the identifiers become the same and conflicts happen.

PR #1079, opened 2024-03-15
Feat/window state/custom state filename

thewh1teagle:
Depends on https://github.com/tauri-apps/plugins-workspace/pull/1078
Add option to use custom filename for window state file with the method `with_filename`

### DIFF
--- a/plugins/window-state/src/lib.rs
+++ b/plugins/window-state/src/lib.rs
@@ -23,13 +23,12 @@ use tauri::{
 use std::{
     collections::{HashMap, HashSet},
     fs::{create_dir_all, File},
-    io::Write,
     sync::{Arc, Mutex},
 };
 
 mod cmd;
 
-pub const STATE_FILENAME: &str = ".window-state";
+pub const STATE_FILENAME: &str = ".window-state.json";
 
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
@@ -38,7 +37,7 @@ pub enum Error {
     #[error(transparent)]
     Tauri(#[from] tauri::Error),
     #[error(transparent)]
-    Bincode(#[from] Box<bincode::ErrorKind>),
+    SerdeJson(#[from] serde_json::Error),
 }
 
 pub type Result<T> = std::result::Result<T, Error>;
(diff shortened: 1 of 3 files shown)


amrbashir:
Can you elaborate why you need this? Also see prior discussion here https://github.com/tauri-apps/plugins-workspace/issues/516

ImageDeeply:
Hello. I work with @thewh1teagle 

I don't mind having a hidden file with pre-defined name for binary data, but I think a json file should be visible and end with '.json'. If everyone agrees, then that suffices. But I assume that some people want to preserve the leading '.' and/or not append '.json' (and/or have other reasons for a custom filename). Hence this PR.

More generally, while I appreciate the benefits of 'opinionated' tools, I'm one of those app developers who want lots of choices so we can make our own tradeoffs.

Either way, thanks for considering.

plugins/window-state/src/lib.rs
```diff
@@ -24,12 +24,15 @@ use std::{
...
+    sync::{Arc,Mutex,OnceLock},
 };
 
 mod cmd;
 
-pub const STATE_FILENAME: &str = ".window-state";
+pub const DEFAULT_STATE_FILENAME: &str = ".window-state";
+static STATE_FILENAME: OnceLock<String> = OnceLock::new();
+
+static USE_JSON: OnceLock<bool> = OnceLock::new();
```
amrbashir: there is no need to use a static variable, you can just store it on the builder and clone/move it to the appropriate places, see https://github.com/tauri-apps/plugins-workspace/pull/546 for previous work.

amrbashir (changes requested):
Alright, let's do this

plugins/window-state/src/lib.rs
thewh1teagle: I restored the branch to latest v2 and [added these changes](https://github.com/thewh1teagle/plugins-workspace/commits/feat/window-state/custom-state-filename/), I think you need to open the PR, 
If it's not possible I'll create another PR

amrbashir: https://github.com/tauri-apps/plugins-workspace/pull/1138

Issue #926, opened 2024-02-02
[window-state] Only saving window position and not size (v2)


jaydevelopsstuff:
The window-state plugin (v2 off of crates.io) fails to save/restore window size (at least on MacOS, I haven't tested on any other operating systems).

[MRE Repo](https://github.com/jaydevelopsstuff/tauri-window-state-bug-mre) (Just run the app, resize, close, relaunch, and observe the size not being restored)

I've seen some similarish looking issues but they've been closed and don't seem to be addressing the same problem. If this has already been fixed and is just finding its way to release let me know.

amrbashir:
Couldn't reproduce on Windows, can't really investigate macos

bartekc7:
I confirm. This error also occurs on my macOS

legends-killer:
I find that this issue is caused by `tauri::window::Window::inner_size()` function. It always returns the width and height defined in `tauri.conf.json` but not the actual number of the current window size.
<img width="911" alt="image" src="https://github.com/tauri-apps/plugins-workspace/assets/47050572/44f31bfe-253f-496d-884b-71d53e044b54">
In the picture above, you'll find that the inner size is always 1024x768 which is the same as what in `tauri.conf.json`, but the outer size changes very differently.
So, it seems to be a bug of tauri v2.0-beta... But it can be a simple workaround by changing it into `self.outer_size()`.
https://github.com/tauri-apps/plugins-workspace/blob/v2/plugins/window-state/src/lib.rs#L269

legends-killer:
Hey guys, I found some PR related to this bug.
https://github.com/tauri-apps/tauri/pull/8280/files#diff-9fff33774bba86e0056dc083e3ec624f275b9d05427031c76dc1b313ade17a82R736
In this PR, Tauri 2.0 beta supported multiple webviews in a single window, but some functions about `get_webview` were marked as **unstable** causing the `get_size` function to return an unexpected value without **unstable** feature. So, you just need to enable the unstable feature in Tauri's dependency and this plugin could work normally.
```
// enable unstable features
tauri = { version = "2.0.0-beta", features = ["unstable"] }
```
Finally, I'm not sure if the `unstable` flag should be removed in Tauri, so I just created a PR to this repo to update docs in v2 README to workaround it. https://github.com/tauri-apps/plugins-workspace/pull/1086

roniemartinez:
[...]

PR #1086, opened 2024-03-18
chore(doc): add the requirement to tauri unstable feature to README.md

legends-killer:
To fix this issue: https://github.com/tauri-apps/plugins-workspace/issues/926
In Tauri v2 beta, macOS users should enable unstable features to make sure the restore window size function in the window-state plugin works correctly.


FabianLars:
I'd much rather see this fixed in tauri than adding an unstable feature flag that's called unstable for a reason tbh. Thank you for contributing though :)

Edit: Your comment in the issue is good enough as a workaround doc for the time being imo.

PR #1117, opened 2024-03-26
refactor(global-shortcut): enhance `un/register` to accept an array, remove `un/registerAll`

### LINKED ISSUE #1101: [global-shortcuts][v2] `unregisterAll()` fails as arguments not supplied
I'm trying to unregister all of the global shortcuts via the Javascript API. However, it consistently throws the same error.

```bash
"invalid args `shortcuts` for command `unregister_all`: command unregister_all missing required key shortcuts"
```

I can't pass any arguments as the `unregisterAll()` method expects 0 arguments. I assume this behavior is very much unexpected.

Is there a way to go around this?

amrbashir:
closes #1101


lucasfernog:
Should we do the same on the Rust side?
```rust
enum ShortcutArg {
  Single(T),
  List(Vec<T>),
}

impl From<T> for ShortcutArg { .. }
impl From<Vec<T>> for ShortcutArg { .. }

pub fn register(&self, shortcut: impl Into<ShortcutArg>) {...}
```

Also should it unregister all registered shortcuts if the list is empty (or we provide a separate API for that)?

amrbashir:
I don't think we should do the same on Rust, it is better to be kept as is.

> Also should it unregister all registered shortcuts if the list is empty (or we provide a separate API for that)?

we could bring back `unregisterAll` I guess

FabianLars (approved):

PR #1118, opened 2024-03-26
Fix LaunchAgent macOS autostart

### LINKED ISSUE #1115: [bug][autostart] LaunchAgent Autostart launcher does not work on macOS
The `LaunchAgent` macOS autostart launcher does not work. I've looked at LaunchAgent plists created by other non-Tauri apps that successfully autostart, and it looks like they all create a Plist item of the form:

```
  <key>ProgramArguments</key>
  <array>
      <string>/Applications/APP.app/Contents/MacOS/APP</string>
  </array>
``` 

Whereas, tauri-plugin-autostart creates a Plist item of the form:
```
  <key>ProgramArguments</key>
  <array>
      <string>/Applications/APP.app
  </array>
``` 
which does not work. 

I've tested manually modifying the plist to add back the `Contents/macOS/APP` bit, and it works perfectly. 

This looks like this is likely caused by PR #125 - which caused this diff, likely to make `autostart` work correctly for the `AppleScript` path.

However, we don't really want to use the `AppleScript` path for the app we're working on as it shows a scary-looking message to users on first boot. 

Ideally, this would correctly:
- a) use the `APP.app` path for `AppleScript`
- b) use the `APP.app/Contents/macOS/APP` path for `LaunchAgent`

venkr:
This is just #841 but for v2, since I noticed the last v1 -> v2 merge mentioned it might be the last time we run that sort of merge.

Fixes #1115

### DIFF
--- a/.changes/fix-autolaunch-macos.md
+++ b/.changes/fix-autolaunch-macos.md
@@ -1,0 +1,5 @@
+---
+"autostart": patch
+---
+
+Fix LaunchAgent-based autostart for macOS.
--- a/plugins/autostart/src/lib.rs
+++ b/plugins/autostart/src/lib.rs
@@ -129,11 +129,12 @@ pub fn init<R: Runtime>(
                 // exe path to not break it.
                 let exe_path = current_exe.canonicalize()?.display().to_string();
                 let parts: Vec<&str> = exe_path.split(".app/").collect();
-                let app_path = if parts.len() == 2 {
-                    format!("{}.app", parts.first().unwrap())
-                } else {
-                    exe_path
-                };
+                let app_path =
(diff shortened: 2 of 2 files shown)


FabianLars:
Thank you. Could you also add a changefile similar to https://github.com/tauri-apps/plugins-workspace/blob/v2/.changes/fix-shutdown-timing.md please?

venkr:
> Thank you. Could you also add a changefile similar to https://github.com/tauri-apps/plugins-workspace/blob/v2/.changes/fix-shutdown-timing.md please?

Added, please take a look! :)

FabianLars (approved):
thanks

PR #1120, opened 2024-03-27
allow updater check command on mobile

### LINKED ISSUE #1119: [updater][v2] enable check command on mobile
Current it will fail with error of "Unsupported OS, expected one of `linux`, `darwin` or `windows`.."

zhangsan946:
to resolve #1119

### DIFF
--- a/plugins/updater/src/updater.rs
+++ b/plugins/updater/src/updater.rs
@@ -747,6 +747,10 @@ pub(crate) fn get_updater_target() -> Option<&'static str> {
         Some("darwin")
     } else if cfg!(target_os = "windows") {
         Some("windows")
+    } else if cfg!(target_os = "ios") {
+        Some("ios")
+    } else if cfg!(target_os = "android") {
+        Some("android")
     } else {
         None
     }


lucasfernog:
the updater isn't implemented on mobile yet, so this change doesn't help, we'd need to implement the whole flow :(

zhangsan946:
@lucasfernog my application just uses the updater  `check` command to check where there is any new version in the rest api & .json file. If there is any, it will open the link for the specific platform & arch in the external web browser and download. So my change should be enough to execute the `check` command. And later on I will open another PR to enable `open` command on mobile for `shell` plugin.

lucasfernog:
you shouldn't run the check() command on mobile, it'll always give you an error. Instead, you can feature flag its usage (to not include the code at compile time) or make a runtime check.

pseudo code:

- feature flag
```rust
#[cfg(desktop)]
fn run_updater(app) {
  app.updater().check()...
}

tauri::Builder::default()
  .setup(|app| {
    #[cfg(desktop)]
    run_updater(app);
    Ok(())
  })
```

- runtime check
```rust
if cfg!(desktop) {
  app.updater().check()...
}

```

PR #1134, opened 2024-04-02
chore: add `tslib` as dev dependency in workspace

amrbashir:


lucasfernog (approved):
I don't really remember why we need tslib (was it private field usage?) but ok

amrbashir:
It is a peer dependency of rollup-typescript plugin,

PR #1159, opened 2024-04-08
feat(upload): do not expect JSON to be return on upload

lyager:
Rather look at the returncode, and return the result as string upon success.


FabianLars:
Sorry for the insane delay here. If you're still around, would you mind merging the upstream branch to resolve the merge conflicts?

lyager:
Looked through current v2, and the issue was fixed and JSON is no longer expected. Closing this one - thanks for getting back though

PR #1166, opened 2024-04-09
fix(http): Enable tls support by default

FabianLars:
supporting https by default is imho a must (especially seeing the amount of reports about this). The other feature flags are reqwest's defaults and they sound reasonable to me but i'm open to discuss those :)


amrbashir (commented):
I think we just need to add `default = ["default-tls"]` and it should work, no?

FabianLars:
not a fan of using native tls by default tbh. Annoying to deal with when cross compiling (iirc especially windows -> android?) and not the best idea in appimages either atm.

plugins/http/Cargo.toml
```diff
@@ -35,14 +35,14 @@ url = { workspace = true }
...
 
 [features]
+default = ["rustls-tls", "http2", "charset", "macos-system-configuration"]
 multipart = ["reqwest/multipart"]
 json = ["reqwest/json"]
 stream = ["reqwest/stream"]
 native-tls = ["reqwest/native-tls"]
 native-tls-vendored = ["reqwest/native-tls-vendored"]
-rustls-tls = ["reqwest/rustls-tls"]
-default-tls = ["reqwest/default-tls"]
```
amrbashir, line 44: we need to keep these two features, so users can disable default features and customize as they want

amrbashir (changes requested):

plugins/http/Cargo.toml
FabianLars: Whoops, rustls was removed by accident, will revert that tomorrow.
default-tls was removed because it's an alias for native-tls (or the other way around) and I think the name is pretty misleading (probably not the right word for it idk) or not self explanatory at least.

FabianLars: Oh wait, just noticed that I just re-ordered it. The rustls flag is still there just further down below lol

FabianLars: Oh and another reason why I removed default-tls is because I used rustls as the default and thought that may also be confusing when default-tls is not default

amrbashir: cool

amrbashir (approved):

PR #1168, opened 2024-04-09
feat: Remove UNC prefix in paths returned to the frontend

### LINKED ISSUE #9407: [feat] Remove UNC prefix from paths returned in public apis
### Describe the problem

Currently our APIs (most of them moved to plugins i guess) still return UNC paths on Windows. For those unaware, those paths start with `\\?\`.

Most ecosystems seem to have problems with that, this includes the js frontend but also a multitude of sidecars. We've also already received reports about things built into Windows not being able to handle those.

use dunce everywhere to safely remove the prefix. I doubt that this will cause any issues but even if, it'd be still less than the unc paths cause lol

### Alternatives considered

do nothing and keep blaming the other languages' ecosystems for not being able to handle unc paths.

### Additional context

_No response_

FabianLars:
fixes https://github.com/tauri-apps/tauri/issues/9407

please double check this, i am super tired lol


plugins/dialog/src/commands.rs
```diff
@@ -121,12 +121,13 @@ pub(crate) async fn open<R: Runtime>(
         #[cfg(desktop)]
         {
             if options.multiple {
-                let folders = dialog_builder.blocking_pick_folders();
-                if let Some(folders) = &folders {
+                let mut folders = dialog_builder.blocking_pick_folders();
```
amrbashir: I think I'd rather avoid mutating here and instead use `.iter().map` later at line 133.

plugins/dialog/src/commands.rs
```diff
@@ -137,33 +138,35 @@ pub(crate) async fn open<R: Runtime>(
...
-                OpenResponse::Folder(folder)
+                OpenResponse::Folder(folder.map(|p| dunce::simplified(&p).to_path_buf()))
             }
         }
         #[cfg(mobile)]
         return Err(crate::Error::FolderPickerNotImplemented);
     } else if options.multiple {
-        let files = dialog_builder.blocking_pick_files();
-        if let Some(files) = &files {
+        let mut files = dialog_builder.blocking_pick_files();
```
amrbashir: ditto

plugins/dialog/src/commands.rs
```diff
@@ -137,33 +138,35 @@ pub(crate) async fn open<R: Runtime>(
...
                     .state::<tauri::scope::Scopes>()
                     .allow_file(&file.path)?;
+                file.path = dunce::simplified(&file.path).to_path_buf();
             }
         }
         OpenResponse::Files(files)
     } else {
-        let file = dialog_builder.blocking_pick_file();
-        if let Some(file) = &file {
+        let mut file = dialog_builder.blocking_pick_file();
```
amrbashir: ditto

plugins/fs/src/commands.rs
```diff
@@ -336,7 +336,7 @@ pub fn read_file<R: Runtime>(
     command_scope: CommandScope<Entry>,
     path: SafePathBuf,
     options: Option<BaseOptions>,
-) -> CommandResult<Vec<u8>> {
+) -> CommandResult<tauri::ipc::Response> {
```
amrbashir: I think this is out of this PR scope, no? and there is a one or more functions that will need this change so better make a new PR

plugins/dialog/Cargo.toml
```diff
@@ -24,6 +24,7 @@ tauri = { workspace = true }
 log = { workspace = true }
 thiserror = { workspace = true }
 tauri-plugin-fs = { path = "../fs", version = "2.0.0-beta.4" }
+dunce = "1"
```
amrbashir: let's move this to the workspace Cargo.toml

plugins/store/Cargo.toml
```diff
@@ -22,3 +22,4 @@ serde_json = { workspace = true }
 tauri = { workspace = true }
 log = { workspace = true }
 thiserror = { workspace = true }
+dunce = "1"
```
amrbashir: ditto

amrbashir (changes requested):

plugins/fs/src/commands.rs
FabianLars: probably committed by accident actually lol

plugins/dialog/src/commands.rs
FabianLars: Like this? I don't think this works that well for the other 2 below that return FileResponse structs but take a look.

https://github.com/tauri-apps/plugins-workspace/pull/1168/commits/68e7d1ebf5a1f02c67cd4dcd89c65223341402d2

amrbashir (dismissed):

amrbashir (approved):

PR #1174, opened 2024-04-11
feat(updater): support non zipped updater

### LINKED ISSUE #1083: [updater] Move updater out of zip
Updaters/installers are already compressed files, there's no reason to place them in a zip file

Even if the user choose no compress on the installer, web servers (update servers) are generally able to send compressed stream anyway

Having to use zip makes it an extra step on the bundler and updater, and this makes the process slower and binary size bigger #1085

This will be a pretty big breaking change though (can be somewhat mitigated by introducing a new field in the update manifest to point to the plain installer)

### LINKED ISSUE #5760: [feat] Windows software update using unzipped msi files
### Describe the problem

tauri-updater generates .msi, .msi.zip, and .msi.zip.sig. The problem is that the ZIP file is **larger** than the bare .msi file. The installer file is already compressed, and extra zipping is counter-productive. 

I'd prefer to just use the bare .msi file, because that's compressed enough, and it would allow me to upload and store only one file for both auto-update as well as initial download for users.

Can you make the Windows updater stop zipping msi installers, and instead sign the .msi file? (so there's app.msi.sig, not app.msi.zip.sig).

### Alternatives considered

_No response_

### Additional context

_No response_

Legend-Master:
Closes #1083
closes tauri-apps/tauri#5760


Legend-Master:
Note: I didn't have to time to test this yet

Legend-Master:
Tested on Windows

plugins/updater/src/updater.rs
```diff
@@ -899,3 +966,50 @@ fn encode_wide(string: impl AsRef<OsStr>) -> Vec<u16> {
...
+        && buf[5] == 0xB1
+        && buf[6] == 0x1A
+        && buf[7] == 0xE1
+}
+
+// Taken from infer crate https://github.com/bojand/infer (MIT License)
+#[cfg(target_os = "linux")]
+fn is_gz(buf: &[u8]) -> bool {
+    buf.len() > 2 && buf[0] == 0x1F && buf[1] == 0x8B && buf[2] == 0x8
+}
```
amrbashir: Let's depend on `infer` crate instead of copying the implementation

plugins/updater/src/updater.rs
```diff
@@ -540,12 +540,71 @@ impl Update {
...
+                #[allow(unused)]
+                temp_path: Option<tempfile::TempPath>,
+            },
+        }
 
-        let paths = fs::read_dir(&tmp_dir)?;
+        let updater = 'updater: {
+            #[cfg(feature = "zip")]
+            {
+                if is_zip(&bytes) {
```
amrbashir: This can be collapsed
```suggestion
            #[cfg(feature = "zip")]
            if is_zip(&bytes) {
```

plugins/updater/src/updater.rs
```diff
@@ -540,12 +540,71 @@ impl Update {
...
+            Msi {
+                path: PathBuf,
+                // For extending temp file's life time
+                #[allow(unused)]
+                temp_path: Option<tempfile::TempPath>,
+            },
+        }
 
-        let paths = fs::read_dir(&tmp_dir)?;
+        let updater = 'updater: {
```
amrbashir: instead of using a labeled block with a break statement, we could split these 3 conditions into 2 functions

```rs
enum UpdaterType {
  Msi,
  Nsis
}

let (updater_type, path, _temp) = if is_zip() {
  Self::extract_zip(&bytes)?
} else {
  Self::extract_bin(&bytes)?
};
``` 

where `extract_bin` and `extract_zip` will check if msi or nsis and write with appropriate file extension then returns whether this update package is MSI or not

```rs
fn extract_zip(bytes: &[u8]) -> Result<(UpdaterType, PathBuf, Option<TempPath>)>;
```

plugins/updater/src/updater.rs
```diff
@@ -641,27 +692,43 @@ impl Update {
...
-                    for mut entry in archive.entries()?.flatten() {
-                        if let Ok(path) = entry.path() {
-                            if path.extension() == Some(OsStr::new("AppImage")) {
-                                // if something went wrong during the extraction, we should restore previous app
-                                if let Err(err) = entry.unpack(&self.extract_path) {
-                                    std::fs::rename(tmp_app_image, &self.extract_path)?;
-                                    return Err(err.into());
+                    if is_gz(&bytes) {
+                        #[cfg(feature = "zip")]
+                        {
```
amrbashir: can be collapsed
```rs

#[cfg(feature = "zip)]
if is_gz() {

}

// else case here
```

amrbashir (changes requested):

plugins/updater/src/updater.rs
Legend-Master: Do we have other things depending on infer? I feel like if we only need 1-2 small functions, it's better not to ship the entire infer crate

Legend-Master: Never mind, tauri utils uses it for mime types

plugins/updater/src/updater.rs
```diff
@@ -558,41 +573,90 @@ impl Update {
...
+        if infer::archive::is_msi(&bytes) {
+            let mut temp_file = tempfile::Builder::new().suffix(".msi").tempfile()?;
+            temp_file.write_all(&bytes)?;
+            let temp_path = temp_file.into_temp_path();
+            return Ok(WindowsUpdaterType::Msi {
+                path: temp_path.to_path_buf(),
+                temp_path: Some(temp_path),
+            });
+        }
+        return Err(crate::Error::InvalidUpdaterFormat);
```
amrbashir: use if else instead of multiple return statements
```suggestion
        if infer::app::is_exe(&bytes) {
            let mut temp_file = tempfile::Builder::new().suffix(".exe").tempfile()?;
            temp_file.write_all(&bytes)?;
            let temp_path = temp_file.into_temp_path();
            Ok(WindowsUpdaterType::Nsis {
                path: temp_path.to_path_buf(),
                temp_path: Some(temp_path),
            })
        } else if infer::archive::is_msi(&bytes) {
            let mut temp_file = tempfile::Builder::new().suffix(".msi").tempfile()?;
            temp_file.write_all(&bytes)?;
            let temp_path = temp_file.into_temp_path();
            Ok(WindowsUpdaterType::Msi {
                path: temp_path.to_path_buf(),
                temp_path: Some(temp_path),
            })
        } else {
            Err(crate::Error::InvalidUpdaterFormat)
        }
```

plugins/updater/src/updater.rs
```diff
@@ -641,27 +703,38 @@ impl Update {
...
                         }
+                        // if we have not returned early we should restore the backup
+                        std::fs::rename(tmp_app_image, &self.extract_path)?;
+                        return Err(Error::BinaryNotFoundInArchive);
                     }
-                    // if we have not returned early we should restore the backup
-                    std::fs::rename(tmp_app_image, &self.extract_path)?;
-
-                    return Err(Error::BinaryNotFoundInArchive);
+                    return match std::fs::write(&self.extract_path, bytes) {
```
amrbashir, line 721: unneeded return
```suggestion
                    match std::fs::write(&self.extract_path, bytes) {
```

plugins/updater/src/updater.rs
```diff
@@ -641,27 +703,38 @@ impl Update {
...
+                        Err(err) => {
+                            // if something went wrong during the extraction, we should restore previous app
+                            std::fs::rename(tmp_app_image, &self.extract_path)?;
+                            return Err(err.into());
```
```suggestion
                            Err(err.into());
```

amrbashir (changes requested):

amrbashir (dismissed):

amrbashir (approved):

PR #1189, opened 2024-04-15
chore: merge v1 into v2 one last time

FabianLars:
~~wip, still gotta add changefiles, just wanna get ci running already.~~

_DON'T SQUASH MERGE THIS!_

p.s. after this pr we will change the default branches :)


plugins/single-instance/src/platform_impl/linux.rs
```diff
@@ -28,14 +31,27 @@ impl<R: Runtime> SingleInstanceDBus<R> {
     }
 }
 
+#[cfg(feature = "semver")]
+fn dbus_id(config: &Config, version: semver::Version) -> String {
+    let mut id = config.identifier.replace(['.', '-'], "_");
+    id.push('_');
+    id.push_str(semver_compat_string(version).as_str());
+    id
+}
```
amrbashir, line 40: we need to do the same for macOS backend as well, no?

plugins/store/examples/AppSettingsManager/dist/.gitkeep
amrbashir, line 1: why do we need to keep this `dist` directory?

plugins/store/examples/AppSettingsManager/src-tauri/.gitignore
```diff
@@ -0,0 +1,4 @@
+# Generated by Cargo
+# will have compiled files and executables
+/target/
+
```
amrbashir:
```suggestion
gen/schemas
```

amrbashir (changes requested):

plugins/single-instance/src/platform_impl/linux.rs
FabianLars: now i regret adding support for macos...

plugins/store/examples/AppSettingsManager/dist/.gitkeep
FabianLars: so that it does not crash on the generate_context line. We do the same in other examples. I'm open to better ideas (that don't involve rewriting the example in vanillajs 😂 ) because i don't like it either.

amrbashir: I thought we don't crash on generte_context anymore, anyways doesn't matter

amrbashir (approved):

plugins/store/examples/AppSettingsManager/dist/.gitkeep
FabianLars: we removed the default feature flag from the templates so that vscode won't show the panic but iirc the CI enables all features. I feel like i also just saw the same error locally before i even pushed it but can't remember anymore 🤔

PR #1191, opened 2024-04-15
fix: use webview's resources table

### LINKED ISSUE #1194: [clipboard manager] v2 tauri 2.0.0-beta.15 changed `into_img` method signature
tauri `2.0.0-beta.15` changed `into_image` to accept `ResourceTable`, it breaks clipboard manager plugin here:

https://github.com/tauri-apps/plugins-workspace/blob/86723a5372af0f0bb4083ef1bf407b0199306728/plugins/clipboard-manager/src/desktop.rs#L42-L46

Maybe we can fix this by:

```
                    let resource_table = self.app.resources_table();
                    let image = image.into_img(&resource_table)?;
```

amrbashir:


pronebird:
```
error[E0308]: mismatched types
   --> /Users/pronebird/.cargo/git/checkouts/plugins-workspace-fd0259fcb18f51ad/3af49d7/plugins/clipboard-manager/src/desktop.rs:46:48
    |
46  |                     let image = image.into_img(&self.app)?;
    |                                       -------- ^^^^^^^^^ expected `&ResourceTable`, found `&AppHandle<R>`
    |                                       |
    |                                       arguments to this method are incorrect
    |
    = note: expected reference `&ResourceTable`
               found reference `&AppHandle<R>`
```

FabianLars (dismissed):

FabianLars (approved):
does the resource table from App contain the resources from Webview? Asking because what if the inner JsImage was created with Webview but we're using App here.

amrbashir:
Let's merge this for now and I will open another PR to refactor and remove the need for resources_table when using the Rust APIs

PR #1192, opened 2024-04-15
feat(http): enable cookies and set origin header

### LINKED ISSUE #1057: [http] Cookies cannot be passed to the server in custom requests
It should be that the cookie was lost after being passed from the fetch to the trust end, and cannot be passed to the remote server
in tauri-plugin-http = "2.0.0-beta.2"

the custom header was only responsible for transmitting to Rust, and Rust, as a true request, was not responsible for transmitting it as is, resulting in the server not receiving it at all

### LINKED ISSUE #1167: [http] Use browser cookies
I'd like `reqwest` on the rust side to be able to send a request to the server as the webview's browser: meaning automatically including the webview's cookies.

In addition, it would be interesting if, when a request is made to `/something`, the default domain was the one of the actual webpage (like the javascript fetch).

amrbashir:
closes #1167


FabianLars:
Do we need to make the origin configurable (including omitting it)? Or should we wait and see if users request that?

amrbashir:
I'd probably wait and see if there is any use-case to omitting/modifying it

plugins/http/src/lib.rs
```diff
@@ -9,7 +9,7 @@
 pub use reqwest;
 use tauri::{
     plugin::{Builder, TauriPlugin},
-    AppHandle, Manager, Runtime,
+    Runtime
```
amrbashir:
```suggestion
    Runtime,
```

FabianLars:
afaik just the feature flag alone doesn't do much, we have to enable cookie support on the builder too. I assume that we'll need to store our own cookie store/jar in a State to share cookies between fetch calls because we don't reuse clients anymore.

amrbashir:
> I assume that we'll need to store our own cookie store/jar in a State to share cookies between fetch calls because we don't reuse clients anymore.

I think so, let me add that real quick

amrbashir:
should be good now

FabianLars (approved):
(i didn't test it)

ToxicalNoob3062:
@amrbashir can you pls fix the error :
```
tauri-runtime-wry v2.0.0-beta.16
error[E0599]: no method named `as_str` found for enum `std::result::Result` in the current scope
   --> /home/rahat/.cargo/git/checkouts/plugins-workspace-fd0259fcb18f51ad/bd07784/plugins/http/src/commands.rs:249:72
    |
249 |                 request = request.header(header::ORIGIN, webview.url().as_str());
    |                                                                        ^^^^^^ method not found in `Result<Url, Error>`
    |
note: the method `as_str` exists on the type `reqwest::Url`
   --> /home/rahat/.cargo/registry/src/index.crates.io-6f17d22bba15001f/url-2.5.0/src/lib.rs:581:5
    |
581 |     pub fn as_str(&self) -> &str {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
help: use the `?` operator to extract the `reqwest::Url` value, propagating a `Result::Err` value to the caller
    |
249 |                 request = request.header(header::ORIGIN, webview.url()?.as_str());
    |                                                                       +

For more information about this error, try `rustc --explain E0599`.
error: could not compile `tauri-plugin-http` (lib) due to previous error
warning: build failed, waiting for other jobs to finish...
error: script "tauri" exited with code 101`
```
It's a very small error ! but for this I can't compile the code of my app!

amrbashir:
@ToxicalNoob3062 this is already fixed in https://github.com/tauri-apps/plugins-workspace/pull/1325 and will be available in the next release, for now, ping `tauri` to a previous version until the next release of this plugin.

FabianLars:
> Do we need to make the origin configurable (including omitting it)? Or should we wait and see if users request that?

Welp, 8 hours after release: https://github.com/tauri-apps/plugins-workspace/issues/1389

amrbashir:
oh well, gotta open another PR

PR #1231, opened 2024-04-22
feat(shell) raw-encoded pipe reader directly outputs buffer (no newline scan)

GCRev:
One of the things that I've had to change in both the previous v1 shell and again in the v2 shell plugin is the way that the pipe reader handles output from a process. I'm working with a process that outputs binary, so breaking at the newline character disrupts the binary output. I think others have also encountered this.

I'm new to rust, but this seems like a somewhat reasonable approach that would preserve the original line-based behavior but present "raw" streams without changing anything between the process and the front end. This is, technically, a breaking change, but I propose that this is worth it going forward for the v2 release.

### DIFF
--- a/plugins/shell/src/process/mod.rs
+++ b/plugins/shell/src/process/mod.rs
@@ -4,7 +4,7 @@
 
 use std::{
     ffi::OsStr,
-    io::{BufReader, Write},
+    io::{BufRead, BufReader, Write},
     path::{Path, PathBuf},
     process::{Command as StdCommand, Stdio},
     sync::{Arc, RwLock},
@@ -53,7 +53,10 @@ pub enum CommandEvent {
 
 /// The type to spawn commands.
 #[derive(Debug)]
-pub struct Command(StdCommand);
+pub struct Command {
+    cmd: StdCommand,
+    raw_out: bool
+}
 
 /// Spawned child process.
 #[derive(Debug)]
@@ -122,7 +125,7 @@ fn relative_command_path(command: &Path) -> crate::Result<PathBuf> {
 
 impl From<Command> for StdCommand {
     fn from(cmd: Command) -> StdCommand {
-        cmd.0
+        cmd.cmd
     }
 }
 
@@ -136,7 +139,7 @@ impl Command {
         #[cfg(windows)]
         command.creation_flags(CREATE_NO_WINDOW);
 
-        Self(command)
+        Self{ cmd: command, raw_out: false }
     }
 
     pub(crate) fn new_sidecar<S: AsRef<Path>>(program: S) -> crate::Result<Self> {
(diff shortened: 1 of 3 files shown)


plugins/shell/src/process/mod.rs
```diff
@@ -364,11 +375,32 @@ fn spawn_pipe_reader<F: Fn(Vec<u8>) -> CommandEvent + Send + Copy + 'static>(
...
+                        reader.consume(length);
+                    }
+                    Err(e) => {
+                        let tx_ = tx.clone();
+                        let _ = block_on_task( async move { tx_.send(CommandEvent::Error(e.to_string())).await });
+                    }
+                }
+            }
+            return;
+        }
```
amrbashir: lets move the current logic and the new logic into two separate function and then call them conditionally:

```rs
if raw {
  read_raw_bytes(&mut reader)
} else {
  read_line(&mut reader)
}
```

plugins/shell/src/process/mod.rs
```diff
@@ -187,14 +190,19 @@ impl Command {
...
 
     /// Sets the working directory for the child process.
     #[must_use]
     pub fn current_dir<P: AsRef<Path>>(mut self, current_dir: P) -> Self {
-        self.0.current_dir(current_dir);
+        self.cmd.current_dir(current_dir);
+        self
+    }
+    
+    pub fn set_raw_out(mut self, raw_out: bool) -> Self {
```
amrbashir, line 210: Missing documentation and you will also need to update the documentation of `CommandEvent` enum

plugins/shell/src/process/mod.rs
```diff
@@ -364,11 +375,32 @@ fn spawn_pipe_reader<F: Fn(Vec<u8>) -> CommandEvent + Send + Copy + 'static>(
...
         let _lock = guard.read().unwrap();
         let mut reader = BufReader::new(pipe_reader);
 
+        if raw_out {
+            loop {
+                let result = reader.fill_buf();
+                match result {
+                    Ok(buf) => {
+                        let length = buf.len();
+                        if length == 0 { break; }
```
amrbashir:
```suggestion
                        if length == 0 { 
                            break; 
                        }
```

amrbashir (changes requested):
Thanks for your contributions and sorry for the late response. 

Could you also add a change file in `.changes` directory?

plugins/shell/src/process/mod.rs
GCRev: Done

plugins/shell/src/process/mod.rs
GCRev: Hopefully implemented the way you expected!

plugins/shell/src/process/mod.rs
GCRev: Done

GCRev:
Let me know if I did the `.changes` file consistently with the others. I'm not sure whether this was a fix or an enhancement.

amrbashir (dismissed):

amrbashir (approved):

amrbashir:
Thank you

PR #1251, opened 2024-04-24
feat(fs): use IPC streaming for read and write file

lucasfernog:
The HTTP request API already uses the streaming IPC, let's do the same for the fs functions.

Needs https://github.com/tauri-apps/tauri/pull/9561 since we pass the path and the options as headers on the writeFile command.


DexYang:
Looking forward to merge this...

ezzaouia:
Hi, I'm currently experiencing issues with read / write file (binary), after updating dependencies in v2.

In js api, `readfile` returns an empty buffer.

The writeFile throws `"invalid args `path` for command `write_file`: command write_file expected a value for key path but the IPC call used a bytes payload"`.
The path is very well a a string.

I suspect the issue might be due to the changes made for this issue #1251.

Any ideas?
Many thanks.

FabianLars:
@ezzaouia This sounds like you only updated the rust crate but not the js package of the plugin, can you check that _both_ are up to date?

ezzaouia:
Thank you very much Fabian, indeed, js package was not up to date with rust.
My issue is resolved upon the update of both sides.

Happy WE.

PR #1259, opened 2024-04-27
fix(window-state): restore window state for stubbed zerord state

Legend-Master:
We currently ignore default all zero state in `restore_state`, but this will prevent `skip_initial_state` + calling `restore_state` manually to show the window if no previous cache was found, we should treat default all zero state as no cache found

### DIFF
--- a/.changes/restore-default-window-state.md
+++ b/.changes/restore-default-window-state.md
@@ -1,0 +1,5 @@
+---
+"window-state": patch
+---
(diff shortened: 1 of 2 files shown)


amrbashir:
Not sure I understand the problem, could you explain a bit more?

amrbashir (approved):
nvm, I get what you mean

PR #1295, opened 2024-05-07
feat(dialog): set `len` and `modified_at` fields in `FileResponse` on desktop

luis-cicada:
### Description:

This pull request introduces a feature enhancement to the FileResponse struct in the Tauri project. The enhancement allows the FileResponse struct to automatically compute and include the size of the file in bytes when creating a new instance.

### Changes:
- Added functionality to retrieve the size of a file in bytes.
- Implemented logic to calculate the size of the file using the standard library's std::fs::metadata method.
- Updated the FileResponse struct's new function to include the size of the file as a field.

### Benefits:
- Users of the FileResponse struct can now easily access the size of a file without needing to manually compute it.
- Simplifies code for developers by providing a convenient way to include file size information in FileResponse instances.

### Considerations:
- Error handling for file metadata retrieval is implemented using unwrap(). Further improvements can be made to handle errors more gracefully.
- File size is reported in bytes. No unit conversion (e.g., kilobytes or megabytes) is performed at this stage.

### DIFF
--- a/.changes/dialog-metadata.md
+++ b/.changes/dialog-metadata.md
@@ -1,0 +1,6 @@
+---
+"dialog": "patch"
+"dialog-js": "patch"
+---
+
+Fill file `len` and `modified_at` fields of `FileResponse` when using the open dialog.
--- a/plugins/dialog/src/lib.rs
+++ b/plugins/dialog/src/lib.rs
@@ -18,6 +18,7 @@ use tauri::{
 };
 
 use std::{
+    fs,
     path::{Path, PathBuf},
     sync::mpsc::sync_channel,
 };
@@ -230,6 +231,8 @@ pub struct FileResponse {
 impl FileResponse {
     #[cfg(desktop)]
     fn new(path: PathBuf) -> Self {
+        let metadata = fs::metadata(&path).unwrap();
+
         Self {
             base64_data: None,
             duration: None,
@@ -239,7 +242,7 @@ impl FileResponse {
             modified_at: None,
             name: path.file_name().map(|f| f.to_string_lossy().into_owned()),
             path,
-            size: 0,
+            size: metadata.len(),
         }
     }
 }
--- a/plugins/dialog/src/lib.rs
+++ b/plugins/dialog/src/lib.rs
@@ -214,7 +214,7 @@ impl<R: Runtime> MessageDialogBuilder<R> {
     }
 }
 
-#[derive(Debug, Deserialize, Serialize)]
+#[derive(Debug, Deserialize, Serialize, Default)]
 #[serde(rename_all = "camelCase")]
 pub struct FileResponse {
     pub base64_data: Option<String>,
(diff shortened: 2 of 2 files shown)


amrbashir:
@FabianLars why are we returning metadata in the dialog plugin? shouldn't that be left to be handled by the fs plugin?

luis-cicada:
@amrbashir we want to get the file size right from the input rather than then using another plugin to read it. Our business case is to prevent files from being processed if they are larger than 100mb

FabianLars:
> @FabianLars why are we returning metadata in the dialog plugin? shouldn't that be left to be handled by the fs plugin?

no idea, wasn't involved in that at all. i assume it makes sense to do so on mobile (not so much on desktop then which is why it's all None etc)

amrbashir:
> @amrbashir we want to get the file size right from the input rather than then using another plugin to read it. Our business case is to prevent files from being processed if they are larger than 100mb

It is not the responsibility of the dialog plugin to access the metadata of the file, it should be done by the `fs` plugin, this ensures that plugins are composable and work together effortlessly. This also removes the duplicated code that we will have to maintain.

amrbashir:
> > @FabianLars why are we returning metadata in the dialog plugin? shouldn't that be left to be handled by the fs plugin?
> 
> no idea, wasn't involved in that at all. i assume it makes sense to do so on mobile (not so much on desktop then which is why it's all None etc)

maybe @lucasfernog has better idea, it should be removed imo

luis-cicada:
> > @amrbashir we want to get the file size right from the input rather than then using another plugin to read it. Our business case is to prevent files from being processed if they are larger than 100mb
> 
> It is not the responsibility of the dialog plugin to access the metadata of the file, it should be done by the `fs` plugin, this ensures that plugins are composable and work together effortlessly. This also removes the duplicated code that we will have to maintain.

Are you saying then that we should invoke dialog and fs just to get the size of the file? I understand the separation of concerns between the 2 modules, but in my opinion metadata access from the dialog is usefull to have on the fly.

amrbashir:
> Are you saying then that we should invoke dialog and fs just to get the size of the file?

yes, this is what I meant.

> I understand the separation of concerns between the 2 modules, but in my opinion metadata access from the dialog is usefull to have on the fly.

It will still call the same APIs under the hood, and even though it is an extra call from JS side, which can be considered more overhead and a bit slower, it is still very very miniscule.

That said, I am gonna close the PR and open another to remove the metadata from the dialog plugin, I recommend you either use `fs` plugin to read the metadata, or create your own custom command to open the dialog and read the metadata at the same time.

amrbashir:
Looks like `FileResponse` metadata is used in the mobile portion for the plugin, and since `fs` plugin doesn't support mobile atm, I will accept this PR but in the future this will probably be removed as soon as `fs` plugin supports mobile.

amrbashir (approved):
Thanks for contribution

PR #1299, opened 2024-05-08
fix(core/shell): speedup `Command.execute` & fix extra new lines

amrbashir:
The speed gains comes from running the Command in Rust fully and returning the result in one go instead of using events.

The extra new lines was a regression from https://github.com/tauri-apps/tauri/pull/6519
ref: https://github.com/tauri-apps/tauri/issues/7684#issuecomment-2100897383


lucasfernog (approved):

1zilc:
This PR caused my page to freeze abnormally, all `.execute()` look like they are executing synchronously 😕

amrbashir:
It is not synchronous but once the program finishes execution it will return its whole output at once, this should drastically increase speed for small to medium outputs, however large output data may notice freezes as the output comes in, in which I'd recommend using `spawn` and process each line as it comes but that will have its penalty as well.

1zilc:
Thanks for the reply, I'll provide a minimal reproduction later

PR #1305, opened 2024-05-10
feat(sql): preserve column order from query with indexmap

thewh1teagle:
When using the `select` command in the `sql` plugin, the field order in each row isn't maintained due to the use of a hashmap. This update ensures that the field order remains consistent.

Example:

```sql
SELECT name, age FROM users;
```

Previously, the returned rows could have inconsistent order, such as 
```
{"name": ..., "age": ...}
```
 or 
```
{"age": ..., "name": ...}
```
With this modification, the order will consistently follow the query, i.e., `name` followed by `age`.

### DIFF
--- a/.changes/sql-column-order.md
+++ b/.changes/sql-column-order.md
@@ -1,0 +1,5 @@
+---
+"sql": patch
+---
+
+Preserve column order from SELECT query with indexmap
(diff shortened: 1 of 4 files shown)


FabianLars:
Thanks for contributing! can you also add a changefile in .changes/ please?

thewh1teagle:
Sure, added new changes file

FabianLars (approved):

PR #1330, opened 2024-05-16
feat(updater): add `download` and `install` js binding

Legend-Master:


Legend-Master:
I think it's better to store downloaded bytes inside of `Update` instead of returning the bytes, that'll be a breaking change though

Legend-Master:
<details>
<summary>I'm using this to test it, sharing it for anyone who wants to test it</summary>

```diff
diff --git a/examples/api/package.json b/examples/api/package.json
index 3312494e..e3c6432f 100644
--- a/examples/api/package.json
+++ b/examples/api/package.json
@@ -23,7 +23,7 @@
     "@tauri-apps/plugin-os": "2.0.0-beta.3",
     "@tauri-apps/plugin-process": "2.0.0-beta.3",
     "@tauri-apps/plugin-shell": "2.0.0-beta.4",
-    "@tauri-apps/plugin-updater": "2.0.0-beta.3",
+    "@tauri-apps/plugin-updater": "file:../../plugins/updater",
     "@zerodevx/svelte-json-view": "1.0.9"
   },
   "devDependencies": {
diff --git a/examples/api/src/views/Updater.svelte b/examples/api/src/views/Updater.svelte
index 819c65e0..7f3fc898 100644
--- a/examples/api/src/views/Updater.svelte
+++ b/examples/api/src/views/Updater.svelte
@@ -27,7 +27,7 @@
     isInstalling = true;
     downloadedSize = 0;
     try {
-      await newUpdate.downloadAndInstall((downloadProgress) => {
+      const bytes = await newUpdate.download((downloadProgress) => {
         switch (downloadProgress.event) {
           case "Started":
             totalSize = downloadProgress.data.contentLength;
@@ -39,6 +39,7 @@
             break;
         }
       });
+      await bytes.install()
       onMessage("Installation complete, restarting...");
       await new Promise((resolve) => setTimeout(resolve, 2000));
       await relaunch();
diff --git a/pnpm-lock.yaml b/pnpm-lock.yaml
index 17cde4df..27eadec0 100644
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -109,8 +109,8 @@ importers:
         specifier: 2.0.0-beta.4
         version: 2.0.0-beta.4
       '@tauri-apps/plugin-updater':
-        specifier: 2.0.0-beta.3
-        version: 2.0.0-beta.3
+        specifier: file:../../plugins/updater
+        version: link:../../plugins/updater
       '@zerodevx/svelte-json-view':
         specifier: 1.0.9
         version: 1.0.9(svelte@4.2.17)
@@ -1026,9 +1026,6 @@ packages:
   '@tauri-apps/plugin-shell@2.0.0-beta.4':
     resolution: {integrity: sha512-y9APB3ZLtkY4IM6syE+dn/R4XR2UTHBcAI/n9L5PgnvOjJAuuEzVYURn10G51nKnLw+sFbxO7LW4VW66Uoaa9g==}
 
-  '@tauri-apps/plugin-updater@2.0.0-beta.3':
-    resolution: {integrity: sha512-bD1ikPz80uK9YJKNYpYlA6StSp9lr0Ob1kGLG2XdmOgspv7SZLLNVzMORtKeqgepxwG99qdYGDDegT3Ll6+UlA==}
-
   '@tauri-apps/toml@2.2.4':
     resolution: {integrity: sha512-NJV/pdgJObDlDWi5+MTHZ2qyNvdL0dlHqQ72nzQYXWbW1LHMPXgCJYl0pLqL1XxxLtxtInYbtVCGVAcwhGxdkw==}
 
@@ -3517,10 +3514,6 @@ snapshots:
     dependencies:
       '@tauri-apps/api': 2.0.0-beta.11
 
-  '@tauri-apps/plugin-updater@2.0.0-beta.3':
-    dependencies:
-      '@tauri-apps/api': 2.0.0-beta.11
-
   '@tauri-apps/toml@2.2.4': {}
 
   '@types/estree@1.0.5': {}
```
</details>

FabianLars:
> I think it's better to store downloaded bytes inside of Update instead of returning the bytes, that'll be a breaking change though

Not sure what exactly you mean tbh but now's the time for breaking changes.

Legend-Master:
What I mean is instead of returning the bytes from `Update::download`, stores it inside `Update` and those bytes are always associated with that `Update` so that we don't need another resource, not sure if it's better though

plugins/updater/guest-js/index.ts
```diff
@@ -55,12 +55,27 @@ class Update extends Resource {
     this.body = metadata.body;
   }
 
+  /** Download the updater package */
+  async download(
+    onEvent?: (progress: DownloadEvent) => void,
+  ): Promise<DownloadedBytes> {
+    const channel = new Channel<DownloadEvent>();
+    if (onEvent !== undefined) {
```
amrbashir: let's just check if truthy intstead of not falsy
```suggestion
    if (onEvent) {
```

plugins/updater/guest-js/index.ts
```diff
@@ -55,12 +55,27 @@ class Update extends Resource {
...
+    return new DownloadedBytes(rid, this);
+  }
+
   /** Downloads the updater package and installs it */
   async downloadAndInstall(
     onEvent?: (progress: DownloadEvent) => void,
   ): Promise<void> {
     const channel = new Channel<DownloadEvent>();
-    if (onEvent != null) {
+    if (onEvent !== undefined) {
```
amrbashir: ditto
```suggestion
    if (onEvent) {
```

plugins/updater/guest-js/index.ts
```diff
@@ -70,6 +85,23 @@ class Update extends Resource {
   }
 }
 
+class DownloadedBytes extends Resource {
```
amrbashir: instead of a new class, let's just store it on Update class, we will probably need to override `Resource` close method to also close the stored bytes if found.

plugins/updater/src/commands.rs
```diff
@@ -75,6 +78,45 @@ pub(crate) async fn check<R: Runtime>(
...
+pub(crate) async fn install<R: Runtime>(
+    webview: Webview<R>,
+    update_rid: ResourceId,
+    bytes_rid: ResourceId,
+) -> Result<()> {
+    let update = webview.resources_table().get::<Update>(update_rid)?;
+    let bytes = webview
+        .resources_table()
+        .get::<DownloadedBytes>(bytes_rid)?;
+    update.install(&bytes.0)?;
```
amrbashir, line 116: we should close the bytes resource after it is installed? it doesn't need to be kept, right?

amrbashir (changes requested):

plugins/updater/guest-js/index.ts
Legend-Master: eslint rule prevents me from doing that, do we want to tweak that eslint rule?

> https://discord.com/channels/616186924390023171/986184094050316358/1240674549637582952
> https://typescript-eslint.io/rules/strict-boolean-expressions/#allownullableobject

plugins/updater/guest-js/index.ts
Legend-Master: Ok, I was concerned about if we change how the resource is cleaned up, we'll need to update it as well, what about we store the `DownloadedBytes` resource to `Update`?

plugins/updater/src/commands.rs
Legend-Master: Yeah, was thinking about it works like this when used in rust side, it doesn't make too much sense for the js binding though

plugins/updater/guest-js/index.ts
amrbashir: I see, just ignore my suggestion

plugins/updater/guest-js/index.ts
```diff
@@ -55,6 +56,27 @@ class Update extends Resource {
...
+    }
+    this.downloadedBytesRid = await invoke<number>("plugin:updater|download", {
+      onEvent: channel,
+      rid: this.rid,
+    });
+  }
+
+  /** Install downloaded updater package */
+  async install(): Promise<void> {
+    await invoke("plugin:updater|install", {
```
amrbashir, line 78: we should check if `this.downloadedBytesRid` is `undefined` and throw an error that `install` is used before `download`

plugins/updater/guest-js/index.ts
amrbashir: sure, let's do `DownloadedBytes` but don't export it

plugins/updater/guest-js/index.ts
```diff
@@ -68,6 +90,15 @@ class Update extends Resource {
       rid: this.rid,
     });
   }
+
+  async close(): Promise<void> {
+    if (this.downloadedBytesRid !== undefined) {
```
amrbashir:
```suggestion
    if (this.downloadedBytesRid) {
```

plugins/updater/guest-js/index.ts
```diff
@@ -68,6 +90,15 @@ class Update extends Resource {
...
+      await invoke("plugin:resources|close", {
+        rid: this.rid,
```
amrbashir:
```suggestion
        rid: this.downloadedBytesRid,
```

I guess it is better to add back your `DownloadedBytes` class and use its close method instead

amrbashir (changes requested):

.changes/updater-download-install-js-binding.md
```diff
@@ -0,0 +1,6 @@
+---
+"updater": "patch"
+"updater-js": "patch"
+---
+
+Add separated `download` and `install` functions to JavaScript side
```
amrbashir:
```suggestion
Add `Update.download` and `Update.install` functions to the JavaScript API
```

plugins/updater/guest-js/index.ts
```diff
@@ -39,12 +39,15 @@ type DownloadEvent =
   | { event: "Progress"; data: { chunkLength: number } }
   | { event: "Finished" };
 
+class DownloadedBytes extends Resource {}
```
amrbashir: I apologize but looks like a new class is not needed after all and we can just use `Resource` class directly instead.

plugins/updater/guest-js/index.ts
```diff
@@ -55,6 +58,32 @@ class Update extends Resource {
...
+      rid: this.rid,
+    });
+    this.downloadedBytes = new DownloadedBytes(downloadedBytesRid);
+  }
+
+  /** Install downloaded updater package */
+  async install(): Promise<void> {
+    if (!this.downloadedBytes) {
+      throw "Update not downloaded yet";
+    }
```
amrbashir:
```suggestion
    if (!this.downloadedBytes) {
      throw "Update.install called before Update.download";
    }

```

plugins/updater/guest-js/index.ts
```diff
@@ -55,6 +58,32 @@ class Update extends Resource {
...
+    await invoke("plugin:updater|install", {
+      updateRid: this.rid,
+      bytesRid: this.downloadedBytes.rid,
+    });
+    // Don't need to call close, we did it in rust side already
+    this.downloadedBytes = undefined;
```
amrbashir, line 84:
```suggestion

    // Don't need to call close, we did it in rust side already
    this.downloadedBytes = undefined;
```

amrbashir (changes requested):

amrbashir (approved):

amrbashir:
Thank you

PR #1348, opened 2024-05-20
feat(upload): Added post to download function

enri90:
Added post functionality to the download function

Example
```typescript
const body = JSON.stringify({ ids: ids });

await download(
  url,
  path,
  ({ progress, total }) => console.log(`Downloaded ${progress} of ${total} bytes`),
  new Map<string, string>([
    ['Content-Type', 'application/json'],
  ]),
  body
);
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "@rollup/plugin-terser": "0.4.4",
     "@rollup/plugin-typescript": "11.1.6",
     "@types/eslint__js": "8.42.3",
-    "covector": "^0.11.0",
+    "covector": "^0.12.0",
     "eslint": "9.6.0",
     "eslint-config-prettier": "9.1.0",
     "eslint-plugin-security": "3.0.1",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -28,8 +28,8 @@ importers:
         specifier: 8.42.3
         version: 8.42.3
       covector:
-        specifier: ^0.11.0
-        version: 0.11.0(mocha@10.5.2)
+        specifier: ^0.12.0
+        version: 0.12.0(mocha@10.5.2)
       eslint:
         specifier: 9.6.0
         version: 9.6.0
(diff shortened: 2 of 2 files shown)


plugins/upload/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "tauri-plugin-upload"
-version = "2.0.0-beta.5"
+version = "2.0.0-beta.6"
```
FabianLars: revert this please. Our publishing pipeline will do this.

plugins/upload/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@tauri-apps/plugin-upload",
-  "version": "2.0.0-beta.4",
+  "version": "2.0.0-beta.5",
```
FabianLars: same

plugins/upload/guest-js/index.ts
```diff
@@ -60,6 +61,7 @@ async function download(
     filePath,
     headers: headers ?? {},
     onProgress,
+    body: body ?? "",
```
FabianLars: I think it's fine to just send null and use `Option<&str>` in Rust. Feels a bit cleaner but i don't really care that much.

plugins/upload/src/lib.rs
```diff
@@ -66,10 +66,14 @@ async fn download(
     file_path: &str,
     headers: HashMap<String, String>,
     on_progress: Channel,
+    body: &str,
 ) -> Result<()> {
     let client = reqwest::Client::new();
-
-    let mut request = client.get(url);
+    let mut request = if !body.is_empty() {
+        client.post(url).body(body.to_owned())
```
FabianLars: If we need an owned body var then you can just use `body: String` as the function argument.

FabianLars (commented):
Thank you for contributing :) I added a few minor comments.

We'll also need a changefile (see the .changes folder for examples).

Do you think it makes sense to have the http request type configurable? Or should get and post cover it good enough for now

plugins/upload/guest-js/index.ts
enri90: I tried with Option< String > in Rust but it gives me problems on condition

enri90:
> Thank you for contributing :) I added a few minor comments.
> 
> We'll also need a changefile (see the .changes folder for examples).
> 
> Do you think it makes sense to have the http request type configurable? Or should get and post cover it good enough for now

I need the post to send data when generating a PDF

PR #1379, opened 2024-05-28
fix(clipboard-manager): match Rust method parameters with native methods

unbeauvoyage:
Ensured that the parameters for the run_mobile_plugin method in Rust code's readText and writeText methods match the expected native method names.

Below is a screenshot of one of the errors fixed:
image

Leaving the other error here as a reference for anyone text searching anything similar:
No command write found for plugin clipboard-manager


amrbashir (approved):

amrbashir:
thank you

PR #1404, opened 2024-06-02
fix: typo

simonhyll:


FabianLars:
can you run `pnpm build`? It should update the api-iife.js file too i think 🤔

amrbashir:
should add a change file as well

FabianLars (approved):

PR #1412, opened 2024-06-03
chore: Update eslint to v9 with flat config

FabianLars:


FabianLars:
why is it not failing locally, u kiddin me?

PR #1450, opened 2024-06-10
fix(notification): Add `register_action_types` permission command

naman-crabnebula:
- Added `register_action_types` to Permission Commands
- Now, new Action types can be registered on Android
- This is necessary path for working of Android actions


FabianLars:
I'm not sure how the permissions work on mobile yet but from the logic of this PR, won't we need the same change for all the other commands in <https://github.com/tauri-apps/plugins-workspace/blob/v2/plugins/notification/guest-js/index.ts>?

naman-crabnebula:
Hi! These are auto-generated files. 

Edit : Wait, yes, I am adding them as per needed, I am converting this to draft PR

PR #1454, opened 2024-06-11
fix(updater): use msiexec to install .msi updates

### LINKED ISSUE #1449: Updater ISSUE in Windows Installer .msi
The updater on Windows instead of running the installation, the Windows Installer window appears with the **arguments**.

Image (Displayed in italian):
image

amrbashir:
closes #1449

### DIFF
--- a/.changes/updater-msiexec.md
+++ b/.changes/updater-msiexec.md
@@ -1,0 +1,5 @@
+---
+"updater": "patch"
+---
+
+Fix regression in updater plugin failing to update using `.msi` installer.
--- a/plugins/updater/src/config.rs
+++ b/plugins/updater/src/config.rs
@@ -49,7 +49,7 @@ impl Display for WindowsUpdateInstallMode {
             f,
             "{}",
             match self {
-                Self::BasicUi => "basicUI",
+                Self::BasicUi => "basicUi",
                 Self::Quiet => "quiet",
                 Self::Passive => "passive",
             }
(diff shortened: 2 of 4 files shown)


FabianLars:
One thing i just noticed because of your nsis_installer_args change is that we never ported the std::env::args change to the v2 msi updater. Maybe we can use a public property for that to tell the msi about them (doesn't have to be this pr). On that note, can it even restart the app right now? We've used powershell for that in the past. Just on my phone rn so didn't test anything (including this pr)

amrbashir:
The restart thing (and consecutively the current args) and the upgrade using TaskScheduler, were removed because we don't use powershell anymore. This is out of this PR scope, so I'd rather discuss this on Discord or in another issue but the generated installer should have a flag or something that could be used to launch the application afterwards.

I am not sure if it is possible but we plan to phase out msi installer anyways since no one in the team really wants to maintain it so might not worth the effort.

Edit: https://stackoverflow.com/questions/14762617/wix-bootstrapper-how-do-i-set-burn-variables-from-the-command-line

FabianLars (approved):
still didn't test it but can't be worse than the current version and the code looks good to me 🤷

songjiachao:
when publish the next version

nwehrhan:
Hmm I still can't `msi` to actually do the update. Says "success".
https://discord.com/channels/616186924390023171/1253345571599814666/1253345571599814666

PR #1458, opened 2024-06-12
feat(sql) Sql batch insert

eleroy:
Hello,

This a pull request for implementing an additional method for batch inserting into a SqlLite database (should also work for other sql databases).

It is based on the [QueryBuild push_values method](https://docs.rs/sqlx/latest/sqlx/struct.QueryBuilder.html#method.push_values) of sqlx rust module.

It is possible to push any number of values but the JS function will split the request into batches of values that does not exceed 999 values since it seems to be the limit for sqlite.

From my benchmark it is 100 times faster than looping through execute commands (with 10000 values, but maybe my benchmark code is not the best). Here is the code I use for testing : 

```js
let db = await Database.load('sqlite:mydatabase.db')
let data = []
for(let i= 0;i<10000;i++){
data.push([(Math.random() + 1).toString(36).substring(7)])
}
let startTime = Date.now()
await db.batchInsert("INSERT into users (name) ", data)
console.log("Batch insert time ", Date.now()-startTime)

startTime = Date.now()
await Promise.all(data.map(v=> db.execute("INSERT into users (name) VALUES ($1)", v)))
console.log("Loop insert time ", Date.now()-startTime)
```
```console
Batch insert time  190
Loop insert time  31462
```

The results for 1000 values:
```
Batch insert time  11
Loop insert time  557
```

### DIFF
--- a/.changes/sql-batch-insert.md
+++ b/.changes/sql-batch-insert.md
@@ -1,0 +1,5 @@
+---
+"sql": minor
+---
+
+Add support for batch inserting a large number of values in the database.
--- a/plugins/sql/README.md
+++ b/plugins/sql/README.md
@@ -105,6 +105,30 @@ const result = await db.execute(
 );
 ```
 
+### Batch Insert
+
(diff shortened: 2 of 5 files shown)


FabianLars:
Hey, thanks for contributing! Just one thing, i don't think we should care about the chunk size in the plugin code as it may be different per config and/or per database type. In my opinion this should be something the user should worry about / can decide how to handle (a docs comment is totally fine for that).

eleroy:
Hi,

No problem, I reverted the js method to a simpler form and added a note in the Readme.

PR #1462, opened 2024-06-14
feat(clipboard-manager): implement `clear` on iOS and Android

unbeauvoyage:
This PR contains iOS and Android clear clipboard functionality that I use in my own project and shared here. Below is a screenshot that contains iOS simulator demonstration (as best as console logs could demonstrate) that it functions in the intended manner, and below that is the Android emulator version.

iOS simulator test:

image

Android emulator test:

image

Also, the below when missed the else part. Not that it was causing errors for now, other than the IDE error, but it might in the future. Thinking it might not be worth a separate PR, I included it in this one. If I should open another one for that, I can send it in a separate PR and take it out of this one.

```kotlin
    val clipData = when (args) {
      is WriteOptions.PlainText -> {
        ClipData.newPlainText(args.label, args.text)
      } else -> {
        invoke.reject("Invalid write options provided")
        return
      }
```

### DIFF
--- a/plugins/clipboard-manager/android/src/main/java/ClipboardPlugin.kt
+++ b/plugins/clipboard-manager/android/src/main/java/ClipboardPlugin.kt
@@ -129,9 +129,7 @@ class ClipboardPlugin(private val activity: Activity) : Plugin(activity) {
   fun clear(invoke: Invoke) {
       if (manager.hasPrimaryClip()) {
           manager.clearPrimaryClip()
-          invoke.resolve()
-      } else {
-          invoke.reject("Clipboard is empty")
       }
+      invoke.resolve()
   }
 }
--- a/plugins/clipboard-manager/android/src/main/java/ClipboardPlugin.kt
+++ b/plugins/clipboard-manager/android/src/main/java/ClipboardPlugin.kt
@@ -92,7 +92,11 @@ class ClipboardPlugin(private val activity: Activity) : Plugin(activity) {
     val clipData = when (args) {
       is WriteOptions.PlainText -> {
         ClipData.newPlainText(args.label, args.text)
+      } else -> {
+        invoke.reject("Invalid write options provided")
+        return
       }
+
     }
 
     manager.setPrimaryClip(clipData)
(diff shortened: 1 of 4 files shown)


plugins/clipboard-manager/ios/Sources/ClipboardPlugin.swift
```diff
@@ -38,6 +38,16 @@ class ClipboardPlugin: Plugin {
       invoke.reject("Clipboard is empty")
     }
   }
+    
+  @objc public func clear(_ invoke: Invoke) throws {
+      let clipboard = UIPasteboard.general
+      if clipboard.items.isEmpty {
+          invoke.reject("Clipboard is empty")
```
amrbashir: I don't think we really need to return with an Error here, we should resolve with success if it is already cleared

plugins/clipboard-manager/android/src/main/java/ClipboardPlugin.kt
```diff
@@ -120,4 +124,14 @@ class ClipboardPlugin(private val activity: Activity) : Plugin(activity) {
...
   }
+
+  @Command
+  fun clear(invoke: Invoke) {
+      if (manager.hasPrimaryClip()) {
+          manager.clearPrimaryClip()
+          invoke.resolve()
+      } else {
+          invoke.reject("Clipboard is empty")
+      }
```
amrbashir: same, we shouldn't resolve with error here

amrbashir (changes requested):
Thanks for your contributions, just 2 small nit-picks and this should be ready to go. Could you also add a change file in `.changes` directory

amrbashir (dismissed):

amrbashir:
Thank you

plugins/clipboard-manager/src/mobile.rs
```diff
@@ -92,9 +92,9 @@ impl<R: Runtime> Clipboard<R> {
     }
 
     pub fn clear(&self) -> crate::Result<()> {
-        Err(crate::Error::Clipboard(
-            "Unsupported on this platform".to_string(),
-        ))
+        self.0
+            .run_mobile_plugin("clear", ())
+            .map_err(Into::into)
```
amrbashir:
```suggestion
        self.0.run_mobile_plugin("clear", ()).map_err(Into::into)
```

amrbashir (approved):

PR #1473, opened 2024-06-17
fix(fs): Make commands async to not block the main thread so long

CaptainTux:
This PR fixes the main thread blocking on disk-write operations.
I raised the issue [here](https://discord.com/channels/616186924390023171/1250912842908045472/1250912842908045472) and this PR solves that issue.


FabianLars:
I'll reopen this under my name so that the covector status PR doesn't fail anymore x) Also the branch isn't complete yet :)

PR #1497, opened 2024-06-25
feat(window-state): add `Builder::map_label` method

ferreira-tb:
This PR implements [amrbashir's suggestion](https://github.com/tauri-apps/plugins-workspace/pull/531#pullrequestreview-1566282300) made when reviewing [this other PR](https://github.com/tauri-apps/plugins-workspace/pull/531).

My use case is incredibly similar to that of the [author of that other PR](https://github.com/tauri-apps/plugins-workspace/issues/507#issuecomment-1644857633), as I'm also working on a comic reader app.

This is my first code PR to the project, so I probably did some mistakes.

This change is already being tested on my WIP app, as can be seen [here](https://github.com/ferreira-tb/kotori/blob/ab6fe9bd1f0d52e5a34d7047b09fda6235e5c241/src-tauri/src/main.rs#L36).

### DIFF
--- a/.changes/window-state-map-label.md
+++ b/.changes/window-state-map-label.md
@@ -2,4 +2,4 @@
 "window-state": patch
 ---
 
-Add `map_label` option to transform the window label before using it internally, based on [amrbashir's suggestion](https://github.com/tauri-apps/plugins-workspace/pull/531#pullrequestreview-1566282300).
+Add `Builder::map_label` option to transform the window label when saving the window state, this could be used to group different windows to use the same state.
--- a/.changes/window-state-map-label.md
+++ b/.changes/window-state-map-label.md
@@ -1,0 +1,5 @@
+---
+"window-state": patch
+---
+
+Add `map_label` option to transform the window label before using it internally, based on [amrbashir's suggestion](https://github.com/tauri-apps/plugins-workspace/pull/531#pullrequestreview-1566282300).
--- a/plugins/window-state/src/lib.rs
+++ b/plugins/window-state/src/lib.rs
@@ -28,6 +28,8 @@ use std::{
 
 mod cmd;
 
+type LabelMapperFn = Fn(&str) -> &str + Send + Sync;
+
 /// Default filename used to store window state.
 ///
 /// If using a custom filename, you should probably use [`AppHandleExt::filename`] instead.
@@ -65,7 +67,7 @@ impl Default for StateFlags {
 
 struct PluginState {
     filename: String,
-    map_label: Option<Box<dyn Fn(&str) -> &str + Send + Sync>>,
+    map_label: Option<Box<dyn LabelMapperFn>,
 }
 
 #[derive(Debug, Deserialize, Serialize, PartialEq)]
@@ -326,7 +328,7 @@ pub struct Builder {
     denylist: HashSet<String>,
     skip_initial_state: HashSet<String>,
     state_flags: StateFlags,
-    map_label: Option<Box<dyn Fn(&str) -> &str + Send + Sync>>,
+    map_label: Option<Box<dyn LabelMapperFn>>,
     filename: Option<String>,
 }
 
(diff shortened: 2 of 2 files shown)


plugins/window-state/src/lib.rs
```diff
@@ -342,9 +351,15 @@ impl Builder {
         self
     }
 
+    pub fn map_label(mut self, map_fn: fn(&str) -> &str) -> Self {
```
amrbashir: missing documentation

plugins/window-state/src/lib.rs
```diff
@@ -309,6 +317,7 @@ pub struct Builder {
     denylist: HashSet<String>,
     skip_initial_state: HashSet<String>,
     state_flags: StateFlags,
+    map_label: Option<Box<dyn (Fn(&str) -> &str) + Send + Sync>>,
```
amrbashir: extra parentheses is not needed 
```suggestion
    map_label: Option<Box<dyn Fn(&str) -> &str + Send + Sync>>,
```

plugins/window-state/src/lib.rs
amrbashir: Let's use a generic instead of a function pointer to allow using closures
```suggestion
    pub fn map_label<F>(mut self, map_fn: F) -> Self 
    where
            F: Fn(&str) -> &str + Sync + Send,
    {
```

amrbashir (changes requested):
Thanks for the PR, We also need to update `save_window_state` function logic to accommodate these changes

.changes/window-state-map-label.md
```diff
@@ -0,0 +1,5 @@
+---
+"window-state": patch
+---
+
+Add `map_label` option to transform the window label before using it internally, based on [amrbashir's suggestion](https://github.com/tauri-apps/plugins-workspace/pull/531#pullrequestreview-1566282300).
```
amrbashir:
```suggestion
Add `Builder::map_label` option to transform the window label when saving the window state, this could be used to group different windows to use the same state.
```

amrbashir (changes requested):

ferreira-tb:
Thanks for the review.

I really overlooked the `save_window_state` function, sorry about that. I updated it now, but I need opinions on whether my approach was appropriate.

About style: when using tuples, it is preferable to access by indexes, i.e. `|(x)| x.1`? Is something like `|(y, _)| y` appropriate?

plugins/window-state/src/lib.rs
```diff
@@ -114,10 +115,22 @@ impl<R: Runtime> AppHandleExt for tauri::AppHandle<R> {
...
             let mut state = cache.0.lock().unwrap();
+
             for (label, s) in state.iter_mut() {
-                if let Some(window) = self.get_webview_window(label) {
+                let window = match &plugin_state.map_label {
+                    Some(map) => {
+                        windows
+                            .iter()
+                            .find(|(l, _)| map(l) == label)
+                            .map(|(_, window)| window)
```
amrbashir:
```suggestion
                            .find_map(|(l, window)| (map(l) == label).then_some(window))
```

Note: you may want to run `cargo fmt` as this suggestion is not properly formatted

amrbashir (changes requested):

plugins/window-state/src/lib.rs
ferreira-tb: That looks way cleaner. Thanks.

plugins/window-state/src/lib.rs
```diff
@@ -65,6 +67,7 @@ impl Default for StateFlags {
 
 struct PluginState {
     filename: String,
+    map_label: Option<Box<dyn LabelMapperFn>,
```
amrbashir:
```suggestion
    map_label: Option<Box<dyn LabelMapperFn>>,
```

amrbashir (approved):
Thanks for contribution

PR #1498, opened 2024-06-25
fix(updater): Start app after update via msi

FabianLars:
requires https://github.com/tauri-apps/tauri/pull/10117

### DIFF
--- a/.changes/fix-updater-msi-autostart.md
+++ b/.changes/fix-updater-msi-autostart.md
@@ -2,4 +2,4 @@
 updater: patch
 ---
 
-Fixed an issue that caused the app to not be started automatically after an update if installed via `.msi`.
+Automatically launch app after updates using `.msi`, to match NSIS `.exe` installer behaviour.
--- a/.changes/fix-updater-msi-autostart.md
+++ b/.changes/fix-updater-msi-autostart.md
@@ -1,0 +1,5 @@
+---
+updater: patch
+---
+
+Fixed an issue that caused the app to not be started automatically after an update if installed via `.msi`.
--- a/plugins/updater/src/updater.rs
+++ b/plugins/updater/src/updater.rs
@@ -538,16 +538,28 @@ impl Update {
 
 #[cfg(windows)]
 enum WindowsUpdaterType {
-    Nsis,
-    Msi,
+    Nsis {
+        path: PathBuf,
+        #[allow(unused)]
+        temp: Option<tempfile::TempPath>,
+    },
+    Msi {
+        path: PathBuf,
+        #[allow(unused)]
+        temp: Option<tempfile::TempPath>,
+    },
 }
 
 #[cfg(windows)]
 impl WindowsUpdaterType {
-    fn extension(&self) -> &str {
-        match self {
-            WindowsUpdaterType::Nsis => ".exe",
-            WindowsUpdaterType::Msi => ".msi",
+    fn nsis(path: PathBuf, temp: Option<tempfile::TempPath>) -> Self {
+        Self::Nsis { path, temp }
+    }
+
+    fn msi(path: PathBuf, temp: Option<tempfile::TempPath>) -> Self {
+        Self::Msi {
+            path: path.wrap_in_quotes(),
+            temp,
         }
     }
 }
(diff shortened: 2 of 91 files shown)


FabianLars:
Ah forgot to test if quotes in cli args will break stuff - sounds pretty likely. May check that tomorrow (today) but honestly don't really care

plugins/updater/src/updater.rs
```diff
@@ -587,6 +587,27 @@ impl Update {
...
+
+            if arg.starts_with('-') {
+                if let Some((a1, a2)) = arg.split_once('=') {
+                    msi_args.push(format!("{a1}=\"\"{a2}\"\""));
+                } else {
+                    msi_args.push(format!("\"\"{arg}\"\""));
+                }
+            } else {
+                msi_args.push(format!("\"\"{arg}\"\""));
+            }
```
amrbashir: this is some wild code, let's move it into a separate function and add extensive unit tests for it

.changes/fix-updater-msi-autostart.md
```diff
@@ -0,0 +1,5 @@
+---
+updater: patch
+---
+
+Fixed an issue that caused the app to not be started automatically after an update if installed via `.msi`.
```
amrbashir:
```suggestion
Automatically launch app after updates using `.msi`, to match NSIS `.exe` installer behaviour.
```

amrbashir (changes requested):

FabianLars:
wtf is the fmt ci check doing 😡

plugins/updater/src/updater.rs
FabianLars: The fun stuff for the tests will be to figure out what's actually the intended result lol

FabianLars: Like this, or would you prefer a function that returns the full LAUNCHAPPARGS string (rather than a vec of escaped strings)

plugins/updater/src/updater.rs
```diff
@@ -1026,6 +1050,40 @@ impl PathExt for PathBuf {
     }
 }
 
+#[cfg(windows)]
+fn escape_msi_property_args(args: &[&OsStr]) -> Vec<String> {
```
amrbashir: I think this should operate on a single value, and return the escaped value and leave the loop stuff out at the calling site, it could just even be used with `iterator.map()`

plugins/updater/src/updater.rs
amrbashir: left a comment below

plugins/updater/src/updater.rs
FabianLars: ok good idea

plugins/updater/src/updater.rs
```diff
@@ -594,6 +594,32 @@ impl Update {
...
+            if arg.starts_with('-') {
+                if let Some((a1, a2)) = arg.split_once('=') {
+                    msi_args.push(format!("{a1}=\"\"{a2}\"\""));
+                } else {
+                    msi_args.push(format!("\"\"{arg}\"\""));
+                }
+            } else {
+                msi_args.push(format!("\"\"{arg}\"\""));
+            }
+        }
```
amrbashir: this is no longer needed

plugins/updater/src/updater.rs
```diff
@@ -1026,6 +1054,32 @@ impl PathExt for PathBuf {
     }
 }
 
+#[cfg(windows)]
+fn escape_msi_property_arg(arg: impl AsRef<OsStr>) -> String {
+    let mut arg = arg.as_ref().to_string_lossy().to_string();
+
+    // Otherwise this argument will get lost in ShellExecute
+    if arg.is_empty() {
```
amrbashir, line 1041: if an arg.is_empty, it should be skipped no?

FabianLars: Nope. This should be a pretty rare case but for example if the user starts the initial app with `./myapp '""'` arg.is_empty() will be true and really just an empty rust string. If we forward it to ShellExecute we'll and up with literally nothing (no argument) since we're appending an empty string to a string. Therefore we need to wrap the empty string in `"{empty string}"` but since we're inside the msi prop we need it to be `""{empty string}""`. Only then will the new app instance get an empty string in `env::args()` like the initial instance.

amrbashir (approved):

amrbashir:
seems like integration tests failing are not related to this PR

PR #1500, opened 2024-06-27
docs: use join in readDir documentation example

mordof:
parent + entry.name just mashes two names together without path delimiters. adding join() in makes the dir a proper path


amrbashir (approved):

amrbashir:
Thank you

PR #1502, opened 2024-06-27
fix(notification): Fix is-installed logic to prevent issue when a custom build.target-dir is set

FabianLars:
The old logic failed because exe_dir was `A:\_target\debug` on my system.

I also tried to read to read the `shell:AppsFolder` folder to support more custom installers but couldn't figure out how to read a virtual folder.

Ideally i'd like to look into lifting the installation requirement but that seems to be a bit more work than this simple PR.

### DIFF
--- a/.changes/fix-notification-install-detection.md
+++ b/.changes/fix-notification-install-detection.md
@@ -1,0 +1,5 @@
+---
+notification: patch
+---
+
+Fixed an issue that prevented notifications from showing up on Windows in dev mode when a custom `build.target-dir` was set.
--- a/plugins/notification/src/desktop.rs
+++ b/plugins/notification/src/desktop.rs
@@ -62,9 +62,6 @@ impl<R: Runtime> Notification<R> {
 mod imp {
     //! Types and functions related to desktop notifications.
 
-    #[cfg(windows)]
-    use std::path::MAIN_SEPARATOR as SEP;
-
     /// The desktop notification definition.
     ///
     /// Allows you to construct a Notification data and send it.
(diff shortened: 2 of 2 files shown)


amrbashir (changes requested):
I don't think we should check for production installation, instead we should keep checking for debug builds or rather uninstalled builds from cargo.

I'd recommend just checking if there is `.pdb` file next to the executable and if `.cargo-lock` exists

FabianLars:
The idea was to make notifications _somewhat_ working for portible builds and/or custom installers. Currently it doesn't show up at all, with this change it would at least show up as powershell.

Otherwise we might as well use `tauri::is_dev()`

amrbashir:
> The idea was to make notifications _somewhat_ working for portible builds and/or custom installers. Currently it doesn't show up at all, with this change it would at least show up as powershell.

What if the portable build was published without `uninstall.exe` next to it? this PR won't work either.

> Otherwise we might as well use `tauri::is_dev()`

I am fine with using `tauri::is_dev()` tbh

amrbashir:
I did a bit of investigating for `AppUserModelId` and here is what I found:

- `AppUserModelId` can be set using `SetCurrentProcessExplicitAppUserModelID` but it doesn't take effect unless there is a shortcut in Start Menu (doesn't need to be pinned)

- We can get notifications to work by pinning the executable to Start Menu (Right Click, choose Pin to Start) or creating a simple shortcut in `%AppData%\Microsoft\Windows\Start Menu\Programs` but when we are showing the notification, we will use the full path to the executable as `AppUserModelId` for the notification. Can be as simple as calling `std::env::current_exe()`.


So, I am recommending:

1. Document this notification quirks in the plugin README and subsequently in our documentation site similar to what Electron did.
2. use `tauri::is_dev()` and always use PowerShell id in development.
3. Expose `app_user_model_id` in the notification plugin api
4. We should also add snippet for portable builds showcasing:
   - how to create a shortcut in `%AppData%\Microsoft\Windows\Start Menu\Programs`
   - how to use the new `app_user_model_id` to get notifications working i.e calling it with `std::env::current_exe()`.
   - [Optional] cleanup shortcut on exist
5. [Optional] Document how users can get their app icon in development:
   - show them how they can pin their exe to start
   - how to use the new `app_user_model_id` to get notifications working i.e calling it with `std::env::current_exe()`.


   
References:
- https://learn.microsoft.com/en-us/windows/win32/shell/enable-desktop-toast-with-appusermodelid
- https://www.electronjs.org/docs/latest/tutorial/notifications#windows

FabianLars:
> What if the portable build was published without uninstall.exe next to it? this PR won't work either.

Before this PR no notification will be shown at all, with this PR a notification with the PowerShell ID will be shown, which if you ask me is better then nothing. This PR was really just meant to be a temporary improvement over the current logic until we can figure out something better (the stuff you researched).

amrbashir:
I see, but I don't think we should have this temporarily and should implement the suggestions I posted above.

PR #1521, opened 2024-07-06
change!(updater): disable zip feature by default

Legend-Master:
As we made a new option in tauri CLI in https://github.com/tauri-apps/tauri/pull/9883, I think we should remove zip from updater plugin's default features, and make people opt-in instead of opt-out

### DIFF
--- a/.changes/updater-zip-default-false.md
+++ b/.changes/updater-zip-default-false.md
@@ -1,0 +1,5 @@
+---
+"updater": "patch:breaking"
+---
+
+Disable zip feature by default, enable this if you need it to be able to use v1 compatible zipped updaters
--- a/plugins/updater/Cargo.toml
+++ b/plugins/updater/Cargo.toml
@@ -47,7 +47,7 @@ tar = "0.4"
 flate2 = "1"
 
 [features]
-default = [ "rustls-tls", "zip" ]
+default = [ "rustls-tls" ]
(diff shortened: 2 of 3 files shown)


FabianLars:
should probably be part of `tauri migrate` now that you mention it.

Legend-Master:
Made a PR

amrbashir:
This should wait until v3, just to give users (who wouldn't necessarily use tauri migrate) an easier upgrade.

Legend-Master:
I don't think this makes a huge difference as we have made this into a plugin, the user either has to use our cli's migrate or go and look up the updater plugin's docs/readme, and this is just an extra line (the user needs to learn what v1 compatible updater is anyway)

In the other hand, for people coming from v2 or started using v2 updaters, they'll need to do no default features which really isn't that great

amrbashir:
> In the other hand, for people coming from v2 or started using v2 updaters, they'll need to do no default features which really isn't that great

That could be argued both ways.

In fact, I think it is better to opt-out of zip instead of opt-in as it allows more flexibility in terms of supporting more features and easier upgrade for current users. I am fine with making this opt-in in v3, because by then "easier upgrade for current users" won't factor in as much.

Legend-Master:
Alright, the zip crate without any compression enabled shouldn't be too big, I don't quite like it but I don't think I want to fight for it either, closing it for now and we can reopen it on v3

PR #1523, opened 2024-07-06
feat(upload): Added body to download function

enri90:
Added post functionality to the download function
Example

```typescript
const body = JSON.stringify({ ids: ids });

await download(
  url,
  path,
  ({ progress, total }) => console.log(`Downloaded ${progress} of ${total} bytes`),
  new Map<string, string>([
    ['Content-Type', 'application/json'],
  ]),
  body
);

```


FabianLars:
Sorry for the long delay in your first PR. What do you think about leveraging the native [`Request` object](https://developer.mozilla.org/en-US/docs/Web/API/Request) for the url and body?

Basically mirroring the http plugin: <https://github.com/tauri-apps/plugins-workspace/blob/v2/plugins/http/guest-js/index.ts#L104>

Let me know what you think and whether you'd be open to work on it (if you agree) or if i should take over :)

enri90:
> Sorry for the long delay in your first PR. What do you think about leveraging the native [`Request` object](https://developer.mozilla.org/en-US/docs/Web/API/Request) for the url and body?
> 
> Basically mirroring the http plugin: https://github.com/tauri-apps/plugins-workspace/blob/v2/plugins/http/guest-js/index.ts#L104
> 
> Let me know what you think and whether you'd be open to work on it (if you agree) or if i should take over :)

Hi! Thank you for the feedback. 
I understand the request to leverage the native `Request` object for the URL and body and think it's a good idea. 
However, I'm not very familiar with Rust and am not sure how to implement this. 
If you could take over this part, that would be fantastic.
Thanks!

FabianLars:
Okay sure, no problem!

FabianLars (dismissed):

FabianLars:
i changed my mind, let's keep it simple for now.

FabianLars (approved):

PR #1541, opened 2024-07-13
covector published pkg check with built-in API call

jbolda:
## Motivation

Covector has built-in functionality now to fetch an API response in the version check. Switching to this and deleting the previously use script that handled it.


FabianLars (dismissed):

jbolda:
Heh, yea sorry, switched it back to draft too. What started as a benign change revealed a couple spots we missed in a huge refactor. This was the first repo with `pre` enabled.

FabianLars (approved):

PR #1544, opened 2024-07-14
enhance(updater): use named tempfile on Windows

Legend-Master:
This makes the UAC prompt for unsigned installers a bit easier to understand (e.g. `My Tauri App-0.1.0-installer.exe` instead of `.tmpnDqw2l.exe`) and also might help somewhat for the anitivirus false alarms


FabianLars:
Nice change! Can we document on the UpdaterBuilder::new function that users should prefer the updater_builder() function?

Legend-Master:
I never realized you can construct this thing yourself 😂

FabianLars (approved):

FabianLars:
Yeah, changes like these make me wish that wouldn't be the case x)

Issue #15, opened 2022-09-14
[store] Question: Why not keep the store cache in the WebView process?


rmburg:
I'm currently developing a program that uses this plugin.
Currently, if I need a value from the store for the user interface, like for example the user's preferred theme,
I need to call an *asynchronous* function that returns a promise which resolves to the requested value.

This leads to a lot of unnecessary boilerplate code, because I can't "just use" the store values, but have to add a wrapper to each part of my code that uses a store value to abstract away the async nature of the store.
This leads to ugly, unnecessary code and lots of frustration.

The contents of the store file are currently cached by the plugin in the main rust process.
Having the content of the store cached in the WebView process would solve this:
Other parts of the user interface code can immediately use the values without needing "async wrappers".

I propose a different approach, which looks like this:
- Expose a single async function in the WebView, which loads the store file initially
- All other get() and set() calls can be synchronous.
- Cache the store in the WebView
- Register event handlers to update the cache or the file on disk if either changes

Benefits: 
- Avoid async boilerplate
- Users can define a scheme of their store in Typescript, get and set functions would thus be automatically typed without needing casts or assertions in user code
- Less IPC calls

Performance should not be affected by this, as the store is usually called infrequently and IPC calls are reduced.

Also, take a look at the [electron-store](https://github.com/sindresorhus/electron-store) library, which I have used in the past and does exactly what I want.

[...]

PR #1550, opened 2024-07-15
feat(store)!: fully rework and add auto save

### LINKED ISSUE #1535: [store] support auto save with debounce
I'm thinking it would be nice to have a built-in auto save feature for the store plugin, maybe a setting to enable a store to debounce save on modification (on calls like `set`, `clear`, `delete`, `reset`)

Legend-Master:
Closes #1535

Rework and add a setting to enable a store to debounce save on modification (on calls like set, clear, delete, reset)


Legend-Master:
To be honest, I wonder why `with_store` works in this way, I'm thinking it should be creating a `Resource` and with that we can store the configs in Rust side so that we don't pass in `path` and `auto_save` in every call

FabianLars:
> To be honest, I wonder why with_store works in this way, I'm thinking it should be creating a Resource and with that we can store the configs in Rust side so that we don't pass in path and auto_save in every call

1) with_store is older than the Resource api
2) Feel free to rewrite the whole impl/api 😂 (i'm serious, the plugin probably deserves it having requests like https://github.com/tauri-apps/plugins-workspace/issues/15 in mind)

Legend-Master:
> with_store is older than the Resource api

😮 makes sense then, ~~I could probably do it in this PR~~ maybe another PR

I don't think we can do synchronous commands right now with Tauri's IPC? The way we currently do thing is having something in the Rust side and give js side the ability to interact with it but things will stay in the Rust side, #15 means we either need the ability to do synchronous IPC or we reverse the process and make rust side request data from the js side

FabianLars:
> I don't think we can do synchronous commands right now with Tauri's IPC?

Nope, and i doubt we'll see this any time soon.

> The way we currently do thing is having something in the Rust side and give js side the ability to interact with it but things will stay in the Rust side, https://github.com/tauri-apps/plugins-workspace/issues/15 means we either need the ability to do synchronous IPC or we reverse the process and make rust side request data from the js side

I mean, it's totally fine to not want to "solve" #15 i just think we never thought much about it.

.changes/store-auto-save.md
```diff
@@ -0,0 +1,7 @@
+---
+"store": patch
+---
+
+Add a setting `auto_save` to enable a store to debounce save on modification (on calls like set, clear, delete, reset)
+
+**Breaking change**: `with_store` now takes one more parameter `auto_save: Option<Duration>`
```
amrbashir: It feels like `with_store` is growing its original need and requiring too many paramters is not a good DX and will confuse users. I never like this function and I think it should only be used internally. We should look into replacing it with an extension trait, similar to `app.updater_builder()` and other plugins.

Legend-Master: An app handle extension would be nice, do you have any idea about what that API might be?

Since we have more options about constructing the store now, maybe separate two different functions one being `init_store` which init the store with settings and put them into the app's managed state and another one being `with_store` which doesn't init the store but simply lock the collection and give the function access to the requested store

amrbashir: I honestly don't have an idea in mind now, but don't be afraid to break the whole API design if you have a better API.

FabianLars: rewriting it from scratch is totally fine too (if it makes sense). Since this is a v1 plugin it is _not_ on track for a quick stable release (with audit etc).

plugins/store/src/store.rs
```diff
@@ -290,6 +307,69 @@ impl<R: Runtime> Store<R> {
...
+                        let collection = app.state::<StoreCollection<R>>();
+                        if let Some(store) = collection
+                            .stores
+                            .lock()
+                            .expect("mutex poisoned")
+                            .values_mut()
+                            .find(|store| store.path == path)
+                        {
+                            let _ = store.save();
+                            store.auto_save_debounce_sender = None;
```
Legend-Master: This right now only works if the store is managed by `app.state`, if we want it to work with plain `Store` we probably need to wrap the inner in an `Arc<Mutex>` which I don't know if it's worth it

.changes/store-auto-save.md
Legend-Master: I'm thinking about something but don't have a good design yet, maybe convert it to an issue and do it in the future, it's kinda burning my mind right now 😂

```rust
pub trait StoreExt<R: Runtime> {
    fn load_or_get_store(
        &self,
        path: &Path,
        auto_save: Option<Duration>,
    ) -> Result<StoreWrapper<R>>;
    fn with_store<T>(&self, path: &Path, f: impl FnOnce(&mut Store<R>) -> Result<T>) -> Result<T>;
}
```

- `load_or_get_store` loads the store (fallback to empty store if can't load) if it's not in `app.state`'s store collection already and returns a `StoreWrapper`
- `StoreWrapper` stores the path and forwards all the functions like `set`, `delete`, ... using `with_store` internally, this way you can do `let store = app.load_or_get_store(path, None);` and use the store directly
- `with_store` works the same as before, but errors out if no store's found instead of creating a new one

JavaScript API can stay the same to be honest, passing in an extra number isn't a big deal and it's managed internally anyway

To be honest, I don't really like the idea of using path as a key to the store in Rust, but I don't know if there's a better way of doing it

amrbashir: This is a good direction IMO, but I think we can:
- Remove `with_store`
- Disable creating `Store` manually, and can only be created through the trait ext:
	```rs
	pub trait StoreExt<R: Runtime> {
	  /// get or create a new store
	  fn store(&self, path: &Path) -> Result<Store<R>>;
	}
	``` 
	example:
	```rs
	app.store("main")?.insert("key", "value");
	```
- Move autosave functionality to be a method setter on `Store`

amrbashir: This will also fix issues on mobile when users directly call `Store::new` and don't initialize the mobile plugin handle.

Legend-Master: If we disable creating `Store` manually (which currently can only be created through `StoreBuilder`), we'll probably need to have a `store_builder` trait method to allow more customization

I'm thinking we should just always manage stores in our state and the user can't create a new store if that path is already in our managed stores, but I don't know how we gonna handle those mutexes to be honest, and in that case we might need a `close` function to close the store if the user don't want it to stay in the memory

Also do you think we should do it in this PR or in another one in the future?

amrbashir: I think we should just go ahead and implement the refactor since this PR feature already requires breaking changes.

amrbashir: > If we disable creating `Store` manually (which currently can only be created through `StoreBuilder`), we'll probably need to have a `store_builder` trait method to allow more customization

We can have another method to create a store builder
```rs
 pub trait StoreExt<R: Runtime> {
   /// get or create a new store
   fn store(&self, path: &Path) -> Result<Store<R>>;
   /// create a store builder
   fn store_builder(&self, path: &Path) -> Result<StoreBuilder<R>>;
 }
```

and similar to `Store`, we shouldn't allow creating `StoreBuilder` manually and prefill it with needed information like mobile plugin handle, so when user eventually call `store_builder.build()` it will behave correctly. This is similar to updater plugin `app.updater_builder()` method.

Legend-Master: > I think we should just go ahead and implement the refactor since this PR feature already requires breaking changes.

kk, need to think about the new design more then

> This is similar to updater plugin app.updater_builder() method

Sounds good to me

The problem I have right now is all about managed stores and non-managed stores, and the mutexes to be honest, unmanaged stores don't make much sense since they are not shared with the js side and also may introduce multiple open stores pointing to the same path, but with managed stores, how we gonna make the experience smooth if every step requires a mutex?

amrbashir: I would keep the performance optimizations in a later PR as I don't see a way around it for now.

Legend-Master: The best I can currently think of would probably be separating the creation and the access, and the store is always managed by our store collection, kinda find it end up being just https://github.com/tauri-apps/plugins-workspace/pull/1550#discussion_r1678687871 but replace `load_or_get_store` with `store` and `store_builder` (and maybe the store wrapper itself has a `with_store` function instead of forwarding all the store methods), also probably a function to remove the store from the managed state

amrbashir: Sure, we can have a remove store from managed state function.

Legend-Master: Actually, `store_builder` kinda doesn't make sense if it returns an existing one with different settings, this is also kinda challenging for lazy loaded stores like what we do for the js APIs

I still don't have a good idea about the new design to be honest

Legend-Master: I'm just thinking, what about the stores are in the states and are unique, but they'll only have the data, not the settings, and we can have store wrappers with all the settings like (de)serialize function and auto save, and the public facing APIs will be only store wrappers, this way, there'll be no conflicts, and if you want to re-use the settings/wrappers, save it to the resource table

- Wrapper is a `Resource` and will have a shared reference to the store
- We manage stores in state but only a weak reference, when all wrappers are dropped, we remove the store
- JS APIs will be pointing to the wrapper

This is kind of be like the store is a server and all the settings/wrappers are clients connecting to the server and request the server to do different things with their own settings, and when everyone's done, close the server

Legend-Master:
@amrbashir Made a proof of concept, take a look see if this looks good to you

We probably need a way to create named stores to allow them to be shared easier, kinda like how the tray icons work, also still need to think about how should save on exit work since we can now having multiple different store settings

Note: didn't do the mobile part yet, didn't do the docs yet

plugins/store/src/store.rs
```diff
@@ -290,14 +327,152 @@ impl<R: Runtime> Store<R> {
...
     }
 }
+
+pub struct Store<R: Runtime> {
+    defaults: Option<HashMap<String, JsonValue>>,
+    serialize: SerializeFn,
+    deserialize: DeserializeFn,
+    auto_save: Option<Duration>,
+    auto_save_debounce_sender: Arc<Mutex<Option<UnboundedSender<AutoSaveMessage>>>>,
+    store: Arc<Mutex<StoreInner<R>>>,
```
amrbashir, line 369: why do we need to use `Arc<Mutex>` here? is it for mutability? can we use RefCell instead?

Legend-Master: Needed for auto save to mutate it in another thread

plugins/store/src/store.rs
```diff
@@ -303,14 +337,152 @@ impl<R: Runtime> Store<R> {
...
+        });
+        Ok(())
+    }
+}
+
+impl<R: Runtime> Drop for Store<R> {
+    fn drop(&mut self) {
+        let auto_save_debounce_sender = self.auto_save_debounce_sender.lock().unwrap();
+        if let Some(ref sender) = *auto_save_debounce_sender {
+            let _ = sender.send(AutoSaveMessage::Cancel);
```
lucasfernog, line 485: are you sure we want to cancel on drop? shouldn't we force save here instead?

lucasfernog (approved):
we should also add a JS option to configure the debounce delay, but i'll take this

lucasfernog:
really good work btw

plugins/store/src/store.rs
Legend-Master: I forget why I did this too be honest, we probably should move this to `save` instead, not sure we really need to do anything on drop

Legend-Master:
> We probably need a way to create named stores to allow them to be shared easier, kinda like how the tray icons work, also still need to think about how should save on exit work since we can now having multiple different store settings

@lucasfernog What do you think about these? Currently we're doing nothing on exit, this is different from what we used to do, also do you think we should just prevent duplicated stores with the same path or maybe add a function to share a store with a name?

lucasfernog:
> do you think we should just prevent duplicated stores with the same path

yes I think that's reasonable

Legend-Master:
I'll need to think about this a bit more on how to do it, I will be out for a few hours, can get back on is when I'm back

HuakunShen:
Where is the file stored by config plugin?
I couldn't find it under `appDataDir()`

Legend-Master:
Not sure what do you mean by config plugin, if you mean the store plugin, you can see the path used with `resolve_store_path`, it should be under `appDataDir` if your store's path is a relative path or if it's an absolute path then it will just use that path

HuakunShen:
Found it, thanks! @Legend-Master

PR #1556, opened 2024-07-16
fix(notification/windows): Grant permission without invoking backend

### LINKED ISSUE #1254: Opening a webpage with WebviewWindow encounters an error STATUS_ACCESS_VIOLATION.
Hi everyone,

I encountered a little issue while using Tauri. I wanted to ask whether this is a bug or something specific to my setup.

After the release of Tauri v2.0.0-beta.14, I eagerly upgraded my program from beta.8 because beta.14 fixed the issue of webviews not being created. However, just when everything seemed to be going well, I found that although my program could create webviews normally, the web page would crash with the error code `STATUS_ACCESS_VIOLATION`.This didn't occur in beta.8.
crash

This issue is almost nonexistent on text web pages, but it occurs almost 100% of the time on video streaming platforms like YouTube and Twitch. Additionally, on websites with Cloudflare verification, it leads to verification failure.
verification

After testing, I found that the issue occurs with`.plugin(tauri_plugin_notification::init())`. Removing this line resolves all the problems.

Since it seems to be related to `tauri_plugin_notification`, I reported it here. Can anyone help?

Thank you very much!!

FabianLars:
fixes #1254 until we use webview2's permission apis which i'd wait for until we dropped win 7 support so we don't have to worry about 2 implementations.


tance77:
I'm excited for this fix

Issue #1347, opened 2024-05-19
tauri-plugin-sql: File based migrations


Unit2795:
This is a feature request. It would be helpful to be able to provide migrations to the SQL plugin in the form of a directory path. So I could provide SQL files like so:

```
1-add_initial_tables.sql
2-add_user_table.sql
```

PR #1561, opened 2024-07-17
feat(sql):Added File based migrations

Sir-Thom:
Added a way to do file based migrations 
I have also added in the README.md. how to do it. 
 
this solve issue #1347


FabianLars:
What is the actual intention? Should it be a built time thing or are the migrations supposed to be dynamically loaded from the user's system (requiring you to bundle the migration files with your app, for example with the resources feature).

To me the build time approach makes sense (and the issue sounds like that to me) but your implementation loads it from the system running the app.

p.s. also, is there a way to get rid of Box::leak? Not a fan of that in this context tbh.

Sir-Thom:
> What is the actual intention? Should it be a built time thing or are the migrations supposed to be dynamically loaded from the user's system (requiring you to bundle the migration files with your app, for example with the resources feature).
> 
> To me the build time approach makes sense (and the issue sounds like that to me) but your implementation loads it from the system running the app.
> 
> p.s. also, is there a way to get rid of Box::leak? Not a fan of that in this context tbh.

Yeah, it makes more sense to do it via build time. I am currently reworking on  it to create an autogenerated file  created during build time that can be used as a migration list so the user only has to set 2 env one for the project root  and one for the migration directory and the user  only has to do `. add_migrations ("sqlite: test. db", migrations())`
here what it look like currently 


https://github.com/user-attachments/assets/35ced8aa-ea4c-475c-ac81-2ca2c0521b36

Issue #1546, opened 2024-07-14
`tauri-window-save-state` causes runtime panic when using `tauri-nspanel`


Hacksore:
So I setup this repro but tl;dr if you leave the window save state plugin enabled this repro will crash when you go to access the `NSPanel`.

https://github.com/Hacksore/tauri-window-save-state-bug

I've removed everything to be super straight to the issue so follow the readme and you can reproduce.

Unsure of what the issue is but here is the short stacktrace and you can find the full one in the repro.

```
Finished `dev` profile [unoptimized + debuginfo] target(s) in 9.16s
thread 'main' panicked at /Users/hacksore/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cocoa-0.24.1/src/appkit.rs:1183:9:
Uncaught exception <NSException: 0x6000016f6940>
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
libc++abi: terminating due to uncaught foreign exception
```

This happens in dev and release modes so not sure what's going on. cc @ahkohd (creator of the [tauri-nspanel](https://github.com/ahkohd/tauri-nspanel)).

related issues:
https://github.com/overlayeddev/overlayed/issues/152
https://github.com/Hacksore/tauri-window-save-state-bug/issues/1

ahkohd:
Since the `NSWindow` has been converted to `NSPanel`, you can no longer call the `window.is_maximised` or maximise the window anymore.

> You can't maximise a panel.

https://github.com/tauri-apps/plugins-workspace/blob/a3fe84296a9a51fed960ab99ee7cebebb7df5390/plugins/window-state/src/lib.rs#L236

this line calls maximise on the window(panel), hence why it crashed.

ahkohd:
https://github.com/tauri-apps/plugins-workspace/blob/a3fe84296a9a51fed960ab99ee7cebebb7df5390/plugins/window-state/src/lib.rs#L235

this line needs to be refactored.

 such that `self.is_maximized` is only called if and only if the state flag includes `MAXIMIZED`

ahkohd:
one more thing,

@Hacksore when calling `saveWindowState`, you will have to exclude `MAXIMIZED`
<img width="1071" alt="image" src="https://github.com/user-attachments/assets/a57e5162-1a09-4140-9d05-2fe4d2e403f6">

ahkohd:
Here is the patch PR: https://github.com/tauri-apps/plugins-workspace/pull/1547

[...]

PR #1568, opened 2024-07-20
refactor: call self.is_maximized if and only if MAXIMIZE is set in the source bit flag

ahkohd:
This PR refactors `update_state`  implementation to call `window.is_maximized` if and only if `MAXIMIZED` flag set in the `StateFlags`.

Please refer to this issue for further context:
https://github.com/tauri-apps/plugins-workspace/issues/1546


ahkohd:
related: https://github.com/tauri-apps/plugins-workspace/pull/1547

lucasfernog (commented):
can you add a change file? see https://github.com/tauri-apps/plugins-workspace/pull/1497/files#diff-fa1115a2c0dad5bce425bf97def7e9365e503d80b91685a7e2e9b06ecda68365 for an example

ahkohd:
> can you add a change file? see https://github.com/tauri-apps/plugins-workspace/pull/1497/files#diff-fa1115a2c0dad5bce425bf97def7e9365e503d80b91685a7e2e9b06ecda68365 for an example

Okay will do.

ahkohd:
@lucasfernog please refer to commit  a60889f

plugins/window-state/src/lib.rs
```diff
@@ -278,7 +278,7 @@ impl<R: Runtime> WindowExtInternal for WebviewWindow<R> {
 
 impl<R: Runtime> WindowExtInternal for Window<R> {
     fn update_state(&self, state: &mut WindowState, flags: StateFlags) -> tauri::Result<()> {
-        let is_maximized = match flags.intersects(StateFlags::MAXIMIZED | StateFlags::SIZE) {
+        let is_maximized = match flags.intersects(StateFlags::MAXIMIZED) {
```
lucasfernog, line 281: after taking a deeper look unfortunately I think this impacts https://github.com/tauri-apps/plugins-workspace/blob/5ec310199f9bf9a8b314ef2c6c1f8ff62c66ad22/plugins/window-state/src/lib.rs#L310 
this was been around since forever, I think we should instead fix the problem on the tao side, maybe return false if it's not a NSWindow?

ahkohd: Hmm that's true. We can do what you suggested.

ahkohd: I will close this PR then and open a new PR on tao.

PR #1581, opened 2024-07-24
remove cargo package, publish without verification

jbolda:
## Motivation

We no longer presume any value in publishing the `.crate` to the GitHub Release.

## Approach

- Removing this, and in turn, the need to compile and check before publishing.
- I also merged the publish workflows as, with these changes, there was no need to maintain separate workflows.
- Finally, I added a step in the workflow to sync the lock file. This should show up in the Version Packages PR now.


.changes/config.json
```diff
@@ -258,17 +254,13 @@
...
-          "dryRunCommand": true
-        },
         {
           "command": "echo '<details>\n<summary><em><h4>Cargo Publish</h4></em></summary>\n\n```'",
           "dryRunCommand": true,
           "pipe": true
         },
         {
-          "command": "cargo publish --features sqlite",
+          "command": "cargo publish --features sqlite --no-verify",
```
FabianLars: I _think_ we can remove the --features flag here if we skip the build via --no-verify, right?

jbolda: oooh you might be right on that. I can try running it with a `cargo publish --dry-run` locally to confirm. Any other ideas to check it or just ship and keep an eye on it?

FabianLars: hmm no, can't think of another option than dry-run or try and see.

FabianLars: I'd also like to make the feature flags additive next week or see anyway at which point the flags wouldn't be needed either way. (I wanted to do that for weeks so don't wait for it lol)

jbolda: Appears that it does work!
```
❯ cargo publish --dry-run
[...snipped...]
Some errors have detailed explanations: E0277, E0412, E0425, E0433.
For more information about an error, try `rustc --explain E0277`.
warning: `tauri-plugin-sql` (lib) generated 3 warnings
error: could not compile `tauri-plugin-sql` (lib) due to 16 previous errors; 3 warnings emitted
warning: build failed, waiting for other jobs to finish...

❯ cargo publish --dry-run --no-verify
    Updating crates.io index
   Packaging tauri-plugin-sql v2.0.0-beta.10 (/Users/jacobbolda/dev/github/tauri-apps/plugins-workspace/plugins/sql)
    Packaged 30 files, 107.7KiB (58.5KiB compressed)
   Uploading tauri-plugin-sql v2.0.0-beta.10 (/Users/jacobbolda/dev/github/tauri-apps/plugins-workspace/plugins/sql)
warning: aborting upload due to dry run
```

FabianLars (approved):

PR #1636, opened 2024-08-08
fix(updater): linux not retaining AppImage permissions, enhance tests

lucasfernog:
- adds a test for migration from v1 to v2
- extends the existing integration test to actually verify if the app was updated
- retain AppImage permission - we need this because the permissions aren't set when downloading appimages without the zip format


.github/workflows/integration-tests.yml
```diff
@@ -41,7 +41,7 @@ jobs:
         if: matrix.platform == 'ubuntu-latest'
```
tweidinger: This should be fixed to `22.04` as it will break when github updates to the new LTS `24.04` as webkit2gtk 4.0 is no longer available in their repositories.

tweidinger (changes requested):

.github/workflows/integration-tests.yml
lucasfernog: we should do that for the entire tauri 1.x and plugin v1 branches

PR #1640, opened 2024-08-09
fix(fs): support any UTF-8 path in writeFile

SRutile:
improper input parsing

In the `writeFile` function, when path is a string that contains Chinese characters or spaces, and `options.baseDir` is not provided, it can lead to errors. This PR introduces a fix: when path is a string and contains Chinese characters or spaces, it converts such paths to URL format, thus preventing these errors. As a result, paths obtained from save dialogs can be directly used in the `writeFile` function without requiring preliminary conversion to URL format.

---

在writeFile函数中，当path为字符串且包含中文字符或空格时，如果没有提供options.baseDir，会导致错误。此PR引入了一个修复方案：当path为字符串并且包含中文字符或空格时，将其转换为URL格式，以防止这些错误的发生。这样一来，通过保存对话框获取的路径可以被直接用于writeFile函数，而不需要预先转换为URL格式。


lucasfernog:
is this a problem only for writeFile? i expected this to happen for any FS API call..
also which error do you have? i assume the path sent to the tauri side is incorrect right?

SRutile:
Only the `writeFile` function has this issue, as it uses `request.body()` to obtain the data. Normally, it should return `Raw([])`, but when there are Chinese characters or spaces, it returns `Json(Object {"0": Number()})`. To ensure it always returns data of the `Raw` type, I converted the `path` to URL format.

plugins/fs/guest-js/index.ts
```diff
@@ -1009,6 +1009,10 @@ async function writeFile(
     throw new TypeError("Must be a file URL.");
   }
 
+  if (!options || !options.baseDir) {
+    path = path instanceof URL ? path : new URL(encodeURI(path));
```
amrbashir: This will throw if provided path is something like this `./file` or just `file`, because `new URL()` expects a valid URL e.g. `<scheme>://<domain>`

SRutile: I'm aware of this issue and appreciate you pointing it out. To resolve this, I decided to abandon using `new URL()` and instead used `TextEncoder`, adding decoding logic on the Rust side.

amrbashir (approved):

xygengcn:
> improper input parsing
> 
> In the `writeFile` function, when path is a string that contains Chinese characters or spaces, and `options.baseDir` is not provided, it can lead to errors. This PR introduces a fix: when path is a string and contains Chinese characters or spaces, it converts such paths to URL format, thus preventing these errors. As a result, paths obtained from save dialogs can be directly used in the `writeFile` function without requiring preliminary conversion to URL format.
> 
> 在writeFile函数中，当path为字符串且包含中文字符或空格时，如果没有提供options.baseDir，会导致错误。此PR引入了一个修复方案：当path为字符串并且包含中文字符或空格时，将其转换为URL格式，以防止这些错误的发生。这样一来，通过保存对话框获取的路径可以被直接用于writeFile函数，而不需要预先转换为URL格式。



> improper input parsing
> 
> In the `writeFile` function, when path is a string that contains Chinese characters or spaces, and `options.baseDir` is not provided, it can lead to errors. This PR introduces a fix: when path is a string and contains Chinese characters or spaces, it converts such paths to URL format, thus preventing these errors. As a result, paths obtained from save dialogs can be directly used in the `writeFile` function without requiring preliminary conversion to URL format.
> 
> 在writeFile函数中，当path为字符串且包含中文字符或空格时，如果没有提供options.baseDir，会导致错误。此PR引入了一个修复方案：当path为字符串并且包含中文字符或空格时，将其转换为URL格式，以防止这些错误的发生。这样一来，通过保存对话框获取的路径可以被直接用于writeFile函数，而不需要预先转换为URL格式。

如果是mac的话，path没有file://协议开头，new URL会出错的，还要把path加上file

PR #1654, opened 2024-08-12
fix: docs.rs

lucasfernog:


plugins/biometric/build.rs
```diff
@@ -5,9 +5,16 @@
...
 fn main() {
-    tauri_plugin::Builder::new(COMMANDS)
+    if let Err(error) = tauri_plugin::Builder::new(COMMANDS)
         .global_api_script_path("./api-iife.js")
         .android_path("android")
         .ios_path("ios")
-        .build();
+        .try_build()
+    {
+        println!("{error:#}");
```
amrbashir: leftover?

amrbashir: also when not just ignore error when building for docsrs and otherwise just unwrap and panic.

plugins/nfc/build.rs
```diff
@@ -5,11 +5,18 @@
...
 fn main() {
-    tauri_plugin::Builder::new(COMMANDS)
+    if let Err(error) = tauri_plugin::Builder::new(COMMANDS)
         .global_api_script_path("./api-iife.js")
         .android_path("android")
         .ios_path("ios")
-        .build();
+        .try_build()
+    {
+        println!("{error:#}");
```
amrbashir: ditto

plugins/biometric/build.rs
lucasfernog: pushed a change to match this @amrbashir

amrbashir (approved):

PR #1591, opened 2024-07-27
feat(dialog, fs): Save file on android

mikoto2000:
This is part of #1494. (Implementation of save dialog and save text file for Android)

I implemented the save dialog for the dialog plugin and save_text_file for the fs plugin.

As mentioned in the following two comments, saving files on Android requires the use of `activity.getContentResolver().openOutputStream(uri)`, so I use Uri instead of file path in the communication between the app and the plugin.

Also, since the fs plugin requires the use of `activity.getContentResolver().openOutputStream(uri)`, I've added an Android implementation to the fs plugin.

- https://github.com/tauri-apps/plugins-workspace/issues/1494#issuecomment-2248352810
- https://github.com/tauri-apps/plugins-workspace/issues/1494#issuecomment-2250623258

tauri-plugin-dialog-android_002

Issue #1494, opened 2024-06-24
File save dialog is not implemented on mobile


songjiachao:
![image](https://github.com/tauri-apps/plugins-workspace/assets/70792885/02f4eca7-c44c-425e-a8a9-76aa5a10b367)

mikoto2000:
@songjiachao @FabianLars 
In my application, I implemented the Android save dialog like this:

https://github.com/mikoto2000/OASIZ_TimeLogger2/commit/aea157f5abab98995a84455a3014b5620be60568

Are there any good ideas for applying this to the save method of the dialog plugin?

FabianLars:
In the best case, you just have to duplicate showFilePicker and modify it to your code. Since the rest of the android implementation is a bit buggy though i don't know how good that'll work.

mikoto2000:
@FabianLars 
Thank you for your comment.
I didn't realize that `open` was partially supported with Android.
It looks like I can create `save` by imitating the code below.
I'll give it a try.

- https://github.com/tauri-apps/plugins-workspace/blob/20a1d24ee004e77c2d12a0e20d258ce120216ed1/plugins/dialog/android/src/main/java/DialogPlugin.kt#L49
- https://github.com/tauri-apps/plugins-workspace/blob/20a1d24ee004e77c2d12a0e20d258ce120216ed1/plugins/dialog/src/mobile.rs#L77
- https://github.com/tauri-apps/plugins-workspace/blob/v2/plugins/dialog/src/lib.rs#L406
- https://github.com/tauri-apps/plugins-workspace/blob/v2/plugins/dialog/src/lib.rs#L517
- https://github.com/tauri-apps/plugins-workspace/blob/v2/plugins/dialog/src/commands.rs#L172

mikoto2000:
![tauri-plugin-dialog-android_001](https://github.com/user-attachments/assets/50cebc82-0a1b-468b-9a06-7b0a420185ce)
I implemented the part where the save dialog opens and the path is returned, but it seems that in Android, the file is created as soon as the file name is specified in the dialog...

Next, I need to create an Android version of the fs plugin's writeFile, otherwise this function itself will be meaningless.

Since it is necessary to obtain the OutputStream of the file entity from the URI using the code `activity.getContentResolver().openOutputStream(uri)`, it is not possible to create a file in the Rust world.

Huh? Isn't there already a function to open a file on an Android device? The implementation of that seems to be helpful.

mikoto2000:
[...]

PR #1658, opened 2024-08-13
feat(fs): resolve content URIs on Android

mikoto2000:
This is part of #1494. (Implementation of write text file for Android)
Separated #1591 for each plugin change.

NOTE: Please this PR merge after #1657

I implemented the save_text_file for the fs plugin.

As mentioned in the following two comments, saving files on Android requires the use of `activity.getContentResolver().openOutputStream(uri)`, so I use Uri instead of file path in the communication between the app and the plugin.

Since the fs plugin requires the use of `activity.getContentResolver().openOutputStream(uri)`, I've added an Android implementation to the fs plugin.

- https://github.com/tauri-apps/plugins-workspace/issues/1494#issuecomment-2248352810
- https://github.com/tauri-apps/plugins-workspace/issues/1494#issuecomment-2250623258


mikoto2000:
Sorry.
I will fix check failings.

lucasfernog:
I've pushed a change to extend the approach for more file APIs (all the read, write, open APIs). We should add dedicated Android APIs for the other functions (reading a directory, renaming files, copying, etc).

lucasfernog:
This is exciting!!

lucasfernog (approved):

PR #1662, opened 2024-08-14
fix(http): properly invoke abort

twlite:
The abort signal does not work as abort is not invoked. This pr fixes that issue and invokes abort when abort signal aborts.


plugins/http/guest-js/index.ts
```diff
@@ -186,7 +186,7 @@ export async function fetch(
     throw new Error(ERROR_REQUEST_CANCELLED);
   }
 
-  signal?.addEventListener("abort", () => abort);
+  signal?.addEventListener("abort", () => void abort());
```
amrbashir, line 189: could you explain what the issue is and how does adding void fixes it?

twlite: It isn't adding void, the original code did not invoke abort when abort signal had aborted. I added void because I saw eslint comment above about unused promise.

twlite: The original code had
`signal?.addEventListener("abort", () => abort);` where abort function does not get invoked thus leading to no-op behavior. This was probably misinterpreted as `signal?.addEventListener("abort", abort);`, which does invoke the abort function.

twlite: @amrbashir I have applied the shorthand syntax, removing void.

amrbashir: @twlite thanks for explaining, I didn't notice missing `()`

amrbashir (dismissed):

twlite:
😄 eslint is not happy about me removing void keyword haha

amrbashir (commented):
could you also add a change file in `.changes` directory?

twlite:
~~@amrbashir hi, I have added it to .changes dir but im not sure if it is okay or not.~~

I guess `http-js` it is.

amrbashir (dismissed):

amrbashir:
thank you

amrbashir (approved):

PR #1668, opened 2024-08-16
Sharesheet plugin

mattyg:
Plugin that provides basic access to the Android Sharesheet and iOS Share Pane. Currently only supports sharing text.

I'm waiting to get access to a Mac mid next week to review / bug fix the iOS implementation. The Android implementation works well.

Please let me know if this is a PR you might accept and if you'd want to see any changes first.


mattyg:
After discussing on discord it sounds like the tauri team does not want to take on maintenance of this plugin until tauri v2 is stabilized. Thus I published it standalone here: https://github.com/buildyourwebapp/tauri-plugin-sharesheet

PR #1696, opened 2024-08-27
feat(fs): `scope-*-recursive` permission allows read dir, closes #1307

### LINKED ISSUE #1307: `fs:scope-resource-recursive` won't allow access to the resource directory
Taking the example from https://beta.tauri.app/references/acl/#capabilities

```
{
  "permissions": [
    {
      "identifier": "fs:scope",
      "allow": [{ "path": "$APPDATA" }, { "path": "$APPDATA/**" }]
    }
  ]
}
```

This example doesn't really work as one would expect, making all content of the appdata directory accessible. In order to scope all content of a directory, the scope `<directory>/**` is not enough. `<directory>/` is needed as well. Since `$APPDATA` itself is not a member of `$APPDATA`, this makes sense, I guess.

As a result, with the current definition of `scope-resource-recursive` (and similar scopes), after adding a permission such as `allow-resource-read-recursive`, one still won't be able to access the directory, since its path only includes the directory's content, but not the directory itself:

```
path = "$RESOURCE/**"
```

Permissions such as `allow-resource-meta` are required to be used together to get access to the directory. They include the path that points to the diretory itself.

When it comes to `allow-resource-read-recursive` (and similar scopes), I don't think it makes sense to be not able to access the directory itself when this scope is used. The [documentation](https://github.com/tauri-apps/tauri-plugin-fs/blob/81ce6c141ec113a581ed55410000fe05ca101421/permissions/autogenerated/reference.md) of `allow-resource-read-recursive` says

> This allows full recursive read access to the complete `$RESOURCE` folder, files and subdirectories.

It didn't come to me intuitively that I don't have the permission to access the directory itself when I have "full" recursive read access to the "complete" directory.

lucasfernog:
We could instead document that you should also enable `scope-*-index`, it is more a matter of semantics.

### DIFF
--- a/.changes/fs-scope-recursive-allow-read-dir.md
+++ b/.changes/fs-scope-recursive-allow-read-dir.md
@@ -1,0 +1,5 @@
+---
+"fs": patch:enhance
+---
+
+The `scope-*-recursive` permissions now also allow reading the contents of the directory.
--- a/plugins/fs/build.rs
+++ b/plugins/fs/build.rs
@@ -112,12 +112,16 @@ identifier = "scope-{lower}-recursive"
 description = "This scope permits recursive access to the complete `${upper}` folder, including sub directories and files."
 
 [[permission.scope.allow]]
+path = "${upper}"
+[[permission.scope.allow]]
 path = "${upper}/**"
 
 [[permission]]
 identifier = "scope-{lower}"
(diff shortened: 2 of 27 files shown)


amrbashir (approved):

tweidinger:
:( had this done locally and you beat me. Maybe next time 🤷

PR #1702, opened 2024-08-29
fix(window-state): saving a minimized window's state changes its position to -32000

Legend-Master:
This happens on windows, but I'm not sure if this is also the case for mac and linux


FabianLars:
> This happens on windows, but I'm not sure if this is also the case for mac and linux

iirc this was either windows only or mac/linux used less extreme values 🤔

FabianLars (approved):

PR #1703, opened 2024-08-29
fix(http): remove unexpectedly changing origin header when unsafe-headers feature enabled

PeraSite:
When you use the unsafe-headers feature flag, the website displays CORS-related errors, because when you turn on the unsafe-headers feature flag, tauri automatically set the Origin header to localhost.
This is very confusing, and it is impossible to debug unless you look at the source code.
It will have to be up to developer to set up Origin header value not tauri.
And it doesn't even fit the name "unsafe-headers".

In the first place, why tauri filter the headers by the unsafe-headers feature?
When I first used plugin-http, I had a hard time because a specific header was not recognized by tauri.
May I know why this feature was introduced?
If it's a security issue, it should be fully explained in documentation.

### DIFF
--- a/plugins/http/src/commands.rs
+++ b/plugins/http/src/commands.rs
@@ -254,15 +254,6 @@ pub async fn fetch<R: Runtime>(
                     request = request.header(header::USER_AGENT, HTTP_USER_AGENT);
                 }
 
-                if cfg!(feature = "unsafe-headers")
-                    && !headers.contains_key(header::ORIGIN.as_str())
-                {
-                    if let Ok(url) = webview.url() {
-                        request =
-                            request.header(header::ORIGIN, url.origin().ascii_serialization());
-                    }
-                }
-
                 if let Some(data) = data {
                     request = request.body(data);
                 }


amrbashir:
The origin is automatically set, as requested in https://github.com/tauri-apps/plugins-workspace/issues/1167, you can override this behavior by setting the origin using `unsafe-headers` feature flag or we can add a new option to disable this behavior.

PeraSite:
I think it's very confusing to make "automatically setting Origin header" behaviour be able to turn off by turning on another feature flag.
It's just different between possible to change Origin headers and forces default Origin value to some other value.

For convenience, I wish that code gets removed.

amrbashir:
I don't mind having an option to disable setting the `origin` header, and doesn't need to be a feature flag, it could be an option set on each `fetch` request

PeraSite:
yeah adding option to each fetch call would be better because dev doesn't need to read external document for that.

amrbashir:
@PeraSite would you like to implement the option in this PR or would you like to close this and create an issue instead of me or someone else can work on it?

PeraSite:
> @PeraSite would you like to implement the option in this PR or would you like to close this and create an issue instead of me or someone else can work on it?

Let me close this PR first.
And after that, I'll post a new PR to edit `ClientOptions` type to have unsafe-header boolean.
Also remove `#[cfg(not(feature = "unsafe-headers"))]` and `is_unsafe_header` part from the Rust codebases.
Adding fetch parameter should give better experience than Rust's feaature flag.

PR #1707, opened 2024-08-31
feat(dialog): implement save API on iOS

lucasfernog:


plugins/dialog/ios/Sources/DialogPlugin.swift
```diff
@@ -112,6 +128,40 @@ class DialogPlugin: Plugin {
...
 
+  @objc public func saveFileDialog(_ invoke: Invoke) throws {
+    let args = try invoke.parseArgs(SaveFileDialogOptions.self)
+
+    // we must pretend that we're moving a file and the user must select its location
+    let fileManager = FileManager.default
+    let destinationFolder = fileManager.urls(for: .documentDirectory, in: .userDomainMask).first!
+    let destinationPath = destinationFolder.appendingPathComponent(args.fileName ?? "file")
+    if !fileManager.fileExists(atPath: destinationPath.path) {
+      // the file contents must be actually provided by the tauri dev after the path is resolved by this API
```
amrbashir: not sure I quite understand this comment or what the logic is trying to do, this comment needs to be improved

amrbashir (changes requested):

plugins/dialog/ios/Sources/DialogPlugin.swift
lucasfernog: done, hope the implementation makes more sense now
this is due to the Apple API design which does not match other operating systems

amrbashir (approved):

PR #1710, opened 2024-09-02
fix(fs): can't use Windows path

### LINKED ISSUE #1704: [bug][fs][v2] unable to get path from XX
When I want to read a text file such as `C://users/user/temp.json`, `fs.readTextFile` is work.

But it reports the `unable to get path from c://users/user/temp.json` error recently.

+ `tauri-plugin-fs`: v2
+ `@tauri-apps/plugin-fs`: 2.0.0-rc.1

Legend-Master:
Url deserializes Windows drive letter to url scheme which is not what we wanted, since we don't use single letter schemes anyways, we can just assume a single letter scheme url to be a path not a url

Fix #1704

### DIFF
--- a/plugins/fs/src/commands.rs
+++ b/plugins/fs/src/commands.rs
@@ -1202,10 +1202,10 @@ fn get_stat(metadata: std::fs::Metadata) -> FileInfo {
 }
 
 mod test {
-    use super::*;
-
     #[test]
     fn safe_file_path_parse() {
+        use super::SafeFilePath;
+
         assert!(matches!(
             serde_json::from_str::<SafeFilePath>("\"C:/Users\""),
             Ok(SafeFilePath::Path(_))
--- a/plugins/fs/src/commands.rs
+++ b/plugins/fs/src/commands.rs
@@ -1205,24 +1205,14 @@ mod test {
     use super::*;
 
     #[test]
-    fn windows_path() {
-        assert_eq!(
-            matches!(
-                serde_json::from_str::<SafeFilePath>("\"C:/Users\""),
-                Ok(SafeFilePath::Path(_))
-            ),
-            true
-        );
-    }
-
-    #[test]
-    fn url() {
-        assert_eq!(
-            matches!(
-                serde_json::from_str::<SafeFilePath>("\"file:///C:/Users\""),
-                Ok(SafeFilePath::Url(_))
-            ),
-            true
-        );
+    fn safe_file_path_parse() {
+        assert!(matches!(
+            serde_json::from_str::<SafeFilePath>("\"C:/Users\""),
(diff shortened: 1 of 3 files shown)


plugins/fs/src/commands.rs
```diff
@@ -26,10 +26,26 @@ use crate::{scope::Entry, Error, FilePath, FsExt};
...
+{
+    let url = url::Url::deserialize(deserializer)?;
+    let scheme = url.scheme();
+    if scheme.len() != 1 {
+        Ok(url)
+    } else {
+        Err(serde::de::Error::custom(format!(
+            "Single letter scheme \"{scheme}\" is not supported because it conflicts with Windows paths"
+        )))
+    }
```
amrbashir: instead of an error here we should fallback to using a `SafePathBuf`, so instead we should implement `Deserialize` for `SafeFilePath` manually, not just one variant.

amrbashir (changes requested):

plugins/fs/src/commands.rs
```diff
@@ -23,13 +23,43 @@ use std::{
...
 
-#[derive(Debug, serde::Deserialize)]
-#[serde(untagged)]
+#[derive(Debug)]
 pub enum SafeFilePath {
     Url(url::Url),
     Path(SafePathBuf),
 }
 
+impl<'de> serde::Deserialize<'de> for SafeFilePath {
```
amrbashir, line 33: I see that we have a duplicated enum definition with and now same deserialize implementation, maybe we can combine this logic into one, by using generics? and still export aliases without a generic:
```rs 
enum FilePathOrUrl<T: FromStr> {
    Url(url::Url),
    Path(T),
}
pub type SafeFilePath = FilePath<SafeFilePath>;
pub type FilePath = FilePath<PathBuf>;
```

amrbashir (changes requested):

plugins/fs/src/commands.rs
Legend-Master: > ```rust
> enum FilePathOrUrl<T: FromStr> {
>     Url(url::Url),
>     Path(T),
> }
> pub type SafeFilePath = FilePath<SafeFilePath>;
> pub type FilePath = FilePath<PathBuf>;
> ```

I'm not sure how that'll work, `SafeFilePath` doesn't implement `FromStr`, and that means we either need to do it in tauri or we need a wrapper

amrbashir: The best is to implement `FromStr` on in tauri. For now, we can just add a TODO then or merge as is.

amrbashir:
Also I think we should add some tests in this PR, just to ensure we don't regress again

amrbashir (approved):

PR #1723, opened 2024-09-02
Remove authenticator plugin

FabianLars:
This was supposed to be a readme update but escalated a bit. I'd love to see it back after some time (hopefully somehow being closer to webauthn) but i don't think we can accomplish that anytime soon.


amrbashir:
cc @lucasfernog

lucasfernog (approved):
i only used this one once in the last couple years..

PR #1743, opened 2024-09-06
fix(window-state)!: saving minimized state

Legend-Master:
#1702 didn't actually work, we save the state on move event, and it fires on minimize so...

Also updated the API example to include window state plugin


Legend-Master:
I wonder if we need to update the size on resize event, since we can't get it if we're existing from minimized state, also I'm seeing that we have `prev_x` and `prev_y` to track the pre-maximized state, we might as well just add `prev_width` and `prev_height`, also in that case, is it a good idea not to track the minimized state in move/resize event and just save the minimize state and use the previous value on restore instead?

Legend-Master:
Went with the basic approach, just checking the maximize and minimize state on resize event

plugins/window-state/src/lib.rs
```diff
@@ -278,10 +274,11 @@ impl<R: Runtime> WindowExtInternal for WebviewWindow<R> {
...
     fn update_state(&self, state: &mut WindowState, flags: StateFlags) -> tauri::Result<()> {
-        let is_maximized = match flags.intersects(StateFlags::MAXIMIZED | StateFlags::SIZE) {
-            true => self.is_maximized()?,
-            false => false,
-        };
+        let is_maximized = flags
+            .intersects(StateFlags::MAXIMIZED | StateFlags::POSITION | StateFlags::SIZE)
+            && self.is_maximized()?;
+        let is_minimized =
+            flags.intersects(StateFlags::POSITION | StateFlags::SIZE) && self.is_minimized()?;
```
amrbashir, line 281: why are we checking for exitence of `Postiion` and `Size` flags here? and later check them again. doesn't make sense, it should be checked later only

.changes/window-state-physical-size.md
```diff
@@ -0,0 +1,5 @@
+---
+"window-state": patch:breaking
+---
+
+Window's size is now stored in physical size instead of logical size
```
amrbashir, line 5: is there a reason why this was changed? 

logical size seems to be a better fit here, for example, if you have 2 monitors, one at 100% DPI, the other is relatively smaller monitor in size with 125% DPI, both have same resolution, just different DPI and physical size. If you close the screen on the primary monitor (the small one), then disconnect that monitor, the big monitor becomes the primary monitor and the app should open relatively at the same position as before.

logical vs physical is always a pain to reason about, and I am not sure If I am right or wrong tbh, but at least the described example should be working like that

plugins/window-state/src/lib.rs
Legend-Master: Because we use this info on those places only and this is so we can re-use the `is_maximized` and `is_minimized` result

amrbashir: I am only talking about flags `flags.intersects(StateFlags::POSITION | StateFlags::SIZE)`

amrbashir: ah nvm, I am dumb

.changes/window-state-physical-size.md
Legend-Master: We store physical position, and this is to align with it, and physical size makes more sense to me, if I open a window that occupies 75% of the screen on 100% DPI, and later on I changed the DPI to 125%, I would like it to still occupy 75% of the screen (this can be debated either ways)

amrbashir: isn't this what logical is supposed to do? i.e. if your window size is 75% of the monitor logical size, it should stay at 75% on a different monitor with the same resolution even if it was different DPI?

Legend-Master: say we have a 1080 height screen, first we store it on 100% DPI, which is 1080 * 0.75 = 810, and then on the 125% screen, it will be converted to 810 * 1.25 = 1012.5

amrbashir: 3 years and I still confuse these two. Thanks for explaining

amrbashir (approved):

PR #1759, opened 2024-09-10
feat(deep-link): implement getCurrent on Windows/Linux

lucasfernog:
checks std::env::args() on initialization, also includes integration with the single-instance plugin

### DIFF
--- a/plugins/deep-link/examples/app/src-tauri/src/lib.rs
+++ b/plugins/deep-link/examples/app/src-tauri/src/lib.rs
@@ -30,22 +30,7 @@ pub fn run() {
             {
                 use tauri_plugin_deep_link::DeepLinkExt;
 
-                let schemes = app
-                    .config()
-                    .plugins
-                    .0
-                    .get("deep-link")
-                    .unwrap()
-                    .get("desktop")
-                    .unwrap()
-                    .get("schemes")
-                    .unwrap()
-                    .as_array()
-                    .unwrap();
-
-                for scheme in schemes {
-                    app.deep_link().register(scheme.as_str().unwrap())?;
-                }
+                app.deep_link().register_all()?;
             }
 
             app.listen("deep-link://new-url", |url| {
--- a/plugins/deep-link/examples/app/src-tauri/src/lib.rs
+++ b/plugins/deep-link/examples/app/src-tauri/src/lib.rs
@@ -13,6 +13,9 @@ fn greet(name: &str) -> String {
 #[cfg_attr(mobile, tauri::mobile_entry_point)]
 pub fn run() {
     tauri::Builder::default()
+        .plugin(tauri_plugin_single_instance::init(|_app, argv, _cwd| {
+            println!("single instance triggered: {argv:?}");
+        }))
         .plugin(tauri_plugin_deep_link::init())
         .plugin(
             tauri_plugin_log::Builder::default()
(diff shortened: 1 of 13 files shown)


plugins/deep-link/examples/app/src-tauri/src/lib.rs
```diff
@@ -13,13 +13,41 @@ fn greet(name: &str) -> String {
...
+                    .plugins
+                    .0
+                    .get("deep-link")
+                    .unwrap()
+                    .get("desktop")
+                    .unwrap()
+                    .get("schemes")
+                    .unwrap()
+                    .as_array()
+                    .unwrap();
```
amrbashir: should we include a helper method on `DeepLinkExt` ? maybe the whole logic actually `app.deep_link().register_desktop()`

plugins/deep-link/src/lib.rs
```diff
@@ -154,23 +166,55 @@ mod imp {
...
         pub(crate) app: AppHandle<R>,
-        #[allow(dead_code)]
         pub(crate) current: Mutex<Option<Vec<url::Url>>>,
+        pub(crate) config: Option<crate::config::Config>,
     }
 
     impl<R: Runtime> DeepLink<R> {
+        #[cfg(desktop)]
+        #[doc(hidden)]
+        pub fn handle_cli_arguments<S: AsRef<str>, I: Iterator<Item = S>>(&self, mut args: I) {
```
amrbashir: if this is used by tauri-plugin-single-instance, maybe we should document it instead

lucasfernog: should only be used internally tbh.. but i agree we should at least comment what it does, on it

amrbashir: it is internal to our workspace yes, but if it is used in another plugin, it should be documented and potentiolly allow other users to add same integration

plugins/deep-link/examples/app/src-tauri/src/lib.rs
lucasfernog: excellent idea i love you, pushed

plugins/deep-link/src/lib.rs
```diff
@@ -154,23 +172,87 @@ mod imp {
...
         pub fn get_current(&self) -> crate::Result<Option<Vec<url::Url>>> {
-            #[cfg(not(any(windows, target_os = "linux")))]
             return Ok(self.current.lock().unwrap().clone());
-            #[cfg(any(windows, target_os = "linux"))]
-            Err(crate::Error::UnsupportedPlatform)
+        }
+
+        /// Registers all schemes defined in the configuration file.
+        ///
+        /// This is useful to ensure the schemes are registered even if the user unregistered it
```
amrbashir: This feels wrong tbh and we shouldn't recommend that at all. If a user unregistered a protocol they should be in control of that and we shouldn't keep annoying them.

amrbashir (dismissed):

plugins/deep-link/src/lib.rs
amrbashir: I know that app images and portable builds on Windows need to manually register that but we either warn against this or provide some checks that user chose to disable this protocol, so you shouldn't re-register them

lucasfernog: oops oops oops :joy:

amrbashir (approved):

PR #1766, opened 2024-09-12
chore(single-instance): put deep link integration behined a feature

### LINKED ISSUE #1017: [single-instance] doesn't shutdown immediately on Windows

Legend-Master:
Also added some docs on the order of registering these 2 plugins

### DIFF
--- a/plugins/deep-link/README.md
+++ b/plugins/deep-link/README.md
@@ -145,7 +145,7 @@ await onOpenUrl((urls) => {
 })
 ```
 
-Note that the Plugin will only emit events on macOS, iOS and Android. On Windows and Linux the OS will spawn a new instance of your app with the URL as a CLI argument. If you want your app to behave on Windows & Linux similar to the other platforms you can use the [single-instance](../single-instance/) plugin with deep-link feature.
+Note that the Plugin will only emit events on macOS, iOS and Android. On Windows and Linux the OS will spawn a new instance of your app with the URL as a CLI argument. If you want your app to behave on Windows & Linux similar to the other platforms you can use the [single-instance](../single-instance/) plugin with the `deep-link` feature enabled.
 
 ## Contributing
 
--- a/plugins/deep-link/README.md
+++ b/plugins/deep-link/README.md
@@ -145,7 +145,7 @@ await onOpenUrl((urls) => {
 })
 ```
 
-Note that the Plugin will only emit events on macOS, iOS and Android. On Windows and Linux the OS will spawn a new instance of your app with the URL as a CLI argument. If you want your app to behave on Windows & Linux similar to the other platforms you can use the [single-instance](../single-instance/) plugin with deep-link feature and make sure to register this plugin before it.
+Note that the Plugin will only emit events on macOS, iOS and Android. On Windows and Linux the OS will spawn a new instance of your app with the URL as a CLI argument. If you want your app to behave on Windows & Linux similar to the other platforms you can use the [single-instance](../single-instance/) plugin with deep-link feature.
 
 ## Contributing
 
--- a/plugins/deep-link/README.md
+++ b/plugins/deep-link/README.md
@@ -145,7 +145,7 @@ await onOpenUrl((urls) => {
 })
 ```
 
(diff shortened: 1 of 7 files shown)


plugins/deep-link/README.md
```diff
@@ -145,7 +145,7 @@ await onOpenUrl((urls) => {
 })
 ```
 
-Note that the Plugin will only emit events on macOS, iOS and Android. On Windows and Linux the OS will spawn a new instance of your app with the URL as a CLI argument. If you want your app to behave on Windows & Linux similar to the other platforms you can use the [single-instance](../single-instance/) plugin.
+Note that the Plugin will only emit events on macOS, iOS and Android. On Windows and Linux the OS will spawn a new instance of your app with the URL as a CLI argument. If you want your app to behave on Windows & Linux similar to the other platforms you can use the [single-instance](../single-instance/) plugin with deep-link feature.
```
FabianLars:
```suggestion
Note that the Plugin will only emit events on macOS, iOS and Android. On Windows and Linux the OS will spawn a new instance of your app with the URL as a CLI argument. If you want your app to behave on Windows & Linux similar to the other platforms you can use the [single-instance](../single-instance/) plugin with the `deep-link` feature enabled.
```

plugins/single-instance/README.md
```diff
@@ -49,6 +49,8 @@ fn main() {
 }
 ```
 
+Note that currently, plugins run in the order they were added in to the builder, you need to register this plugin first if you don't want the other plugins to run on the new instance
```
FabianLars:
```suggestion
Note that currently, plugins run in the order they were added in to the builder, you need to register this plugin first if you don't want the other plugins to run on the temporary instances.
```

really not sure about this one, "new" sounds wrong to me. Temporary isn't much better though, maybe secondary? (secondary is what i use in discussions on discord at least but idk if it's good)

Legend-Master: what about `non-main instances`? they all sound weird if there's no context explaining how this thing actually works to be honest

FabianLars: We can also leave out the explanation (for now?) and just do
```
Note that currently, plugins run in the order they were added in to the builder, so make sure that this plugin is registered first.
```

FabianLars (approved):

plugins/single-instance/src/lib.rs
```diff
@@ -35,7 +34,8 @@ pub fn init<R: Runtime, F: FnMut(&AppHandle<R>, Vec<String>, String) + Send + Sy
     mut f: F,
 ) -> TauriPlugin<R> {
     platform_impl::init(Box::new(move |app, args, cwd| {
-        if let Some(deep_link) = app.try_state::<DeepLink<R>>() {
+        #[cfg(feature = "deep-link")]
+        if let Some(deep_link) = app.try_state::<tauri_plugin_deep_link::DeepLink<R>>() {
```
lucasfernog, line 38: actually for this to work I believe the deep link plugin must be registered first - so the DeepLink state has been initialized

lucasfernog: though maybe that's not the case?? i remember it worked on linux idk

Legend-Master: This is only called when the second instance notified the first one, and it's run on the first instance which should have initialized all the plugins at this point, not sure if it's going to cause problem if we spawned the second instances too fast though

PR #1770, opened 2024-09-12
fix(deep-link): emit new-url event on app load

lucasfernog:
matching the behavior on all desktop platforms

### DIFF
--- a/plugins/deep-link/src/lib.rs
+++ b/plugins/deep-link/src/lib.rs
@@ -70,17 +70,14 @@ fn init_deep_link<R: Runtime>(
     #[cfg(desktop)]
     {
         let args = std::env::args();
-        let current = if let Some(config) = api.config() {
-            imp::deep_link_from_args(config, args)
-        } else {
-            None
-        };
-
-        Ok(DeepLink {
+        let deep_link = DeepLink {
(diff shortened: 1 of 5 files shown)


plugins/deep-link/src/lib.rs
```diff
@@ -216,17 +188,31 @@ mod imp {
...
+                }
+
+                if let Some(url) = arg.and_then(|arg| arg.as_ref().parse::<url::Url>().ok()) {
+                    if config.desktop.contains_scheme(&url.scheme().to_string()) {
+                        let mut current = self.current.lock().unwrap();
+                        current.replace(vec![url.clone()]);
+                        let _ = self.app.emit("deep-link://new-url", vec![url]);
+                    } else if cfg!(debug_assertions) {
+                        log::warn!("argument {url} does not match any configured deep link scheme; skipping it");
+                    }
```
amrbashir, line 214: do we really need to warn! here, shouldn't we just skip it?

lucasfernog: i think it's still important in case the dev is using runtime schemes and isn't aware of why it isn't being picked up for missing a note in the docs

amrbashir (approved):

PR #1773, opened 2024-09-13
refactor(geolocation): simplify API, defer permission checks

### LINKED ISSUE #1631: Make explicit permission api use required in geolocation plugin
Currently it automatically requests permissions but after talking to lucas our approach in the plugins is to simply fail on missing permissions and let the dev use the permission apis to handle it. This is already the case for other plugins like the barcode scanner.

lucasfernog:


lucasfernog:
Closes #1631

amrbashir (dismissed):

examples/api/src-tauri/gen/schemas/mobile-schema.json
amrbashir, line 1: seems like we have a misconfigured .gitignore, this file and directory shouldn't be commited

amrbashir (changes requested):

amrbashir (approved):

FabianLars (approved):

PR #1774, opened 2024-09-13
fix(fs): app scopes not properly enabling paths

lucasfernog:


plugins/fs/permissions/app.toml
```diff
@@ -1,82 +1,114 @@
...
 [[permission.scope.allow]]
-path = "$APP"
+path = "$APPLOG"
 
 # Sets Section
 # This section combines the scope elements with enablement of commands
 
 [[set]]
 identifier = "allow-app-read-recursive"
 description = "This allows full recursive read access to the complete `$APP` folder, files and subdirectories."
```
Legend-Master: We need to update those descriptions as well

Legend-Master (commented):
Maybe also change `create-app-specific-dirs` and `read-app-specific-dirs-recursive` to use `scope-app-*` as well?

plugins/fs/permissions/app.toml
lucasfernog: nice catch

Legend-Master (approved):

PR #1783, opened 2024-09-14
[upload - http-extra] fix: download content to file when unsuccessful response (and test) #1750

### LINKED ISSUE #1750: tauri plugin upload doesn't check if an url leads to a 404 when using download()
Instead, the 404 response (e.g. "Not found" or whatever the 404 message is) is downloaded to disk. I don't think this is what should happen. Rather, an error should be thrown there was a 404.

fxsalazar:
fix #1750

### DIFF
--- a/plugins/upload/Cargo.toml
+++ b/plugins/upload/Cargo.toml
@@ -22,7 +22,7 @@ serde_json = { workspace = true }
 tauri = { workspace = true }
 log = { workspace = true }
 thiserror = { workspace = true }
-tokio = { version = "1", features = ["fs", "macros"] }
+tokio = { version = "1", features = ["fs"] }
 tokio-util = { version = "0.7", features = ["codec"] }
 reqwest = { version = "0.12", default-features = false, features = [
   "json",
--- a/plugins/upload/Cargo.toml
+++ b/plugins/upload/Cargo.toml
@@ -38,3 +38,4 @@ rustls-tls = ["reqwest/rustls-tls"]
 
 [dev-dependencies]
 mockito = "1.5.0"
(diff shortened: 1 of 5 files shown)


plugins/upload/Cargo.toml
```diff
@@ -22,7 +22,7 @@ serde_json = { workspace = true }
 tauri = { workspace = true }
 log = { workspace = true }
 thiserror = { workspace = true }
-tokio = { version = "1", features = ["fs"] }
+tokio = { version = "1", features = ["fs", "macros"] }
```
FabianLars:
```suggestion
tokio = { version = "1", features = ["fs"] }
```
see next suggestion

plugins/upload/Cargo.toml
```diff
@@ -35,3 +35,6 @@ read-progress-stream = "1.0.0"
 native-tls = ["reqwest/native-tls"]
 native-tls-vendored = ["reqwest/native-tls-vendored"]
 rustls-tls = ["reqwest/rustls-tls"]
+
+[dev-dependencies]
+mockito = "1.5.0"
```
FabianLars, line 40:
```suggestion
[dev-dependencies]
mockito = "1.5.0"
tokio = { version = "*", features = ["macros"] }
```

FabianLars (commented):
Thank you! Just a tiny code suggestion.

Also, we'd need a changefile (see .changes dir for what it should look like) before we can merge this :)

fxsalazar:
@FabianLars done

lucasfernog (dismissed):

lucasfernog (approved):

Issue #1804, opened 2024-09-18
[bug] Application path for cmd in Windows registry needs to be wrapped in quotes


co-p:
On Windows 10/11, when a protocol/scheme is registered, the application path for the cmd is not wrapped in quotes, which fails if the application path contains spaces.

Should be an easy fix in line 281 of .../plugins/deep-link/src/lib.rs

change: `cmd_reg.set_string("", &format!("{} \"%1\"", &exe))?;`
to: `cmd_reg.set_string("", &format!("\"{}\" \"%1\"", &exe))?;`

PR #1806, opened 2024-09-18
Added quotes for cmd

TheRobustRoast:
fix for https://github.com/tauri-apps/plugins-workspace/issues/1804

### DIFF
--- a/.changes/deep-link-space-in-path.md
+++ b/.changes/deep-link-space-in-path.md
@@ -1,0 +1,5 @@
+---
+"deep-link": patch
+---
+
+Fix fails to start when having spaces in the main binary path on Windows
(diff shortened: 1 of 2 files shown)


Legend-Master (dismissed):

Legend-Master (approved):

Legend-Master:
Thanks

FabianLars:
@Legend-Master did you check whether is_registered still works?

Legend-Master:
Oops, didn't know we actually check the registry value instead of checking if it exists, I'll make PR to fix that

FabianLars:
yeah that's to try to not return true when another app handles that same scheme (though that may need an explicit return type as simply overwriting would be a bit messed up)

PR #1813, opened 2024-09-19
refactor!(dialog): dialog builders

Legend-Master:


Legend-Master:
@FabianLars what do you think about this? I feel like this is not doing much good to be honest 😂 Maybe I'm not on the right track

FabianLars:
hmmm, hard to say without playing around with it but i think i'd not split up the builder struct for desktop and mobile and maybe also revert the function split up and just keep the title setting to where it was (or inline the rfd builder set into the function with a cfg flag instead of the whole method), not as optimized but hopefully cleaner.

Also on first glance i don't see why the buttons are now in 2 options

Legend-Master:
> hmmm, hard to say without playing around with it but i think i'd not split up the builder struct for desktop and mobile and maybe also revert the function split up and just keep the title setting to where it was

If we don't split the struct then do we just add a field like what we did for `parent` before?

> (or inline the rfd builder set into the function with a cfg flag instead of the whole method), not as optimized but hopefully cleaner.

Good ideal

> Also on first glance i don't see why the buttons are now in 2 options

outer option is for whether or not it needs a ok/cancel button, inner option is for using the default label from the system or a custom label

FabianLars:
> If we don't split the struct then do we just add a field like what we did for parent before?

yes

FabianLars:
> outer option is for whether or not it needs a ok/cancel button, inner option is for using the default label from the system or a custom label

hmmm okay, i'll check it out again later then, thanks

Legend-Master:
> > If we don't split the struct then do we just add a field like what we did for parent before?
> 
> yes

That way we will store the info twice, one in our builder one in rfd's builder, doesn't feel right to me 🤔

I'm kinda lost to be honest, not sure if I can make this any better than before other than the default label thing, do you have any ideas? Maybe I'll just open a separate PR for the default labels?

I'm thinking we just mirror the way rfd's buttons work instead of using labels here, they don't really make much sense to be honest

Legend-Master:
@FabianLars Do you have any ideas or suggestions? If we don't have a clue on how to refactor it yet, I want to make another PR changing `ok_button_label` and `cancel_button_label` to `buttons` like the rfd one for now

FabianLars:
Sorry I wasn't able to look at it over the week and now I'm out sick, do what you want/ and act like I never said anything x)

Legend-Master:
No worries, take care, I'll close this for now, feel free to keep the conversation if/when you got some ideas

PR #1814, opened 2024-09-20
feat(updater)!: add option to use insecure transport protocol

### LINKED ISSUE #4687: [feat] allow http for local updater distribution
### Describe the problem

im on a network where the clients are not exposed to the internet and i need to push updates to their machines. it is terribly inconvenient to have to set up SSL (exhibit a: i don't know how) when my updater server can just be a simple express http server. there is no need for the security hoops to jump through when i'm just pushing updates on a local network.

allow http updater endpoints when just on a local network.

### Alternatives considered

_No response_

### Additional context

_No response_

amrbashir:
closes tauri-apps/tauri#4687


plugins/updater/src/config.rs
```diff
@@ -91,47 +91,57 @@ where
...
     {
-        let url = Url::deserialize(deserializer)?;
-
-        if url.scheme() != "https" {
-            #[cfg(debug_assertions)]
-            eprintln!("[\x1b[33mWARNING\x1b[0m] The configured updater endpoint doesn't use `https` protocol. This is allowed in development but will fail in release builds.");
+        #[derive(Deserialize)]
+        #[serde(rename_all = "camelCase")]
+        pub struct Config {
+            #[serde(alias = "dangerous-insecure-transport-protocol")]
```
amrbashir:
```suggestion
            #[serde(default, alias = "dangerous-insecure-transport-protocol")]
```

.changes/updater-endpoints-result.md
tweidinger, line 1: Shouldn't this be a `!`? Wasn't aware we break. Can you make sure that on release we have migration guide and migration tool to handle this?

plugins/updater/src/config.rs
```diff
@@ -91,47 +91,57 @@ where
...
+            pub windows: Option<WindowsConfig>,
+        }
 
-            #[cfg(not(debug_assertions))]
-            return Err(serde::de::Error::custom(
-                "The configured updater endpoint must use the `https` protocol.",
-            ));
+        let config = Config::deserialize(deserializer)?;
+
+        if !config.dangerous_insecure_transport_protocol {
```
tweidinger: Can you rewrite this for positive matching (when flag x then y, otherwise default behavior)? I assume we will add other methods or flags and having a structure to fit this in would be great.

plugins/updater/src/updater.rs
```diff
@@ -147,9 +147,17 @@ impl UpdaterBuilder {
         self
     }
 
-    pub fn endpoints(mut self, endpoints: Vec<Url>) -> Self {
+    pub fn endpoints(mut self, endpoints: Vec<Url>) -> Result<Self> {
+        if !self.config.dangerous_insecure_transport_protocol {
```
tweidinger: I think we should create a validate endpoint function (or go with the old custom struct and do validation inside it) and call it here? Can be also used during config and we don't have risk of validation mismatch. We do validation in two places with the same code right now.

tweidinger (changes requested):
Some changes from my perspective to make this design resistant to future changes and clarify the breaking change.

.changes/updater-endpoints-result.md
amrbashir: We can't really migrate Rust code. The migration should be fairly simple, no need for a guide IMO and it is only a breaking change between two alpha versions.

plugins/updater/src/config.rs
amrbashir: This doesn't make sense since only if this flag is not enabled, that we check and error out if needed. Doing it the other way around will just result in more boilerplate and confusion.

plugins/updater/src/updater.rs
amrbashir: shared the validation logic in [c592c87e079b6fb3d53f3c2ccce60146b953acee](https://github.com/tauri-apps/plugins-workspace/pull/1814/commits/c592c87e079b6fb3d53f3c2ccce60146b953acee)

.changes/updater-endpoints-result.md
tweidinger: We are in `rc` from an outsider perspective and yeah just a short doc of the migration would be helpful (tauri docs migration guide). Also you're right we don't do rust migrations so just a note in the migration guide.

plugins/updater/src/config.rs
tweidinger: Agreed that it wouldn't be the most optimal code for this one case. With your latest changes this also makes my above comment obsolete.

tweidinger (approved):

plugins/updater/src/config.rs
```diff
@@ -91,47 +91,69 @@ where
...
+pub(crate) fn validate_endpoints(
+    endpoints: &[Url],
+    dangerous_insecure_transport_protocol: bool,
+) -> crate::Result<()> {
+    if !dangerous_insecure_transport_protocol {
+        for url in endpoints {
+            #[cfg(debug_assertions)]
             #[cfg(debug_assertions)]
-            eprintln!("[\x1b[33mWARNING\x1b[0m] The configured updater endpoint doesn't use `https` protocol. This is allowed in development but will fail in release builds.");
+            eprintln!("[\x1b[33mWARNING\x1b[0m] The updater endpoint \"{url}\" doesn't use `https` protocol. This is allowed in development but will fail in release builds.");
```
lucasfernog, line 147: this is printing a warning without even checking the scheme, we must be careful with this kind of change

PR #1822, opened 2024-09-21
[positioner] handleIconState in JS

jbolda:
## Motivation

We need to handle events to update the icon state in this plugin. Previously, we could only do this in Rust, which meant it was not possible create the TrayIcon in JS with an `action`. This adds a function to handle the same.


jbolda:
I think our types might be wrong here.
image

amrbashir:
@jbolda that is indeed the case, the Rust type is actually an enum:
```rs
enum Position {
  Physical { .. },
  Logical { .. },
}
```
and it will be serialized into 
```json
{
  "Physical or Logical": { .. }
}
```

So we need to at runtime, convert these values into the JS `PhysicalPosition` class

lucasfernog (approved):

jbolda:
Looks like the JS vs Rust have a mismatch last commit. /cc @lucasfernog 
https://github.com/jbolda/stream-todo/pull/1/files#diff-28a82398545129d0c8905ddb24bf85b6d28af04026fb6d45394899bd1ad777ac

amrbashir:
@jbolda  fixed in https://github.com/tauri-apps/plugins-workspace/pull/1858

PR #1837, opened 2024-09-24
fix(fs): ignore OS specific paths in scope deserialization

### LINKED ISSUE #9205: [bug][v2] Adding OS specific paths in fs scope breaks scope checks for all paths
### Describe the bug

Can't set below permissions on macOS,  error: "error deserializing scope: unknown path".
but included in `../gen/schemas/macOS-schema.json`.
```json
{
  "permissions": [
    "fs:allow-exe-meta",
    "fs:allow-exe-meta-recursive",
    "fs:allow-exe-read",
    "fs:allow-exe-read-recursive",
    "fs:allow-exe-write",
    "fs:allow-exe-write-recursive"
]
```

### Reproduction

_No response_

### Expected behavior

_No response_

### Full `tauri info` output

```text
[⚠] Environment
    - OS: Mac OS 13.6.4 X64
    ✔ rustc: 1.76.0 (07dca489a 2024-02-04) (Homebrew)
    ✔ cargo: 1.76.0
    ⚠ rustup: not installed!
      If you have rust installed some other way, we recommend uninstalling it
      then use rustup instead. Visit https://rustup.rs/
    ⚠ Rust toolchain: couldn't be detected!
      Maybe you don't have rustup installed? if so, Visit https://rustup.rs/
    - node: 18.18.2
    - npm: 9.8.1

[-] Packages
    - tauri [RUST]: 2.0.0-beta.11
    - tauri-build [RUST]: 2.0.0-beta.9
    - wry [RUST]: 0.37.0
    - tao [RUST]: 0.26.1
    - @tauri-apps/api [NPM]: 2.0.0-beta.5
    - @tauri-apps/cli [NPM]: 2.0.0-beta.9

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - framework: Vue.js
    - bundler: Vite
```

### Stack trace

_No response_

### Additional context

_No response_

FabianLars:
fixes https://github.com/tauri-apps/tauri/issues/9205

ngl never really looked into the ACL stuff so this is probably not the best solution but it's what i came up with 🤷

### DIFF
--- a/.changes/fix-fs-scope-unknown-path.md
+++ b/.changes/fix-fs-scope-unknown-path.md
@@ -2,4 +2,4 @@
 fs: patch
 ---
 
-Fixed an issue causing any `fs` APIs to throw an `error deserializing scope: unknown path` error when the permissions/scopes allowed OS specific paths that aren't available on the currently running OS.
+Fix failing to deserialize capability file when using an OS specific path in the scope that is not available on the current OS.
--- a/.changes/fix-fs-scope-unknown-path.md
+++ b/.changes/fix-fs-scope-unknown-path.md
@@ -1,0 +1,5 @@
+---
+fs: patch
+---
+
+Fixed an issue causing any `fs` APIs to throw an `error deserializing scope: unknown path` error when the permissions/scopes allowed OS specific paths that aren't available on the currently running OS.
--- a/plugins/fs/src/lib.rs
+++ b/plugins/fs/src/lib.rs
@@ -360,6 +360,7 @@ impl ScopeObject for scope::Entry {
 
         match app.path().parse(path) {
             Ok(path) => Ok(Self { path: Some(path) }),
+            #[cfg(not(target_os = "android"))]
             Err(tauri::Error::UnknownPath) => Ok(Self { path: None }),
             Err(err) => Err(err.into()),
         }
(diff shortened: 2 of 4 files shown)


plugins/fs/src/lib.rs
```diff
@@ -353,17 +353,17 @@ impl ScopeObject for scope::Entry {
...
+            scope::EntryRaw::Value(path) => path,
+            scope::EntryRaw::Object { path } => path,
         })?;
 
-        Ok(Self {
-            path: app.path().parse(entry.path)?,
-        })
+        match app.path().parse(path) {
+            Ok(path) => Ok(Self { path: Some(path) }),
+            #[cfg(not(target_os = "android"))]
```
amrbashir, line 363: why not android as well?

.changes/fix-fs-scope-unknown-path.md
```diff
@@ -0,0 +1,5 @@
+---
+fs: patch
+---
+
+Fixed an issue causing any `fs` APIs to throw an `error deserializing scope: unknown path` error when the permissions/scopes allowed OS specific paths that aren't available on the currently running OS.
```
amrbashir:
```suggestion
Fix failing to deserialize capability file when using an OS specific path in the scope that is not available on the current OS.
```

plugins/fs/src/lib.rs
FabianLars: UnknownPath has that cfg so without this it won't compile.

I'll check if Android has a similar issue

FabianLars: Android works fine as it is

amrbashir (approved):

PR #1838, opened 2024-09-24
refactor(sql): Allow multiple drivers at the same time

FabianLars:
I know the code is absolutely disgusting. I just don't have enough time for this atm but really want to unblock this. Also, all those generics (in sqlx) are a pain to work with.

Opening this as a draft because i'll head off for today but wanted to kick off CI already. Feel free to take a look already though.


FabianLars:
btw since the api didn't change except for the DbPool type i'd prefer to merge this quickly and iterate over the actual implementation later, especially since i'll be ooo for at least a week.

I removed the public apis for now (except the DbPool enum itself of course) for that.

lucasfernog (approved):

PR #1839, opened 2024-09-25
fix(shell): fix schema requiring `sidecar` property even though it is optional

amrbashir:


FabianLars:
schema noob here: Is it possible to say that either cmd or sidecar is required? Because now only `name` is required and that alone doesn't do anything. No biggie if this is not possible but wanted to ask just in case it is.

amrbashir:
> schema noob here: Is it possible to say that either cmd or sidecar is required? Because now only `name` is required and that alone doesn't do anything. No biggie if this is not possible but wanted to ask just in case it is.

It is probably possible with a custom JsonSchema implementation, I don't think we can do it with the derive macro

FabianLars (dismissed):

amrbashir:
@FabianLars actually we could just use untagged enum, take a look and see if that's what you meant

FabianLars:
I don't have access to a computer until tomorrow so I can't check out the generated schema but I assume this is what I meant, or at least close enough x)

FabianLars:
Okay so I got my hands on my laptop for a minute 😂 It seems to work, cool. vscode just seems to think that args is always required though.

Laptop is gone again so now I really can't test again until tomorrow but with just the args thing I don't think that's a problem

amrbashir:
alright, fixed that one as well

lucasfernog (approved):
i love you

PR #1858, opened 2024-10-02
fix(positioner): pass correct values through IPC

amrbashir:
followup to https://github.com/tauri-apps/plugins-workspace/pull/1822


lucasfernog (approved):

lucasfernog-crabnebula:
I actually think this broke the impl - we expect a PhysicalPosition and not a Position in the command, reverted on #1911

PR #1860, opened 2024-10-02
refactor(store)!: more reworks

### LINKED ISSUE #984: [store] Absolute vs. relative paths to the same store file are cached separately
Came across an interesting edge case today in the store plugin. If I create a store in JS like this:
```javascript
const store = new Store(appDataDir() + "config.file")
```
and I try to access it from Rust like this:
```rust
let stores = app.state::<StoreCollection<Wry>>();
let path = PathBuf::from("config.file");

with_store(app_handle, stores, path, |store| store.insert("a".to_string(), json!("b")))
```
these will load from and save to the same file on the filesystem (thanks to https://github.com/tauri-apps/plugins-workspace/blob/v1/plugins/store/src/store.rs#L179), but they'll be cached separately because `with_store` just uses the key you give it (which in this case is an absolute path in JS and a relative path in Rust) - the line in question is here: https://github.com/tauri-apps/plugins-workspace/blob/v1/plugins/store/src/lib.rs#L57.

Since these two paths point to the same store file and load from/save to the same file, it feels to me like `with_store` should key on the absolute path constructed by `load` and `save` instead of just whatever the caller gives it.

Happy to throw together a quick PR in this direction if we think it makes sense!

### LINKED ISSUE #1690: Store json file can be save format beautify style?
image
to this
image

### LINKED ISSUE #1865: tauri-plugin-store: typescript mismatch for attribute autoSave
The `autoSave` has the wrong type in the typescript definitions.

```
export type StoreOptions = {
    /**
     * Auto save on modification with debounce duration in milliseconds
     */
    autoSave?: boolean;
};
```

`autoSave` should be of type number, not boolean.

### LINKED ISSUE #1877: After upgrade to v2: invalid args `rid` for command `get`: command get missing required key rid
I'm seeing many of these console errors in the webview2 inspector after updating tauri and tauri-plugin-store from 2.0.0-rc to 2.0.0 final. The error points to the plugin-store, or my call point: `store.get("my-key")`. No code changed, just Cargo and npm package upgrades to @tauri-apps v2 release.

My code:

```js
import { Store } from "@tauri-apps/plugin-store"

const Setting = {
  editorOpen: "editor-visible",
  windowSplit: "window-split"
}

class WindowSettings {
  #defaults
  #store

  constructor(defaults, store = new Store(".settings.dat")) {
    this.#defaults = defaults
    this.#store = store
  }

  #saveSetting(key, value) {
    this.#store.set(key, value).then(() => this.#store.save())
  }

  /**
   * Return bounded split value in pixels
   * @returns number
   */
  getSplit = () =>
    this.#store.get(Setting.windowSplit).then(split =>
      typeof split !== "number"
        ? this.#defaults.split.min
        : Math.min(
            Math.max(split, this.#defaults.split.min),
            this.#defaults.split.max
          )
    )

  setSplit = px => {
    this.#saveSetting(Setting.windowSplit, px)
  }
}
```

image

I've tried changing my key names and store file (`.dat`), neither helped.

Legend-Master:
Follow up of #1550

Changes:

- Disallow calling `create_store` when there're still active stores with the same path alive
- Save and cancel pending auto save on drop
- Use absolute path as store's key, fix #984

New features:

- Add `getStore`/`get_store` share stores across js and rust side
- Add default (de)serialize functions settings
- Allow js to use pre-stored (de)serialize functions
- Add back lazy store (implemented in js)

Default changes:

- Share store to resource table by default
- Enable auto save with 100ms debounce time by default
- Use pretty json by default, close #1690


lucasfernog:
can't we automatically share the store? like always make it available in the same collection, so a call from another language would pick it up anyway?

Legend-Master:
Yeah I was thinking about it, but that means you'll need to always preserve that store, sometimes you might just want to read the data from it and forget about it, maybe share/preserve it by default, and add a close function or something

Legend-Master:
I need to take a break, this thing is haunting me right now 😂

Legend-Master:
Note: don't merge it yet, I didn't have the time to test and write the docs yet, got some other things to handle, might be able get back on this tomorrow or something like that

plugins/store/src/lib.rs
```diff
@@ -12,164 +12,231 @@
...
+    AppHandle, Manager, ResourceId, RunEvent, Runtime, State,
 };
 
 mod error;
 mod store;
 
 #[derive(Serialize, Clone)]
+#[serde(rename_all = "camelCase")]
 struct ChangePayload<'a> {
     path: &'a Path,
```
Legend-Master, line 35: Not sure if we still want to keep `path: &'a Path,`, this used to be the way to identify a store from the js side

Legend-Master:
Did some basic testing and docs, should be ready for merge now

plugins/store/src/lib.rs
```diff
@@ -12,164 +12,260 @@
...
     fn store_builder(&self, path: impl AsRef<Path>) -> StoreBuilder<R>;
+    /// Get an existing store
+    fn get_store(&self, path: impl AsRef<Path>) -> Option<Arc<Store<R>>>;
 }
 
 impl<R: Runtime, T: Manager<R>> StoreExt<R> for T {
-    fn store(&self, path: impl AsRef<Path>) -> Store<R> {
+    fn store(&self, path: impl AsRef<Path>) -> Arc<Store<R>> {
+        self.get_store(&path)
+            .unwrap_or_else(|| self.create_store(&path).unwrap())
```
Legend-Master: This is not thread safe since the store is locked inside of `get_store` and `create_store`, the same is true for the js `LazyStore`, but I'm kinda stuck, I don't know how to actually make this work without duplicating a whole bunch of code

Legend-Master: Well, I went with just duplicating the code 😢

.changes/store-plugin-rework.md
```diff
@@ -0,0 +1,24 @@
+---
+"store": minor:breaking
+---
+
+**Breaking change**: Renamed `StoreCollection` to `StoreState`
```
amrbashir: We can no longer make breaking changes, we need to revert this or keep it as an alias with a deprecation warning.

.changes/store-plugin-rework.md
```diff
@@ -0,0 +1,24 @@
...
+
+Changes:
+
+- Disallow calling `create_store` when there're still active stores with the same path alive
```
amrbashir: Can't we just return the active one then?

examples/api/src/views/Store.svelte
```diff
@@ -1,34 +1,71 @@
...
+    } catch (error) {
+      onMessage(error);
+    }
+    await refreshEntries();
+  }
+
+  async function close() {
+    try {
+      await store.close();
+      onMessage("Store is now closed, any new operations will now errors out");
```
amrbashir:
```suggestion
      onMessage("Store is now closed, any new operations will error out");
```

plugins/store/README.md
```diff
@@ -81,14 +81,11 @@ if (val) {
...
 }
-
-// This manually saves the store.
-await store.save()
 ```
 
 ### Persisting Values
 
-As seen above, values added to the store are not persisted between application loads unless the application is closed gracefully.
+Modifications mode to the store are automatically saved by defaut
```
amrbashir:
```suggestion
Modifications made to the store are automatically saved by default
```

plugins/store/guest-js/index.ts
```diff
@@ -8,50 +8,225 @@ import { invoke, Resource } from '@tauri-apps/api/core'
...
+  }
+}
+
+/**
+ * A key-value store persisted by the backend layer.
+ */
+export class Store extends Resource implements IStore {
+  private constructor(
+    rid: number
+    // private readonly path: string
```
amrbashir: no longer needed? just remove then

plugins/store/guest-js/index.ts
```diff
@@ -8,50 +8,225 @@ import { invoke, Resource } from '@tauri-apps/api/core'
...
+ */
+export async function getStore(path: string): Promise<Store | undefined> {
+  return await Store.getStore(path)
 }
 
 /**
  * A lazy loaded key-value store persisted by the backend layer.
  */
-export class Store extends Resource {
+export class LazyStore implements IStore {
```
amrbashir, line 81: I would also consider this a breaking change as well and we can't just change the behavior of normal `Store` and add a new `LazyStore`type.

I also honestly don't see why we need two Store types, I believe it can just be one.

plugins/store/src/store.rs
```diff
@@ -12,39 +12,39 @@ use std::{
...
-) -> Result<HashMap<String, JsonValue>, Box<dyn std::error::Error + Send + Sync>> {
-    serde_json::from_slice(bytes).map_err(Into::into)
+pub(crate) fn resolve_store_path<R: Runtime>(
+    app: &AppHandle<R>,
+    path: impl AsRef<Path>,
+) -> PathBuf {
+    dunce::simplified(
+        &app.path()
+            .resolve(path, BaseDirectory::AppData)
+            .expect("failed to resolve app dir"),
```
amrbashir: no expect, return a result instead

plugins/store/src/store.rs
```diff
@@ -174,42 +173,119 @@ impl<R: Runtime> StoreBuilder<R> {
...
+                auto_save: self.auto_save,
+                auto_save_debounce_sender: Arc::new(Mutex::new(None)),
+                store: Arc::new(Mutex::new(store_inner)),
+            };
+
+            let store = Arc::new(store);
+            let rid = self.app.resources_table().add_arc(store.clone());
+            stores.insert(self.path, rid);
+            (store, rid)
+        }
```
amrbashir: this should call `self.build_inner()` and `build_or_existing` should return a result as well

plugins/store/src/lib.rs
```diff
@@ -12,164 +12,297 @@
...
-    let store = webview.resources_table().get::<Store<R>>(rid)?;
+async fn save<R: Runtime>(app: AppHandle<R>, rid: ResourceId) -> Result<()> {
+    let store = app.resources_table().get::<Store<R>>(rid)?;
     store.save()
 }
 
 pub trait StoreExt<R: Runtime> {
-    fn store(&self, path: impl AsRef<Path>) -> Store<R>;
+    /// Create a store or get an existing store with default settings at path
+    fn store(&self, path: impl AsRef<Path>) -> Arc<Store<R>>;
```
amrbashir: This is another breaking change

plugins/store/src/store.rs
```diff
@@ -174,42 +173,119 @@ impl<R: Runtime> StoreBuilder<R> {
...
+    /// # Examples
+    /// ```
+    /// tauri::Builder::default()
+    ///   .plugin(tauri_plugin_store::Builder::default().build())
+    ///   .setup(|app| {
+    ///     let store = tauri_plugin_store::StoreBuilder::new(app, "store.json").build()?;
+    ///     Ok(())
+    ///   });
+    /// ```
+    pub fn build(self) -> crate::Result<Arc<Store<R>>> {
```
amrbashir, line 237: This is also a breaking change

plugins/store/src/store.rs
```diff
@@ -428,19 +538,37 @@ impl<R: Runtime> Store<R> {
...
+    pub fn close_store(self) {
+        let store = self.store.lock().unwrap();
+        let app = store.app.clone();
+        let state = app.state::<StoreState>();
+        let stores = state.stores.lock().unwrap();
+        if let Some(rid) = stores.get(&store.path).copied() {
+            drop(store);
+            drop(stores);
+            let _ = app.resources_table().take::<Store<R>>(rid);
+        }
```
amrbashir: this should be a drop implementation instead

plugins/store/src/store.rs
```diff
@@ -428,19 +538,37 @@ impl<R: Runtime> Store<R> {
...
+    /// and the store is only truely closed once all reference held by you are also dropped
+    pub fn close_store(self) {
+        let store = self.store.lock().unwrap();
+        let app = store.app.clone();
+        let state = app.state::<StoreState>();
+        let stores = state.stores.lock().unwrap();
+        if let Some(rid) = stores.get(&store.path).copied() {
+            drop(store);
+            drop(stores);
+            let _ = app.resources_table().take::<Store<R>>(rid);
```
amrbashir: why not use `app.resources_table().close()`

.changes/store-plugin-rework.md
Legend-Master: That is in `createOrExistingStore` now

.changes/store-plugin-rework.md
Legend-Master: Isn't the stability of plugins are still undetermined?

plugins/store/src/store.rs
Legend-Master: This is to remove it from the resource table, we can't really do it in drop

plugins/store/guest-js/index.ts
Legend-Master: The normal `Store`'s behavior was changed in the last rework PR, this PR didn't change anything about it, just added a new one

The normal `Store` is a primitive if they need more fine grained control, but If you think it's not needed, we can just keep the `LazyStore` and name it `Store`

plugins/store/src/store.rs
Legend-Master: That means we need to return result in `StoreBuilder::new` and all the related places, will this ever fail?

plugins/store/src/store.rs
Legend-Master: > this should call `self.build_inner()`

As I said here https://github.com/tauri-apps/plugins-workspace/pull/1860#discussion_r1787633128 I didn't figure out of way of doing that

> `build_or_existing` should return a result as well

`build` currently can only errors out if there's an existing store, `build_or_existing` doesn't make much sense to return an error

.changes/store-plugin-rework.md
amrbashir: yeah, I was talking in terms of breaking changes cause this sounds like another breaking change or a bug fix, debatable tbh.

plugins/store/guest-js/index.ts
amrbashir: I see, so it is just outdated/incorrect documentation then. I am fine with keeping `LazyStore` but I don't find a really good use for it? I think `Store` should just be initialized once the class is constructed and so users have full control over the open/read/write/close

plugins/store/src/store.rs
amrbashir: we can delay resolving the path until `StoreBuilder::build` and there we can return a result.

plugins/store/src/store.rs
amrbashir: why not? this `close_store` method already takes ownership of `self` so it can no longer be used which is what the drop implementation does.

plugins/store/guest-js/index.ts
Legend-Master: Actually I would want to promote `LazyStore` to be used in most cases since you don't need to handle the async creation of the store and also you get the lazy load

plugins/store/src/store.rs
Legend-Master: I don't know if we want to make `app.store("some-store")` return a result to be honest, also what about `get_store`?

plugins/store/src/store.rs
Legend-Master: This is should be used from an `Arc` and since another reference is in the resource table, this wouldn't necessarily drop the `Store` I don't think

plugins/store/src/store.rs
amrbashir: any function that calls `resolve_store_path` should propagate the error, we can't ensure `app.path().resolve()` will always work.

plugins/store/src/store.rs
amrbashir: how about?
```rs
     {
        let state = self.app.state::<StoreState>();
        let mut stores = state.stores.lock().unwrap();

        if let Some(rid) = stores.get(&self.path) {
            return (self.app.resources_table().get(*rid).unwrap(), *rid);
        }
     }    
     
     self.build_inner()  
```

> `build` currently can only errors out if there's an existing store, `build_or_existing` doesn't make much sense to return an error

If it calls `self.build_inner`, it will need to propagate its result. Also once `resolve_store_path` is changed to return a result and moved to `build` method, it will need to return a result as well

plugins/store/src/store.rs
Legend-Master: Ok

plugins/store/src/store.rs
Legend-Master: If we change it to return a result then this will work, before this, we can't guarantee the thread safety

plugins/store/guest-js/index.ts
amrbashir: Let's keep it, although I wouldn't want it to be the default in README and docs

plugins/store/src/store.rs
amrbashir: taking `self` will not work with `Arc`, it will fail with compile error. You need to remove/unwrap the `Arc` in order to call this method. At which point, you can just drop it

Legend-Master: Got it, that makes sense

What's your suggestion on this, a static method that takes `Arc<Self>` instead or put the method on `Manager`? Or something else?

amrbashir: if the idea is to only remove resources, just take `&self` and call it `close_resource` or something

Legend-Master:
I just realized, will this work for mobile? I'm seeing that #1695 removed the mobile plugins and seems to now using rust's std fs with path from `AppData` path which I don't think it actually support android and ios

.changes/store-plugin-rework-js-feat.md
```diff
@@ -0,0 +1,7 @@
+---
+"store-js": minor:feat
+---
+
+- Add `getStore`
+- Add an option to use pre-stored (de)serialize functions
+- Add back lazy store `LazyStore`
```
amrbashir:
```suggestion
- Add `LazyStore`
```

.changes/store-plugin-rework-js-feat.md
```diff
@@ -0,0 +1,7 @@
+---
+"store-js": minor:feat
+---
+
+- Add `getStore`
+- Add an option to use pre-stored (de)serialize functions
```
amrbashir:
```suggestion
- Add an option to use pre-stored (de)serialize functions (registered on rust)
```

.changes/store-plugin-rework.md
```diff
@@ -0,0 +1,24 @@
+---
+"store": minor:breaking
+---
+
+**Breaking changes**:
```
amrbashir:
```suggestion
### Breaking changes:
```

.changes/store-plugin-rework.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Renamed `StoreCollection` to `StoreState`
+- `StoreBuilder::build` now returns a `Result`
+- `StoreExt::store` now returns `Result<Arc<Store>>`
+
+Other Changes:
```
amrbashir:
```suggestion
### Enhancements:
```

.changes/store-plugin-rework.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Save and cancel pending auto save on drop
+- Use absolute path as store's key, fix #984
+- Share store to resource table by default
+- Enable auto save with 100ms debounce time by default
+- Use pretty json by default, close #1690
+
+New features:
```
amrbashir:
```suggestion
### New features:
```

.changes/store-plugin-rework.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Add `getStore`/`get_store` share stores across js and rust side
```
amrbashir:
```suggestion
- Add `get_store` to get shared stores across js and rust side
```

.changes/store-plugin-rework.md
```diff
@@ -0,0 +1,24 @@
...
+- Add default (de)serialize functions settings
```
amrbashir: should mention the new APIs instead

.changes/store-plugin-rework.md
```diff
@@ -0,0 +1,24 @@
...
+- Allow js to use pre-stored (de)serialize functions
```
amrbashir: this is covered by the JS change file, no need to mention here I guess

.changes/store-plugin-rework.md
```diff
@@ -0,0 +1,24 @@
...
+- Add back lazy store (implemented in js)
```
amrbashir: this doesn't belong here

.changes/store-plugin-rework.md
amrbashir: or rather mention the relevant public Rust APIs that are used to register these functions

plugins/store/README.md
```diff
@@ -68,9 +68,9 @@ fn main() {
 Afterwards all the plugin's APIs are available through the JavaScript guest bindings:
 
 ```typescript
-import { Store } from '@tauri-apps/plugin-store'
+import { LazyStore } from '@tauri-apps/plugin-store'
```
amrbashir: Should be kept as `Store`, we can mention `LazyStore` afterwards in a note.

plugins/store/README.md
```diff
@@ -108,60 +105,28 @@ await store.load()
...
-As you may have noticed, the `Store` crated above isn't accessible to the frontend. To interoperate with stores created by JavaScript use the exported `with_store` method:
-
-```rust
-use tauri::Wry;
-use tauri_plugin_store::StoreExt;
-
-let store = app.store_builder("app_data.bin").build();
-store.insert("key", "value");
-```
+The store created from both Rust side and JavaScript side are stored in the app's resource table and can be accessed by both sides, you can access it by using the same path, with `getStore` and `LazyStore` in the JavaScript side and `get_store` and `store` in the Rust side
```
amrbashir, line 139:
```suggestion
The store created from both Rust side and JavaScript side are stored in the app's resource table and can be accessed by both sides, you can access it by using the same path, with `getStore` and `Store` in the JavaScript side and `get_store` and `Store` in the Rust side
```

plugins/store/guest-js/index.ts
```diff
@@ -8,50 +8,211 @@ import { invoke, Resource } from '@tauri-apps/api/core'
...
+  ) {}
+
+  /**
+   * Init/load the store if it's not already
+   */
+  async init(): Promise<void> {
+    await this.store
+  }
+
+  async set(key: string, value: unknown): Promise<void> {
```
amrbashir, line 108: the methods of `LazyStore` and `Store` should have documentation here instead of `IStore`

plugins/store/src/lib.rs
```diff
@@ -12,164 +12,297 @@
...
     rid: ResourceId,
     key: String,
-) -> Result<Option<JsonValue>> {
-    let store = webview.resources_table().get::<Store<R>>(rid)?;
-    Ok(store.get(key))
+) -> Result<(Option<JsonValue>, bool)> {
+    let store = app.resources_table().get::<Store<R>>(rid)?;
+    let value = store.get(key);
+    let exists = value.is_some();
+    Ok((value, exists))
```
amrbashir, line 157: why return a bool here? returning `Option<T>` will result in `T | null` on JS side so we don't need to check `is_some` manually

plugins/store/src/store.rs
```diff
@@ -361,38 +427,45 @@ impl<R: Runtime> std::fmt::Debug for StoreInner<R> {
...
+}
 
 impl<R: Runtime> Store<R> {
-    pub fn with_store<T>(
-        &self,
-        f: impl FnOnce(&mut StoreInner<R>) -> crate::Result<T>,
-    ) -> crate::Result<T> {
+    /// Do something with the inner store,
+    /// useful for batching some work if you need higher performance
+    pub fn with_store<T>(&self, f: impl FnOnce(&mut StoreInner<R>) -> T) -> T {
```
amrbashir: sounds like we need a better API, exposing `StoreInner` is not a good thing IMO

amrbashir (changes requested):

plugins/store/src/lib.rs
Legend-Master: `null` and `undefined` means different things, `null` means a `null` in the json file, `undefined` means no value exists in that key

plugins/store/README.md
Legend-Master: I would like to push people to use `LazyStore` most of the times, what's your thoughts?

plugins/store/guest-js/index.ts
Legend-Master: They share the same docs, why do we want to duplicate that?

plugins/store/README.md
amrbashir: I usually don't like hidden/magic code, and so `LazyStore` should be sort of opt-in when you understand how it behaves, otherwise I fear we will receive issues that goes like this "I copied this snippet, but the behavior is not as I expected"

amrbashir: Showing what `LazyStore` offers after showing `Store` will explain it better and shows how it enhances over `Store`.

plugins/store/guest-js/index.ts
amrbashir: My main concern is hover documentation and documentation on our website, also `IStore` shouldn't be public as well, which could mess up the target documentation cases (didn't test).

plugins/store/src/lib.rs
amrbashir: I see, this is fine then

plugins/store/guest-js/index.ts
Legend-Master: Hover documentation works fine for me, not sure the docs on the website though, I think `IStore` needs to be public so people could use the two stores interchangeably

plugins/store/README.md
Legend-Master: I think a lazy store is what most people want, I would prefer we document the primitives after the higher level API `LazyStore` in this case

plugins/store/src/store.rs
Legend-Master: I think exposing it here is fine, the reason for the outer `Store` is really just because of auto save and shared references, the inner store is the actual APIs and the outer store is just a convenient wrapper

amrbashir: I really don't think it should be exposed, inner structs should always be internal, surface API should be refactor and changed to hide it.

amrbashir: Even if `InnerStore` and `Store` share the same APIs, never expose it in a user API. It will only cause confusion and won't allow us to modify this type freely.

Legend-Master: I don't want this either to be honest, but people might want higher performance operations or if they want to batch some modifications atomically

What about we remove this function for now and see if people actually need it

amrbashir: sure, if we want to do patch operations, we may want to rethink the API with a different one.

plugins/store/README.md
amrbashir: Doesn't make sense to show `LazyStore` after `Store`. As a user, I wouldn't understand what `LazyStore` does unless I understand `Store` first.

plugins/store/guest-js/index.ts
amrbashir: but why would user need to use two types of stores interchangeably?

Legend-Master: Well, you might want to put the store inside a context (in react/solidjs/etc term) and type it as `IStore` instead of a specific store, if you really don't like this we can not to expose it for now and see if anyone requests for it

plugins/store/README.md
Legend-Master: `LazyStore` works exactly like the old `Store`, and it's quite simple to understand, but documenting the primitives also make sense, I'll change it

plugins/store/guest-js/index.ts
amrbashir: but why would I store in a context with `IStore`, if I am using `LazyStore` why not just store it as `LazyStore`. I fail to see a real use-case here tbh, I'd recommend removing.

amrbashir: Also realistically speaking you're gonna use one or the other but never both. And it is part of why I don't really like the addition of `LazyStore` but I am willing to keep it for now.

amrbashir: I honestly think the better usage is through `createStore` and `Store`.

`LazyStore` feels kinda magical and it also delays the errors emitted during initialization of the store until a method is used which will confuse the user as to why this method is throwing this unrelated error.

Legend-Master: Just realized that `IStore` was never exposed

> Also realistically speaking you're gonna use one or the other but never both

That's fair

> I honestly think the better usage is through createStore and Store.
> 
> LazyStore feels kinda magical and it also delays the errors emitted during initialization of the store until a method is used which will confuse the user as to why this method is throwing this unrelated error.

I thought this way as well first when I made the last rework, but then I realized that in UI usages, something like `localStorage` is the easiest to use and despite the downsides, not many people would choose indexeddb over it in simple use cases, I think this is somewhat true for our case as well

plugins/store/guest-js/index.ts
```diff
@@ -8,50 +8,250 @@ import { invoke, Resource } from '@tauri-apps/api/core'
...
+ * @param path Path to save the store in `app_data_dir`
+ * @param options Store configuration options
  */
-export async function createStore(path: string, options?: StoreOptions) {
-  const resourceId = await invoke<number>('plugin:store|create_store', {
-    path,
-    ...options
-  })
-  return new Store(resourceId, path)
+export async function newOrExisting(
```
amrbashir: I don't quite like this name, I think it should be
```suggestion
export async function createNewOrLoadExisting(
```
or `createNewOrGetExisting`

plugins/store/guest-js/index.ts
```diff
@@ -8,50 +8,250 @@ import { invoke, Resource } from '@tauri-apps/api/core'
...
+  path: string,
+  options?: StoreOptions
+): Promise<Store> {
+  return await Store.newOrExisting(path, options)
+}
+
+/**
+ * @param path Path of the store
+ */
```
amrbashir: Should probably match the documentation of `Store.get` which has a bit more info

plugins/store/guest-js/index.ts
```diff
@@ -8,50 +8,250 @@ import { invoke, Resource } from '@tauri-apps/api/core'
...
+   * let store = await Store.get('store.json');
+   * if (!store) {
+   *   store = await Store.newOrExisting('store.json');
+   * }
+   * ```
+   *
+   * @param path Path to save the store in `app_data_dir`
+   * @param options Store configuration options
    */
+  static async newOrExisting(
```
amrbashir: same as above, it should be `createNewOrLoadExisting` or `createNewOrGetExisting`

plugins/store/src/store.rs
```diff
@@ -174,142 +164,206 @@ impl<R: Runtime> StoreBuilder<R> {
...
-            });
-        drop(stores);
-        Store {
-            defaults: self.defaults,
-            serialize: self.serialize,
-            deserialize: self.deserialize,
-            auto_save: self.auto_save,
-            auto_save_debounce_sender: Arc::new(Mutex::new(None)),
-            store,
+    pub fn create(self) -> crate::Result<Arc<Store<R>>> {
```
amrbashir: I don't quite like `create` on a builder, I think it should be kept as `build`

plugins/store/src/store.rs
```diff
@@ -174,142 +164,206 @@ impl<R: Runtime> StoreBuilder<R> {
...
+    /// # Examples
+    /// ```
+    /// tauri::Builder::default()
+    ///   .plugin(tauri_plugin_store::Builder::default().build())
+    ///   .setup(|app| {
+    ///     let store = tauri_plugin_store::StoreBuilder::new(app, "store.json").new_or_existing();
+    ///     Ok(())
+    ///   });
+    /// ```
+    pub fn new_or_existing(self) -> crate::Result<Arc<Store<R>>> {
```
amrbashir: this can also be `build_new_or_load_existing` or simpler `build_new_or_existing`

amrbashir (changes requested):

plugins/store/guest-js/index.ts
Legend-Master: I would probably like `createOrExisting` a bit more like how it was @lucasfernog

plugins/store/src/store.rs
Legend-Master: +1 for `build` and `build_or_existing` @lucasfernog

plugins/store/guest-js/index.ts
lucasfernog: `createOrExisting` just feels like it's missing a word in there.. omg all these names are awful

lucasfernog:
@amrbashir @Legend-Master i've removed the rust's `create` from the StoreBuilder and renamed `new_or_existing` to `build`.

I also renamed the `load` function to `reload` just to be able to call the `newOrExisting` function `load`. Let me know what you think.

lucasfernog:
if you're still not happy with this name then we can just agree we're all awful at this and we'll never agree, and remove the Store class and just go with the LazyStore as default (that way we can just the constructor instead of having to name a static function 😂 )

plugins/store/README.md
```diff
@@ -68,9 +68,9 @@ fn main() {
 Afterwards all the plugin's APIs are available through the JavaScript guest bindings:
 
 ```typescript
-import { Store } from '@tauri-apps/plugin-store'
+import { createStore } from '@tauri-apps/plugin-store'
```
amrbashir: I think the function was renamed in last commit to just `create`

amrbashir (commented):
I think I am happy with latest changes, what do you think @Legend-Master? 

now we just need to update the change files and README and this should be good to go.

plugins/store/guest-js/index.ts
```diff
@@ -8,50 +8,294 @@ import { invoke, Resource } from '@tauri-apps/api/core'
...
+ * let store = await Store.get('store.json');
+ * if (!store) {
+ *   store = await Store.load('store.json');
+ * }
+ * ```
+ *
+ * @param path Path to save the store in `app_data_dir`
+ * @param options Store configuration options
+ *
+ * @throws If a store at that path is already loaded
```
Legend-Master: Isn't this the same as `create`?

plugins/store/src/store.rs
```diff
@@ -174,7 +166,57 @@ impl<R: Runtime> StoreBuilder<R> {
...
+
+        let mut store_inner = StoreInner::new(
+            self.app.clone(),
+            self.path.clone(),
+            self.defaults.take(),
+            self.serialize_fn,
+            self.deserialize_fn,
+        );
+
+        if !self.create_new {
```
Legend-Master, line 207: Why we don't want to load the store here?

Legend-Master (commented):
I'm fine with the new names

plugins/store/src/store.rs
amrbashir: I think because `create_new` means a new Store so don't load from disk and following modifications to the store should override the one stored in disk.

plugins/store/guest-js/index.ts
amrbashir: I don't think so, `create` will create a new store and override old values. This will persist values from disk

plugins/store/src/store.rs
Legend-Master: The current description for it is `/// Force create a new store even if it already exists.`, I think a better way to do this is changing it to `returns Error on build if a store with the same path has already been loaded` and just always load the store on build

amrbashir: what if you want to force create a new one and override old one?

Legend-Master: Then close the old one first, also the logic here doesn't do what it says, it will always return the existing store instead of creating a new one

Legend-Master: @lucasfernog Didn't you mean to change the behavior of `newOrExisting` (or `load` now) or is it a mistake?

amrbashir: maybe we should adjust that logic too, if `create_new` is true, don't return existing store and create new one

Legend-Master: From the description, it seems like if you try to use `create` from js side or `build` with `create_new` when there's an existing store, instead of returning an error, it will remove the old one from the resource table and put a new one in instead

This sounds good to me (I still think even an override needs to call `load`), it's just the descriptions and the implementations are all kinda messed up that none of them align with each other

amrbashir: why call `load` if we overriding?

Legend-Master: Ah, I guess I misunderstood the intention here, `create_new` force creates a new empty store, I was still thinking the old behavior that it loads the store

But honestly I can't think of any use cases for this, from what I think, we can have

- Create store: new store, put it in the resource table, throws if the store's already there (this can be used when you're sure you're only creating it once)
- Get store: get the existing store (get the store if you're sure you have created it)
- Get or create store: get or create in one lock (should be used most of the time, just give me a store, I don't care)

amrbashir: Those cases are fine but creating a new store and overriding existing one can still be used as well, and we can provide an API to make it more performant instead of creating a store then calling `reset` on it. I see this as low hanging fruit that we can just add without breaking other cases.

Legend-Master: Yeah sure, I'll update the code and description to align the behavior, see if this is what you think it should do

plugins/store/README.md
Legend-Master: Since this creates a new store every time, I think we should document `load` instead

Legend-Master:
To be honest, I still find `create` to be very confusing, I don't think anyone will expect it to mean open an empty store in the rust side without loading from disk while `load` gets the opened store or create a new store but also loads it

amrbashir (dismissed):

lucasfernog:
> To be honest, I still find `create` to be very confusing, I don't think anyone will expect it to mean open an empty store in the rust side without loading from disk while `load` gets the opened store or create a new store but also loads it

I've removed the create function from the Rust side, maybe we do the same on the JS side too. I'll push this change for you to take a look.

lucasfernog:
@amrbashir i wanted to use the Resource class close() method instead of creating a dedicated one but I can't because we're using the AppHandle resource table and not the Webview one.. should the resources plugin allow me to configure which table i'm referring to? or fallback to other tables?

amrbashir:
I was thinking something similar but Idk if the auditors will approve, maybe if we allow only closing resources from the same webview, or the same window or the app handle, It should be fine then.

amrbashir:
I.e. webview can't close resources from other webviews.

lucasfernog:
yeah so fallback to the parent window or AppHandle, that should be good

lucasfernog:
usually you shouldn't even know the RID if you can't also close the resource.. otherwise users can use a wrapper ID instead

lucasfernog:
all good for us to merge and release this? @amrbashir and @Legend-Master 
no more breaking changes pls :D

amrbashir (approved):

lucasfernog (approved):

Legend-Master:
Except for `close` doesn't actually work right now, looks good to me

lucasfernog:
we'll fix on the tauri side

PR #1863, opened 2024-10-02
feat: update to tauri v2 stable, promote all packages to 2.0.0

lucasfernog:


FabianLars:
wasn't the idea to _not_ do this? don't care too much, i can introduce breaking changes in a non breaking way if needed, just surprised. I guess this is easier for consumers 🤷

lucasfernog:
we agreed on possibly doing some breaking changes on minor releases - though we won't need that for the most used plugins out of the gate (but store, NFC rust api still needs some work)

FabianLars:
ok

tweidinger (dismissed):
🤞

PR #1919, opened 2024-10-11
Add global version_comparator

n1ght-hunter:
This pr allows you do set a `version_comparator` when building the plugin that will get applied to all subsequent calls. 
this does not change overriding if you call the builder manually but allows a way to alter version_comparator for js command calls


plugins/updater/src/updater.rs
```diff
@@ -221,6 +224,11 @@ impl UpdaterBuilder {
         self
     }
 
+    pub(crate) fn global_version_comparator(mut self, f: GlobalVersionComparator) -> Self {
+        self.global_version_comparator.replace(f);
+        self
+    }
```
amrbashir: this is the same as `version_comparator` method, no need to add a new one

plugins/updater/src/updater.rs
```diff
@@ -93,12 +93,14 @@ impl RemoteRelease {
...
 
 pub type OnBeforeExit = Arc<dyn Fn() + Send + Sync + 'static>;
+pub(crate) type GlobalVersionComparator = Arc<dyn Fn(Version, RemoteRelease) -> bool + Send + Sync>;
 
 pub struct UpdaterBuilder {
     app_name: String,
     current_version: Version,
     config: Config,
     version_comparator: Option<Box<dyn Fn(Version, RemoteRelease) -> bool + Send + Sync>>,
+    global_version_comparator: Option<GlobalVersionComparator>,
```
amrbashir: no need to add a new field either

plugins/updater/src/updater.rs
```diff
@@ -93,12 +93,14 @@ impl RemoteRelease {
 }
 
 pub type OnBeforeExit = Arc<dyn Fn() + Send + Sync + 'static>;
+pub(crate) type GlobalVersionComparator = Arc<dyn Fn(Version, RemoteRelease) -> bool + Send + Sync>;
```
amrbashir: and this type can also be removed

plugins/updater/src/lib.rs
```diff
@@ -88,6 +93,10 @@ impl<R: Runtime, T: Manager<R>> UpdaterExt<R> for T {
             builder = builder.current_exe_args(args);
         }
 
+        if let Some(version_comparator) = version_comparator {
+            builder = builder.global_version_comparator(version_comparator.clone());
```
amrbashir: and then we can just call `version_comparator` here instead of `global_version_comparator`

amrbashir (changes requested):

plugins/updater/src/updater.rs
n1ght-hunter: its not this one is in a Arc not a box. if you are fine with change the normal one to an Arc then alot of the boiler plate can be removed

amrbashir: why does it have to be `Arc`?

n1ght-hunter: because it has to exist in both the global object as well as the update object when build at the same time

amrbashir: I see, then feel free to change it to `Arc`

n1ght-hunter:
ok will fix it up this week

plugins/updater/src/lib.rs
n1ght-hunter: there is a slight issue with this as the types dont match.
since calling `version_comparator` has a param type of `F: Fn(Version, RemoteRelease) -> bool + Send + Sync + 'static` but the global one has `Arc<dyn Fn(Version, RemoteRelease) -> bool + Send + Sync>` so i have made the field `pub(crate)` but could also use another builder method only in the crate

amrbashir (changes requested):
@n1ght-hunter  This LGTM, can you also add a change file in `.changes` directory?

plugins/updater/src/lib.rs
```diff
@@ -163,9 +172,18 @@ impl Builder {
         self
     }
 
+    pub fn version_comparator<F: Fn(Version, RemoteRelease) -> bool + Send + Sync + 'static>(
```
n1ght-hunter: @amrbashir do you think that this name is good or should it be named `global_version_comparator` or `default_version_comparator`

amrbashir: `default_version_comparator` sounds good, let's rename to that

.changes/add-global-updater-version-comparator.md
```diff
@@ -0,0 +1,5 @@
+---
+'updater': 'minor'
+---
+
+Add a new method `version_comparator` on updater builder to set the default version comparator for the updater. This is useful for the js since it can't set a version comparator.
```
amrbashir: Apologies for the delay, just this simple change and I can merge


```suggestion
Add `tauri_plugin_updater::Builder::default_version_comparator` method to set the default version comparator for the updater.
```

amrbashir: just this simple change and I an merge

amrbashir (approved):

.changes/add-global-updater-version-comparator.md
```diff
@@ -0,0 +1,5 @@
+---
+'updater': 'minor'
```
amrbashir, line 2:
```suggestion
'updater': 'minor'
'updater-js': 'minor'
```

amrbashir: You also need to run `cargo fmt`

amrbashir (changes requested):

amrbashir (approved):

PR #1924, opened 2024-10-12
fix: autoSave Inconsistency

hylarucoder:
The Tauri Store plugin has a type mismatch for the autoSave option.  In TypeScript, it's documented as a boolean, but the Rust backend expects a u64. Setting autoSave to true or false in createStore causes an error:

```bash
Unhandled Promise Rejection: invalid args `autoSave` for command `create_store`: invalid type: boolean `true`, expected u64 
```


Legend-Master:
Thanks for the fix, but this is tracked in #1865 and will be included the next rework #1860

PR #1932, opened 2024-10-14
fix(sql): don't panic on missing migration when preload is enabled

### LINKED ISSUE #1930: SQL Plugin: The `preload` parameter has a potential `migration` requirement.
# Problem
When setting up the sql plugin, the ‘preload’ parameter is only used for migrations and assumes that the migration is given

https://github.com/tauri-apps/plugins-workspace/blob/7e1c17a635f10d69e10ea9f35081a68c71276203/plugins/sql/src/lib.rs#L145-L153

This does not seem to match its semantics.

If I use the SQL plugin and give a `preload` parameter but don't provide any migrations, the programme reports the following panic message
image

This is very unintuitive, and although `preload` is currently only used by `migration`, it would be better to report an error like `no migration providing`.

# Improvements

I have the following ideas to fix the current problem
1. rename `preload` to something else related to `migration`, such as `migration-preload`.
2. `preload` is not only used for migrations, but also for opening connections and managing them
3. provide a more user friendly panic message

Goodjooy:
this pr is the implement of #1930 .

# tasks
- [x] skip migration if no migration provided in preload

### DIFF
--- a/plugins/sql/README.md
+++ b/plugins/sql/README.md
@@ -168,7 +168,6 @@ fn main() {
 
 To apply the migrations when the plugin is initialized, add the connection string to the `tauri.conf.json` file:
 
-> Notice: If `migration` is not provided, make sure that `preload` is kept empty
 
 ```json
 {
--- a/plugins/sql/README.md
+++ b/plugins/sql/README.md
@@ -168,6 +168,8 @@ fn main() {
 
 To apply the migrations when the plugin is initialized, add the connection string to the `tauri.conf.json` file:
 
+> Notice: if no migration provided, keeping the `preload` empty
+
 ```json
 {
   "plugins": {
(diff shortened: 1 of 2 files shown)


FabianLars:
Thanks for the PR! I don't think we should panic here. I think i would have went with simply replacing the `unwrap()` that caused the panic with `unwrap_or_default()` so `preload` is basically ignored if no migrations were set.

Goodjooy:
> Thanks for the PR! I don't think we should panic here. I think i would have went with simply replacing the `unwrap()` that caused the panic with `unwrap_or_default()` so `preload` is basically ignored if no migrations were set.

This solution is really simpler and fits the semantics of the `preload` itself, the code change is done

Goodjooy:
Looks like CI is having trouble installing webkit2gtk. :(

FabianLars:
oh, github changed the default runners

FabianLars (approved):

PR #1934, opened 2024-10-14
fix(store): `autoSave` type to `number`

ayangweb:
rust receives a type of `u64` and ts passes a `boolean`, resulting in a type error!

https://github.com/tauri-apps/plugins-workspace/blob/7e1c17a635f10d69e10ea9f35081a68c71276203/plugins/store/src/lib.rs#L50

iShot_2024-10-14_21 29 07


Ludea:
there is already https://github.com/tauri-apps/plugins-workspace/pull/1860

PR #1941, opened 2024-10-15
feat: allow http calls without origin header

Nipsuli:
Example [Clerk](https://clerk.com/) auth provider does not work when both Authorization and Origin headers are set. This allows one to skip setting the origin header explicitly.

In my application I'm routing all requests via tauri-http by patching global fetch. And I'm using custom Clerk setup to get the token shared between js and rust. To do that I need to use Authorization header instead of cookies to communicate with Clerk. And Clerk fails if there is both Authorization header and Origin header, as such I need a way to send requests without origin header.

I do like the default behavior of setting the origin header if it's not set as that mimics the browser behavior. But I also need a way to not send it. This does not change any existing behaviors, and Origin header can anyways be:
```
Origin: null
Origin: <scheme>://<hostname>
Origin: <scheme>://<hostname>:<port>
```
So empty string would be invalid header value according to the spec. As such I was thinking that if client has sent empty origin header we could just omit the origin header.

### DIFF
--- a/.changes/http-allow-skip-origin.md
+++ b/.changes/http-allow-skip-origin.md
@@ -1,0 +1,6 @@
+---
+"http": "patch"
+"http-js": "patch"
+---
+
+Allow skipping sending Origin header in HTTP requests by sending empty string as the value for Origin header.
--- a/.changes/http-allow-skip-origin.md
+++ b/.changes/http-allow-skip-origin.md
@@ -3,4 +3,4 @@
 "http-js": "patch"
 ---
 
-Allow skipping sending Origin header in HTTP requests by sending empty string as the value for Origin header.
+Allow skipping sending `Origin` header in HTTP requests by setting `Origin` header to an empty string when calling `fetch`.
--- a/plugins/http/src/commands.rs
+++ b/plugins/http/src/commands.rs
@@ -264,6 +264,14 @@ pub async fn fetch<R: Runtime>(
                     }
                 }
 
+                // In case empty origin is passed, remove it. Some services do not like Origin header
+                // so this way we can remove it in explicit way. The default behaviour is still to set it
+                if cfg!(feature = "unsafe-headers")
+                    && headers.get(header::ORIGIN) == Some(&HeaderValue::from_static(""))
+                {
+                    headers.remove(header::ORIGIN);
+                };
+
                 if let Some(data) = data {
                     request = request.body(data);
                 }


plugins/http/src/commands.rs
```diff
@@ -264,6 +264,14 @@ pub async fn fetch<R: Runtime>(
                     }
                 }
 
+                // In case empty origin is passed, remove it. Some services do not like Origin header
+                // so this way we can remove it in explicit way. The default behaviour is still to set it
+                if cfg!(feature = "unsafe-headers")
+                    && headers.get(header::ORIGIN) == Some(&HeaderValue::from_static(""))
```
amrbashir, line 270: I don't think we should match on an empty string, because there might be some users who want to send empty string for `Origin`. I'd rather add a new option when making the request, for example:
```js
fetch(url, {
  sendOriginHeader: false
});
```

amrbashir (changes requested):
Thank you, Could you also add a change file in `.changes` directory?

plugins/http/src/commands.rs
Nipsuli: This is something I was thinking a lot. Like I'd like to avoid adding extra parameters to the fetch and try to keep the fetch as close to global fetch. And in case of patching global fetch with tauri-http (example to get control what fetches actually can be made) the consumers assume standard fetch. Different libraries support some pre/post hooks for different operations that allow example inject headers and such like for custom auth and they operate with the `RequestInit` object.

And the syntax of Origin header is:
```
Origin: null
Origin: <scheme>://<hostname>
Origin: <scheme>://<hostname>:<port>
```
[MDN origin header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin)
[RFC 6454](https://datatracker.ietf.org/doc/html/rfc6454#section-6.1)

Sending an empty string is invalid format. In case user wants to hide the origin for example privacy reasons they should set it to `null` string literal. 

For these reasons I ended up with sending empty string in the origin header as the mechanism.

I hope my reasoning makes sense.

amrbashir: I was comparing with how node and deno runtimes behave and both allow setting `Origin` to an empty string. However your reasoning is quite valid and I think we should go with your approach until someone else asks for this specific behavior.

.changes/http-allow-skip-origin.md
```diff
@@ -0,0 +1,6 @@
+---
+"http": "patch"
+"http-js": "patch"
+---
+
+Allow skipping sending Origin header in HTTP requests by sending empty string as the value for Origin header.
```
amrbashir: Just a slight wording change and this should be ready to go
```suggestion
Allow skipping sending `Origin` header in HTTP requests by setting `Origin` header to an empty string when calling `fetch`.
```

amrbashir (approved):

amrbashir:
Thank you

PR #1945, opened 2024-10-16
Remove cameraView from superview after scanning

d0m1n1kr:
Remove from superview when camera subview is dismantled. 

If the view is not removed, all inputs are captured by this (transparent) view.
After removing, all events are propagated to the WebView as expected.


FabianLars:
I completely forgot about this PR after requesting the review and the change was done in https://github.com/tauri-apps/plugins-workspace/pull/2440 instead - sorry!!

PR #1947, opened 2024-10-17
feat(shell) add wait fn to get event from child process

### LINKED ISSUE #1920: [shell] feat: close event on child process
it could be awesome if this plugin get an event on closed child.

https://doc.rust-lang.org/std/process/struct.Child.html#method.wait

Ludea:
closes https://github.com/tauri-apps/plugins-workspace/issues/1920


plugins/shell/src/commands.rs
```diff
@@ -302,6 +303,20 @@ pub fn kill<R: Runtime>(
     Ok(())
 }
 
+#[tauri::command]
+pub fn wait<R: Runtime>(
```
ahqsoftwares: It will block the whole app to wait for the process

plugins/shell/src/process/mod.rs
```diff
@@ -84,6 +84,12 @@ impl CommandChild {
     pub fn pid(&self) -> u32 {
         self.inner.id()
     }
+
+    /// Waits for the child to exit completely, returning the status that it exited with.
+    pub fn wait(self) -> crate::Result<std::process::ExitStatus> {
```
ahqsoftwares: It will block the whole app to wait for the process to exit, we need to use something async / non blocking

Ludea: I use async fn

plugins/shell/src/commands.rs
Ludea: I use async to avoid blocking thread

plugins/shell/src/process/mod.rs
ahqsoftwares: async fn will also block it

plugins/shell/src/commands.rs
```diff
@@ -302,6 +302,19 @@ pub fn kill<R: Runtime>(
     Ok(())
 }
 
+#[tauri::command]
```
ahqsoftwares, line 305: Use try_wait and if it's ready then wait for the response

plugins/shell/src/process/mod.rs
```diff
@@ -84,6 +84,14 @@ impl CommandChild {
     pub fn pid(&self) -> u32 {
         self.inner.id()
     }
+
+    /// Waits for the child to exit completely, returning the status that it exited with.
+    pub async fn wait(self) -> crate::Result<ExitStatus> {
```
ahqsoftwares, line 89: https://doc.rust-lang.org/std/process/struct.Child.html#method.try_wait

ahqsoftwares:
Also, you may use tokio::process https://docs.rs/tokio/latest/tokio/process/struct.Command.html instead of std::process

PR #1958, opened 2024-10-20
fix(fs): fix compilation on targets with pointer width 16 and 32

bWanShiTong:
Fixed compilation for systems where `target_pointer_width`  is 16 or 32


amrbashir:
Thank you

amrbashir (approved):

PR #1963, opened 2024-10-21
docs(sql): docs of frontend Database select method

yoggys:
This fixes the documentation on the frontend side with an example in the select method.


joao-salomao (approved):
LGTM

FabianLars (approved):

PR #1964, opened 2024-10-21
feat(fs): support `ReadableStream<Unit8Array>` for `writeFile` API

### LINKED ISSUE #1598: Use stream in tauri-plugin-fs `writeFile`
### Describe the problem

For now I’m using
```
const dataBuffer = await blob.arrayBuffer();
const data = new Uint8Array(dataBuffer);
await writeFile(filename, data)
```
to write file. 

Can we use `blob.stream` to get a readable stream and pass it to writeFile.  Since v2’s IPC is based on Response, we can use new Response() to generate a streaming response to improve performance and reduce memory usage. 

### Alternatives considered

Can Transferable object improve performance?

### Additional context

_No response_

amrbashir:
closes #1598


plugins/fs/permissions/autogenerated/commands/write_file.toml
```diff
@@ -5,9 +5,22 @@
...
 
 [[permission]]
 identifier = "deny-write-file"
 description = "Denies the write_file command without any pre-configured scope."
-commands.deny = ["write_file"]
+
+[permission.commands]
+allow = [
+    "open",
+    "write",
```
FabianLars: this is wrong i think?

plugins/fs/permissions/autogenerated/commands/read_text_file_lines.toml
```diff
@@ -5,9 +5,18 @@
...
+]
+deny = []
 
 [[permission]]
 identifier = "deny-read-text-file-lines"
 description = "Denies the read_text_file_lines command without any pre-configured scope."
-commands.deny = ["read_text_file_lines"]
+
+[permission.commands]
+allow = ["read_text_file_lines_next"]
```
FabianLars: same(?)

plugins/fs/permissions/autogenerated/commands/write_file.toml
amrbashir: nice catch, fixed

FabianLars:
tbh i'm still a bit confused why the allow and deny arrays don't match but i assume it's correct. Ah no, is it because users may want to deny write_file but still use open/write?

FabianLars (approved):

amrbashir:
> tbh i'm still a bit confused why the allow and deny arrays don't match but i assume it's correct. Ah no, is it because users may want to deny write_file but still use open/write?

yeah

tjramage:
Hey @amrbashir @FabianLars – just a quick heads-up that I don't think this actually works on macOS (15.3.1). I get a TypeError on the `for await (const chunk of data)` line: 

**TypeError: undefined is not a function (near '...chunk of data...')**

Async iterator support on Safari is not currently in place (https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream#browser_compatibility)

I had to drop in a polyfill for this to work:

```js
ReadableStream.prototype[Symbol.asyncIterator] = async function* () {
  const reader = this.getReader()
  try {
    while (true) {
      const {done, value} = await reader.read()
      if (done) return
      yield value
    }
  }
  finally {
    reader.releaseLock()
  }
}
```

Do you need me to create a separate issue? Happy to provide more info if needed.

amrbashir:
@tjramage thanks for the heads up, I forgot to check if it was supported on all platforms. Would you like to open a PR to change from async iterator to manual read, same as you did in the polyfill? or just open an issue if you like so I don't forget.

tjramage:
@amrbashir – no problem, leave the PR with me 👍

PR #1965, opened 2024-10-22
fix(fs): fix `writeTextFile` converting UTF-8 characters in path into replacement character

### LINKED ISSUE #1849: Tauri V2: writeFile can not handle UTF-8 characters in the filename
Hi, I migrated today my app to Tauri V2.

But now I'm facing an issue with the writeFile method of the “@tauri-apps/plugin-fs” plugin.
When I’m saving a file with this method, where the filename contains special German characters like “äöü”, then they are replaced with a “?”.

It looks like that the method does not support UTF-8 characters in the filename.

I tried it on Windows 11 on ARM and Windows 11 X64.

Here is a screenshot from the debugger.
image

Here is the result in the explorer:

image

Here are among others my dependencies in the package.json:

```JSON
"@tauri-apps/plugin-dialog": "^2.0.0-rc.1",
"@tauri-apps/plugin-fs": "^2.0.0-rc.2",
"@tauri-apps/plugin-os": "^2.0.0-rc.1",
```

Here are among others my dependencies in cargo.toml:

```
tauri-plugin-dialog = "2.0.0-rc"
tauri-plugin-fs = "2.0.0-rc"
tauri-plugin-os = "2.0.0-rc"
```

Can you please fix it?

Best regards
Frank

amrbashir:
closes #1849


Legend-Master:
~~I can't reproduce this problem with the current head on Windows 10~~ I can now, not sure what happened

image

Couldn't we just use `encodeURI` here?

amrbashir:
we were using `encodeURIcomponent` which is similar and didn't work

Legend-Master:
Both `encodeURIcomponent` and `encodeURI` worked for me here

image

```diff
- path: path instanceof URL ? path.toString() : path,
+ path: encodeURIComponent(path instanceof URL ? path.toString() : path),
```

Also I can't reproduce this bug with `writeFile`

amrbashir:
looks like I was testing `writeTextFile` but was looking at the code of `writeFile`, oh well all that time is wasted I guess

Legend-Master (approved):

PR #1972, opened 2024-10-25
fix(sql): replace Mutex with RwLock to enable concurrent SQL execution

### LINKED ISSUE #1935: [sql] select queries are blocked and not executed concurrently (sqlite)
Using this short code:

```js
console.time("long");
console.time("short");
// This takes 3 seconds
db.select(
  "WITH RECURSIVE cte(x) AS (SELECT 1 UNION ALL SELECT x + 1 FROM cte LIMIT 10000000) SELECT count(x) FROM cte;"
).then(() => {
  console.timeEnd("long");
});
db.select("SELECT 1;").then(() => {
  console.timeEnd("short");
});
```

I have to wait the end of the first query to finish its execution to get the second query result.
It's as if each query is waiting for the following one before running itself.

Here is the result:

```
long: 3580.368ms
short: 3580.434ms
```

As you can see both query took 3.5 seconds even if the second one is fast.

Could this be fixed?

cijiugechu:
When using a `Mutex`, the `MutexGuard` will be held until the SQL execution is complete before it is dropped, which is unnecessary and prevents the concurrency capability of the SQL.

Resolves #1935 .

### DIFF
--- a/.changes/fix-sql-blocking.md
+++ b/.changes/fix-sql-blocking.md
@@ -1,0 +1,5 @@
+---
+sql: patch
+---
+
+Replace `Mutex` with `RwLock` to enable concurrent SQL execution.
--- a/plugins/sql/src/commands.rs
+++ b/plugins/sql/src/commands.rs
@@ -23,7 +23,7 @@ pub(crate) async fn load<R: Runtime>(
         pool.migrate(&migrator).await?;
     }
 
-    db_instances.0.lock().await.insert(db.clone(), pool);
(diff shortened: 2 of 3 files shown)


FabianLars:
thank you! please also add a changefile similar to https://github.com/tauri-apps/plugins-workspace/blob/30bcf5dcc22e1bb1fb983a8d2887edc39404e6df/.changes/feat-multiple-sql-backends.md

FabianLars (approved):

rap2hpoutre:
Wow thank you for this fix!

PR #1981, opened 2024-10-29
chore: print a warning in development for forbidden headers

### LINKED ISSUE #1979: [plugin-http] Add a warning when using forbidden headers without the unsafe-headers feature
When sending a request using the `fetch` method provided by the http-client, [unsafe headers](https://github.com/tauri-apps/plugins-workspace/blob/4341d7f500290165a94e05006abbfe7f40045fc9/plugins/http/src/commands.rs#L386) are filtered out for security reason. While this is beneficial, it can be confusing when users need to use these headers, since the removal of unsafe headers result in no extra output. It took me a long time to find out why I cannot set the Cookies.

I propose adding [output statements here](https://github.com/tauri-apps/plugins-workspace/blob/4341d7f500290165a94e05006abbfe7f40045fc9/plugins/http/src/commands.rs#L194) in development mode to provide users with hints related to the usage of unsafe headers.

amrbashir:
closes #1979

### DIFF
--- a/plugins/http/src/commands.rs
+++ b/plugins/http/src/commands.rs
@@ -194,7 +194,7 @@ pub async fn fetch<R: Runtime>(
             #[cfg(debug_assertions)]
             {
                 eprintln!("[\x1b[33mWARNING\x1b[0m] Skipping {name} header as it is a forbidden header per fetch spec https://fetch.spec.whatwg.org/#terminology-headers");
-                eprintln!("[\x1b[33mWARNING\x1b[0m] if this is a desired behavior, you can enable `unsafe-headers` feature flag in your Cargo.toml");
+                eprintln!("[\x1b[33mWARNING\x1b[0m] if keeping the header is a desired behavior, you can enable `unsafe-headers` feature flag in your Cargo.toml");
             }
             continue;
         }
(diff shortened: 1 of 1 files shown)


plugins/http/src/commands.rs
```diff
@@ -191,6 +191,11 @@ pub async fn fetch<R: Runtime>(
         let name = HeaderName::from_str(&h)?;
         #[cfg(not(feature = "unsafe-headers"))]
         if is_unsafe_header(&name) {
+            #[cfg(debug_assertions)]
+            {
+                eprintln!("[\x1b[33mWARNING\x1b[0m] Skipping {name} header as it is a forbidden header per fetch spec https://fetch.spec.whatwg.org/#terminology-headers");
+                eprintln!("[\x1b[33mWARNING\x1b[0m] if this is a desired behavior, you can enable `unsafe-headers` feature flag in your Cargo.toml");
```
FabianLars: this sounds like `if skipping the header is a desired behavior`. what about "if keeping the header is desired" instead?

amrbashir:
```suggestion
                eprintln!("[\x1b[33mWARNING\x1b[0m] if keeping the header is a desired behavior, you can enable `unsafe-headers` feature flag in your Cargo.toml");
```

amrbashir: updated

FabianLars (approved):

PR #1982, opened 2024-10-29
feat(localhost): add custom host binding to allow external access

arihav:
This PR introduces the ability to specify a custom host binding for the server, allowing external access. Previously, the server was hardcoded to bind to `localhost`, limiting accessibility to only the local machine. With this change, users can configure the server to bind to any IP address or hostname, including `0.0.0.0`, which enables connections from external devices.

### DIFF
--- a/plugins/localhost/src/lib.rs
+++ b/plugins/localhost/src/lib.rs
@@ -59,8 +59,9 @@ impl Builder {
         }
     }
 
-    pub fn host<H: AsRef<str>>(mut self, host: H) -> Self {
-        self.host = Some(host.as_ref().to_string());
+    // Change the host the plugin binds to. Defaults to `localhost`.
+    pub fn host<H: Into<String>>(mut self, host: H) -> Self {
+        self.host = Some(host.into());
         self
     }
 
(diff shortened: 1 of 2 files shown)


plugins/localhost/src/lib.rs
```diff
@@ -46,17 +46,24 @@ type OnRequest = Option<Box<dyn Fn(&Request, &mut Response) + Send + Sync>>;
...
     pub fn new(port: u16) -> Self {
         Self {
             port,
+            host: None,
             on_request: None,
         }
     }
 
+    pub fn host<H: AsRef<str>>(mut self, host: H) -> Self {
+        self.host = Some(host.as_ref().to_string());
```
FabianLars:
```suggestion
    // Change the host the plugin binds to. Defaults to `localhost`.
    pub fn host<H: Into<String>>(mut self, host: H) -> Self {
        self.host = Some(host.into());
```

FabianLars: the diff looks a bit weird but the main change is AsRef -> Into (since we're turning it into a string anyway). Just added a short doc sentence too.

FabianLars (approved):

PR #1986, opened 2024-10-30
fix(clipboard-manager): `read_image` wrongly set the image rgba data with binary PNG data.

### LINKED ISSUE #1985: [bug] [clipboard-manager]  `read_image` wrongly set the image rgba data with binary PNG data.
## Minimal Reproduction

Simply initialize a tauri project, install `clipboard-manager` and modify `main.ts` like this
``` ts
import { readImage } from "@tauri-apps/plugin-clipboard-manager";

async function greet() {
  readImage().then(async (image) => {
    const rgba = await image.rgba();
    let s = ''
    for (let b of rgba) {
      s += String.fromCharCode(b)
    }
    alert(s)
  });
}

greet();
```

Run `npm run tauri dev` with an image in your clipboard, you will find the decoded rgba data shows 

image

Obviously it is a binary PNG data, however according to [the references](https://tauri.app/reference/javascript/api/namespaceimage/#rgba), the return value should be rgba data in pixel.

## Fix
In the following code, the rgba data got by `clipboard.get_image()` is encoded using `image` crate, which is not necessary.

https://github.com/tauri-apps/plugins-workspace/blob/0ca4cc914c5ea995c98f9e60a2ab49827c219350/plugins/clipboard-manager/src/desktop.rs#L84-L103

I will open a pr for fixing this.

RikaKagurasaka:
fix #1985 

The dependency of `image` crate may be unnecessary now. I'm not sure whether it should be removed.

### DIFF
--- a/.changes/fix-1985.md
+++ b/.changes/fix-1985.md
@@ -1,0 +1,5 @@
+---
+"clipboard-manager": patch
+---
+
+Fix that `read_image` wrongly set the image rgba data with binary PNG data.
--- a/plugins/clipboard-manager/Cargo.toml
+++ b/plugins/clipboard-manager/Cargo.toml
@@ -37,4 +37,3 @@ tauri = { workspace = true, features = ["wry"] }
 
 [target."cfg(any(target_os = \"macos\", windows, target_os = \"linux\", target_os = \"dragonfly\", target_os = \"freebsd\", target_os = \"openbsd\", target_os = \"netbsd\"))".dependencies]
 arboard = "3"
-image = "0.25"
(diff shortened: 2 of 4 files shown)


amrbashir:
> The dependency of `image` crate may be unnecessary now. I'm not sure whether it should be removed.

yeah looks like it, feel free to remove it

amrbashir (changes requested):
Could you add a change file in `.changes` directory? also you need to run `cargo fmt`

RikaKagurasaka:
> Could you add a change file in `.changes` directory? also you need to run `cargo fmt`

OK! I think I finished. Could you kindly approve another check workflow?

amrbashir (approved):

amrbashir:
thank you

PR #1991, opened 2024-11-01
feat(updater): Add .deb Package Support to Linux Updater

jLynx:
## Description

This PR extends the Linux updater functionality to support .deb package updates alongside the existing AppImage support. The implementation includes multiple authentication methods for package installation, providing a user-friendly update experience for Debian-based distributions.

_Note: This PR relies on https://github.com/tauri-apps/tauri/pull/11562_

### Key Features

- Support for both .deb packages and AppImage updates
- Multiple privilege escalation methods:

1. `pkexec` (primary, graphical)
2. `zenity` for GUI password prompt
3. Terminal-based `sudo` as fallback

- Automatic detection of installation type (AppImage vs .deb)
- Secure handling of authentication and package installation

### Technical Details

- Added new error types for authentication and installation failures
- Implemented .deb package extraction from both direct files and tar.gz archives
- Added helper methods for privilege escalation and authentication
- Maintained backward compatibility with existing AppImage functionality

### Expected Archive Structures

```javascript
├── [AppName]_[version]_amd64.AppImage.tar.gz
│   └──[AppName]_[version]_amd64.AppImage
├── [AppName]_[version]_amd64.deb
└── ...
```

## Testing

- Tested on Debian-based distributions
- Verified both .deb and AppImage update paths
- Tested various authentication scenarios

## Notes

- Requires `zenity` for graphical password prompt (optional)
- Falls back to terminal-based authentication if graphical methods fail
- Maintains security by properly handling sensitive authentication data


jLynx:
Removed support for .deb.tar.gz as requested by @FabianLars https://discord.com/channels/616186924390023171/1301285233991356506/1302596935101976616

plugins/updater/src/updater.rs
```diff
@@ -835,6 +842,108 @@ impl Update {
...
     }
+
+    fn is_deb_package(&self) -> bool {
+        // Check if we're running from a .deb installation
+        // Typically installed in /usr/bin or /usr/local/bin
+        self.extract_path
+            .to_str()
+            .map(|p| p.starts_with("/usr"))
+            .unwrap_or(false)
+    }
```
amrbashir: how does that handle `rpm` or if installed within Arch Linux (through tools that convert .deb, or in the future when we add pacman support to tauri-bundler)?

amrbashir: I am also not sure why we would want to add support for `.deb` in the updater when the system itself has a package manager that can do the updates.

I think I prefer to notify the user there is an update and let them update through apt, or whatever other package manager for `.rpm`

jLynx: @amrbashir Thank you for the feedback! Let me address your points:

1. Regarding RPM and Arch Linux packages: You're right that this implementation is currently .deb specific. The intention was to start with .deb support as a first step as for our use case this is what we currently need.
2. About using system package managers: I understand your perspective about letting the system package manager handle updates. However, there are some use cases where direct .deb installation might be preferred:

- When the app isn't distributed through official repositories
- For organizations that want to control their own update distribution
- For testing/beta versions that aren't in package repositories

At the end of the day, the developer can choose to use this update mechanism or the package manager, it's just giving them another option. This implementation now brings Linux .deb packages in sync with other operating systems that support in-app updates. (Mac & Windows)

We personally couldn't use AppImages (which already support updates) in our project because they don't support deeplinks or icons, which .deb packages handle natively. This makes .deb packages a better choice for applications that need these features.

amrbashir: cool, I am convinced we should add support for .deb and potentially other formats like RPM and future pacman, but we need to fix the implementation to be 100% it is running from `.deb`

jLynx: Great news, Do you have any suggestions @amrbashir for how we could detect that?

jLynx: Would you be ok with this?
Consider it a deb package only if:
1. We're in a system path AND
2. We have Debian package management tools AND
3. The binary is tracked by dpkg

amrbashir: yeah that sounds good

jLynx: Have now updated that detection method @amrbashir

jLynx: Just fixed the linting, sorry about that @amrbashir

plugins/updater/src/updater.rs
```diff
@@ -760,12 +760,19 @@ impl Update {
...
     fn install_inner(&self, bytes: &[u8]) -> Result<()> {
+        if self.is_deb_package() {
+            self.install_deb_update(bytes)
+        } else {
+            // Handle AppImage or other formats
+            self.install_appimage_update(bytes)
+        }
+    }
+
+    fn install_appimage_update(&self, bytes: &[u8]) -> Result<()> {
```
```suggestion
    fn install_appimage(&self, bytes: &[u8]) -> Result<()> {
```

plugins/updater/src/updater.rs
```diff
@@ -835,6 +842,132 @@ impl Update {
...
+        };
+
+        // Consider it a deb package only if:
+        // 1. We're in a system path AND
+        // 2. We have Debian package management tools AND
+        // 3. The binary is tracked by dpkg
+        dpkg_exists && apt_exists && package_in_dpkg
+    }
+
+    fn install_deb_update(&self, bytes: &[u8]) -> Result<()> {
```
```suggestion
    fn install_deb(&self, bytes: &[u8]) -> Result<()> {
```

plugins/updater/src/updater.rs
```diff
@@ -760,12 +760,19 @@ impl Update {
...
     /// ├── [AppName]_[version]_amd64.AppImage.tar.gz    # GZ generated by tauri-bundler
     /// │   └──[AppName]_[version]_amd64.AppImage        # Application AppImage
+    /// ├── [AppName]_[version]_amd64.deb                # Debian package
     /// └── ...
     ///
-    /// We should have an AppImage already installed to be able to copy and install
-    /// the extract_path is the current AppImage path
-    /// tmp_dir is where our new AppImage is found
     fn install_inner(&self, bytes: &[u8]) -> Result<()> {
+        if self.is_deb_package() {
```
amrbashir, line 767: we also need to check if `bytes` are actually for `.deb` see https://docs.rs/infer/latest/infer/archive/fn.is_deb.html

plugins/updater/src/updater.rs
```diff
@@ -835,6 +842,132 @@ impl Update {
...
+        // 2. We have Debian package management tools AND
+        // 3. The binary is tracked by dpkg
+        dpkg_exists && apt_exists && package_in_dpkg
+    }
+
+    fn install_deb_update(&self, bytes: &[u8]) -> Result<()> {
+        // Create a temporary directory
+        let tmp_dir = tempfile::Builder::new()
+            .prefix("tauri_deb_update")
+            .tempdir_in("/tmp")?;
```
amrbashir: when doing app images, we actually try in 3 different temp locations, see L779-L783, we should do the same here

plugins/updater/src/updater.rs
```diff
@@ -835,6 +842,132 @@ impl Update {
...
+        let deb_path = tmp_dir.path().join("package.deb");
+
+        // Direct .deb file
+        std::fs::write(&deb_path, bytes)?;
+
+        // Try different privilege escalation methods
+        let installation_result = self.try_install_with_privileges(&deb_path);
+
+        // Clean up
+        let _ = std::fs::remove_file(&deb_path);
```
amrbashir: this will be done automatically by tempfile crate, no?

plugins/updater/src/updater.rs
```diff
@@ -835,6 +842,132 @@ impl Update {
...
+                return Ok(());
+            }
+        }
+
+        // 3. Final fallback: terminal sudo
+        let status = std::process::Command::new("sudo")
+            .arg("dpkg")
+            .arg("-i")
+            .arg(deb_path)
+            .status()?;
```
amrbashir, line 940: won't `sudo` block and wait to type a password? which won't work for a GUI app since they are not necessarily started from a terminal?

amrbashir (changes requested):

amrbashir (commented):
you also need to add a change file in .changes directory

plugins/updater/src/updater.rs
jLynx: It still prompts for password, here is a screenshot of it in progress. This was opened from the app selector on Ubuntu
image

amrbashir: I see, but I remember some Desktop Enivornments might not have this behavior, anyways, let's ship and see really world usage.

plugins/updater/src/updater.rs
```diff
@@ -835,6 +842,132 @@ impl Update {
...
+
+        if status.success() {
+            Ok(())
+        } else {
+            Err(Error::DebInstallFailed)
+        }
+    }
+
+    fn get_password_graphically(&self) -> Result<String> {
+        let output = std::process::Command::new("zenity")
```
amrbashir: what if `zenity` doesn't exist, I am not sure if all desktop environments have similar things but I think KDE has `kdialog`

_I am starting to hate linux as a developer because of all these fragmentation_

amrbashir (changes requested):

plugins/updater/src/updater.rs
jLynx: If it doesnt exist, then it falls back to the other options, so having zenity installed it 100% optional on the users side and not required by any means

jLynx:
Will jump on your requested changes tomorrow morning as its EOD for me (NZ Timezone)

plugins/updater/src/updater.rs
amrbashir: then should we add support for `kdialog` as well, maybe as another fallback for KDE if `zenity` doesn't exist?

plugins/updater/src/updater.rs
jLynx: Do you want to use the same locations as appimage, or should I use 
`let tmp_dir_locations = ["/tmp", "/var/tmp", "/dev/shm"];`

plugins/updater/src/updater.rs
jLynx: very true, I'll remove this line

plugins/updater/src/updater.rs
jLynx: Updated. Will let you resolve if you are happy

plugins/updater/src/updater.rs
jLynx: Updated. Will let you resolve if you are happy

plugins/updater/src/updater.rs
jLynx: I have pushed the change with this, let me know if you want me to change it

plugins/updater/src/updater.rs
jLynx: Updated. Will let you resolve if you are happy

plugins/updater/src/updater.rs
jLynx: Added support for kdialog as well

plugins/updater/src/updater.rs
jLynx: Good call, have now added this in

jLynx:
Ok, I think I have addressed all the comments here 😄

plugins/updater/src/updater.rs
```diff
@@ -835,6 +842,156 @@ impl Update {
...
+
+        let deb_path = tmp_dir.path().join("package.deb");
+
+        // Direct .deb file
+        std::fs::write(&deb_path, bytes)?;
+
+        // Try different privilege escalation methods
+        let installation_result = self.try_install_with_privileges(&deb_path);
+
+        installation_result
```
```suggestion
        self.try_install_with_privileges(&deb_path)
```

plugins/updater/src/updater.rs
amrbashir: I would prefer to use the same implementation as appimage, just because it is battle tested

.changes/deb-update-support.md
```diff
@@ -0,0 +1,5 @@
+---
+"updater": "minor"
+---
+
+Added support for .deb package updates on Linux systems, including new error types (TempDirNotFound, AuthenticationFailed, DebInstallFailed) and implementation of .deb package installation with privilege escalation handling.
```
amrbashir:
```suggestion
Added support for `.deb` package updates on Linux systems.
```

plugins/updater/src/updater.rs
jLynx: Updated

.changes/deb-update-support.md
jLynx: Updated

plugins/updater/src/updater.rs
```diff
@@ -835,6 +842,162 @@ impl Update {
...
+                } else {
+                    None
+                }
+            })
+            .ok_or_else(|| Error::TempDirNotFound)?;
+
+        let deb_path = tmp_dir.path().join("package.deb");
+
+        // Direct .deb file
+        std::fs::write(&deb_path, bytes)?;
```
amrbashir: `std::fs::write` could fail and in which case we want to try the next temp dir location, let's change this to a for loop, similar to appimage implementaion

amrbashir (commented):
I almost forgot, we have a test for each updater format in https://github.com/tauri-apps/plugins-workspace/blob/v2/plugins/updater/tests/app-updater/tests/update.rs could you also include .deb in that test as well?

plugins/updater/src/updater.rs
jLynx: Ok fair point. Have updated it to be inline with the appimage implementation

jLynx:
@amrbashir sent you a DM on discord, just around the testing stuff

plugins/updater/tests/app-updater/tests/update.rs
```diff
@@ -105,20 +105,39 @@ impl Default for BundleTarget {
...
-        return Self::AppImage;
+        {
+            // Check if we're on a Debian-based system
+            if std::path::Path::new("/var/lib/dpkg").exists()
+                && std::path::Path::new("/etc/apt").exists()
+            {
+                return Self::Deb;
+            }
+            return Self::AppImage;
+        }
```
amrbashir: you can keep this one as is it was before

plugins/updater/tests/app-updater/tests/update.rs
```diff
@@ -105,20 +105,39 @@ impl Default for BundleTarget {
...
+            )),
+        )]
+    } else {
+        vec![(
+            BundleTarget::AppImage,
+            root_dir.join(format!(
+                "target/debug/bundle/appimage/app-updater_{version}_amd64.AppImage"
+            )),
+        )]
+    }
```
amrbashir: we need to test both formats, not one or the other

jLynx: Updated

plugins/updater/tests/app-updater/tests/update.rs
jLynx: Updated

.github/workflows/integration-tests.yml
```diff
@@ -34,8 +34,13 @@ jobs:
         with:
           fetch-depth: 0
 
+      # - name: install stable
```
amrbashir: why this was removed and now manually installing?

plugins/updater/tests/app-updater/tests/update.rs
```diff
@@ -105,20 +105,29 @@ impl Default for BundleTarget {
         #[cfg(any(target_os = "macos", target_os = "ios"))]
         return Self::App;
         #[cfg(target_os = "linux")]
-        return Self::AppImage;
+        return Self::Deb;
```
amrbashir: default in this test should stay AppImage

.github/workflows/integration-tests.yml
jLynx: This was for testing the pipeline as it doesnt work with sudo as it needs to be installed with sudo to work. will DM you

amrbashir:
@jLynx let's revert the updater test changes, I will merge the PR and add the test later in #2060 as I can't figure how to get MSI test to pass yet and I don't want to hold this PR for that.

jLynx:
ok cool, I will jump onto this in the next couple of hours

jLynx:
Removed the tests, so should be good to go @amrbashir

jLynx:
Fixed the formatting @amrbashir

amrbashir (approved):

PR #2008, opened 2024-11-05
fix(deep-link): Remove getCurrent call in onOpenUrl

### LINKED ISSUE #1879: [deep-link][v2] onOpenUrl Listener is being triggered when mounting even if the even was already processed
Hey there, I am trying to setup the `onOpenUrl` listener on my Tauri app, which is using React and React Router on the FE.

I have setup a single hook to hear on the frontend for the deep links event and generate a redirect based on the information. 
What I see is that when the redirect is triggered, a new listener will be mounted and I will get the deep link event again. 

I have already checked that the issue is not the double mounting React does on dev builds.

Is this a bug? Or I am missing something? 

Including a small snippet of the hook:

```
import { onOpenUrl } from "@tauri-apps/plugin-deep-link";
import { useEffect, useRef } from "react";
import { useNavigate } from "react-router-dom";

export const UseDeepLinkUrl = () => {
  const navigate = useNavigate();
  const isListenerRegistered = useRef(false);

  useEffect(() => {
    if (isListenerRegistered.current) {
      return;
    }

    const unlisten = onOpenUrl((url: string[]) => {
      console.log("url", url);
      const path: string = url[0].split("://")[1];
      console.log("Setting link to:", path);
      navigate(path);
    });
    isListenerRegistered.current = true;

    return () => {
      unlisten.then((f) => {
        console.log("unlisten", f);
        isListenerRegistered.current = false;
        return f();
      });
    };
  }, []);

  return null;
};
export default UseDeepLinkUrl;

```

FabianLars:
I received multiple reports that this behavior causes issues, like when reloading pages or re-mounting components. It's also not the behavior people seem to expect. Also, the function docs actually told users to use getCurrent themselves.

I'm not 100% sure about the rust side. The issues you see on the rust side aren't as apparent and the docs are also correct there but i think it would make sense to also change it.

### DIFF
--- a/.changes/fix-deep-link-onopenurl-current.md
+++ b/.changes/fix-deep-link-onopenurl-current.md
@@ -1,0 +1,5 @@
+---
+deep-link-js: patch
+---
+
+`onOpenUrl()` will now not call `getCurrent()` anymore, matching the documented behavior.
--- a/.changes/fix-deep-link-onopenurl-current.md
+++ b/.changes/fix-deep-link-onopenurl-current.md
@@ -1,4 +1,5 @@
 ---
+deep-link: patch
 deep-link-js: patch
 ---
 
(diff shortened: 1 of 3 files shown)


FabianLars:
i did the same for the rust side, can revert if desired

lucasfernog (approved):
as long as the examples show that you must check current before listening for events that's ok

PR #2019, opened 2024-11-07
feat: add `tauri-plugin-opener`

### LINKED ISSUE #999: [shell] Show in finder/explorer
### Describe the problem

The `shell.open` API can be used to open local files (in their respective app) or opening a directory in the finder/explorer-equivalent. In addition to this, I need to "show in finder", i.e. open a directory with a set of files/directories selected.

At a minimum, a function that accepts one file or directory and opens its containing directory with that file/dir selected. Ideally, multiple files can be provided, but I think this may be harder to do cross-platform.

### Alternatives considered

_No response_

### Additional context

If API additions are difficult, I'm also interested in workarounds.

### LINKED ISSUE #1045: [shell] shell.open scope should accept regex and fs scope.
Since it's very common to use shell.open with file paths we should make it easier to define scopes for paths. Regex is really hard to use for paths, especially if you want to use urls _and_ paths.

amrbashir:
closes #999
closes #1045


amrbashir:
marking as a draft until I release an empty packages on crates.io and npm

plugins/opener/Cargo.toml
```diff
@@ -0,0 +1,64 @@
...
+regex = "1"
+
+[dependencies]
+serde = { workspace = true }
+serde_json = { workspace = true }
+tauri = { workspace = true }
+thiserror = { workspace = true }
+open = { version = "5", features = ["shellexecute-on-windows"] }
+urlpattern = { workspace = true }
+regex = "1"
```
Legend-Master: Personally I would love to see it not to use regex, at least making this optional (regex is around 1MB in size, don't really like the entire app being around 3MB while adding this simple plugin alone bumps it to 4MB)

amrbashir: I am not sure if there is a better alternative, same dependencies are used by http plugin but I definitely want to avoid using it here.

amrbashir: decided to go with `glob`, we already depend on it in tauri and is light weight enough while providing a simple syntax for matching against ulrs for secuirty.

plugins/opener/guest-js/init.ts
```diff
@@ -0,0 +1,40 @@
...
+}
+
+if (
+  document.readyState === 'complete' ||
+  document.readyState === 'interactive'
+) {
+  openLinks()
+} else {
+  window.addEventListener('DOMContentLoaded', openLinks, true)
+}
```
Legend-Master: What about we just listen `click` events on `window`?

Legend-Master: Also, about finding if an anchor tag is clicked, what about we use something like this https://github.com/solidjs/solid-router/blob/main/src/data/events.ts

amrbashir: > What about we just listen `click` events on `window`?

let's keep it as is for now, I don't know if this was done on purpose or not and I don't want to break anything 

> Also, about finding if an anchor tag is clicked, what about we use something like this [solidjs/solid-router@`main`/src/data/events.ts](https://github.com/solidjs/solid-router/blob/main/src/data/events.ts?rgh-link-date=2024-11-12T02%3A39%3A26Z)

sure, but is our implementation faulty?

Legend-Master: > I don't know if this was done on purpose or not

I think it's probably because it was listening on body which requires the dom to be loaded first?

> sure, but is our implementation faulty?

I don't think this works well with custom elements? We detect the target and its parents but in custom elements, the event target is the custom element instead of the inner anchor tag

plugins/opener/Cargo.toml
```diff
@@ -0,0 +1,65 @@
...
+thiserror = { workspace = true }
+open = { version = "5", features = ["shellexecute-on-windows"] }
+url = { workspace = true }
+glob = { workspace = true }
+
+[target."cfg(windows)".dependencies]
+dunce = { workspace = true }
+
+[target."cfg(windows)".dependencies.windows]
+version = "0.54"
```
Legend-Master: We need to update this

Legend-Master (changes requested):

plugins/opener/guest-js/index.ts
```diff
@@ -0,0 +1,73 @@
...
+ * await open('/path/to/file');
+ * ```
+ *
+ * @param path The path or URL to open.
+ * @param openWith The app to open the file or URL with.
+ * Defaults to the system default application for the specified path type.
+ *
+ * @since 2.0.0
+ */
+export async function open(path: string, openWith?: string): Promise<void> {
```
Legend-Master: Let's give `openWith` a type here if it can only be those types: `'firefox' | 'google chrome' | 'chromium' | 'safari' | 'open' | 'start' | 'xdg-open' | 'gio' | 'gnome-open' | 'kde-open' | 'wslview'`

Legend-Master (changes requested):

plugins/opener/src/scope.rs
```diff
@@ -0,0 +1,122 @@
...
+        Self {
+            manager,
+            allowed,
+            denied,
+            _marker: PhantomData,
+        }
+    }
+
+    pub fn is_allowed(&self, path_or_url: &str) -> crate::Result<bool> {
+        let url = Url::parse(path_or_url).ok();
```
Legend-Master: This will not work, Windows paths like `C:/Users/Tony/.gitconfig` will be parsed as an URL

Legend-Master (changes requested):

plugins/opener/src/scope.rs
amrbashir: looks like we need to separate the two commands then

examples/api/src/views/Opener.svelte
```diff
@@ -0,0 +1,50 @@
+<script>
+  import { open } from '@tauri-apps/plugin-opener'
```
Legend-Master: Need to update this, also should we also include an example of `revealItemInDir`?

plugins/opener/guest-js/init.ts
```diff
@@ -0,0 +1,43 @@
...
+
+// open <a href="..."> links with the API
+window.addEventListener('click', function (evt) {
+  if (
+    evt.defaultPrevented ||
+    evt.button !== 0 ||
+    evt.metaKey ||
+    evt.altKey ||
+    evt.ctrlKey ||
+    evt.shiftKey
```
Legend-Master: The solid router's check is so the same page navigation gets caught, we need the opposite, ctrl and shift both open the link in another page, we should check `!(evt.ctrlKey || evt.shiftKey || a.target === '_blank')` instead

plugins/opener/guest-js/init.ts
```diff
@@ -0,0 +1,43 @@
+// Copyright 2019-2023 Tauri Programme within The Commons Conservancy
```
Legend-Master, line 1: Do you think this init script should be optional? I think we need a config option to enable/disable it

plugins/opener/src/reveal_item_in_dir.rs
```diff
@@ -0,0 +1,189 @@
...
+
+        let dir = HSTRING::from(dir);
+        let dir_item = unsafe { ILCreateFromPathW(PCWSTR::from_raw(dir.as_ptr())) };
+
+        let file_h = HSTRING::from(file);
+        let file_item = unsafe { ILCreateFromPathW(PCWSTR::from_raw(file_h.as_ptr())) };
+
+        unsafe {
+            if let Err(e) = SHOpenFolderAndSelectItems(dir_item, Some(&[file_item]), 0) {
+                if e.code().0 == ERROR_FILE_NOT_FOUND.0 as i32 {
```
Legend-Master: We can simplify this

```suggestion
        let dir_item = unsafe { ILCreateFromPathW(&HSTRING::from(dir)) };

        let file_h = HSTRING::from(file);
        let file_item = unsafe { ILCreateFromPathW(&file_h) };

        unsafe {
            if let Err(e) = SHOpenFolderAndSelectItems(dir_item, Some(&[file_item]), 0) {
                if e == ERROR_FILE_NOT_FOUND.into() {
```

Also it there any reason to fallback to `ShellExecute` on `ERROR_FILE_NOT_FOUND`? If we do need to fallback to it, we can't do `?` on `ShellExecuteExW` to early return here, or we'll leak `dir_item` and `file_item`

Legend-Master (changes requested):

plugins/opener/guest-js/init.ts
amrbashir: let's add that option in another release if requested

plugins/opener/guest-js/init.ts
amrbashir: good catch, pushed fix

examples/api/src/views/Opener.svelte
amrbashir: updated

plugins/opener/guest-js/init.ts
Legend-Master: Also we need to consider middle mouse button

plugins/opener/guest-js/init.ts
Legend-Master: Ok

plugins/opener/src/reveal_item_in_dir.rs
amrbashir: fixed the leak and simplified the code

plugins/opener/permissions/autogenerated/commands/reveal_in_dir.toml
```diff
@@ -0,0 +1,13 @@
+# Automatically generated - DO NOT EDIT!
```
Legend-Master: Is this a left over file?

plugins/opener/guest-js/init.ts
```diff
@@ -0,0 +1,41 @@
...
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-License-Identifier: MIT
+
+import { invoke } from '@tauri-apps/api/core'
+
+// open <a href="..."> links with the API
+window.addEventListener('click', function (evt) {
+  if (evt.defaultPrevented || evt.button !== 0 || evt.metaKey || evt.altKey)
```
Legend-Master: We will probably need to consider middle mouse button as well

I'm thinking is it possible we handle this from a [new window handler](https://docs.rs/wry/latest/wry/struct.WebViewBuilder.html#method.with_new_window_req_handler) instead of doing it in js?

examples/api/src/views/Opener.svelte
```diff
@@ -0,0 +1,99 @@
...
+
+  function openUrl() {
+    opener
+      .openUrl(url, urlProgram === 'Default' ? undefined : urlProgram)
+      .catch(onMessage)
+  }
+
+  function openPath() {
+    opener
+      .openPath(path, pathProgram === 'Default' ? undefined : urlProgram)
```
Legend-Master:
```suggestion
      .openPath(path, pathProgram === 'Default' ? undefined : pathProgram)
```

Legend-Master (changes requested):

plugins/opener/src/open.rs
```diff
@@ -0,0 +1,166 @@
...
+
+use std::{ffi::OsStr, fmt::Display, path::Path, str::FromStr};
+
+/// Program to use on the [`open()`] call.
+#[derive(Debug)]
+pub enum Program {
+    /// Use the `open` program.
+    Open,
+    /// Use the `start` program.
+    Start,
```
Legend-Master: I can't remember what do `open` and `start` actually do? They don't seem to work on Windows at least

plugins/opener/src/open.rs
```diff
@@ -0,0 +1,166 @@
...
+
+            #[cfg(target_os = "macos")]
+            Self::Firefox => "Firefox",
+            #[cfg(not(target_os = "macos"))]
+            Self::Firefox => "firefox",
+
+            #[cfg(target_os = "macos")]
+            Self::Chrome => "Google Chrome",
+            #[cfg(not(target_os = "macos"))]
+            Self::Chrome => "google-chrome",
```
Legend-Master: This doesn't seem to work on Windows, it should be `chrome`

Legend-Master (changes requested):

Legend-Master:
Also a small thing, when opening folders with the current shell execute open-rs implementation, the left panel of the file explorer will expand the folder to the current location, and I find it being quite annoying, the old implementation using `open` as operation (it's `explorer` now) doesn't have this problem

After seeing you use `SHOpenFolderAndSelectItems` I found that passing in the folder like this `SHOpenFolderAndSelectItems(folder, Some(&[folder]), 0)` will open that folder without expanding the left panel, also I don't think this will cause the explorer to freeze, do you think we can bring this upstream?

plugins/opener/permissions/autogenerated/commands/reveal_in_dir.toml
amrbashir: yeah, nice catch

plugins/opener/guest-js/init.ts
amrbashir: new window handler is not supported on Android iirc, will just add middle mouse support

amrbashir: looks like we may not be able to add middle mouse support atm, webview2 opens a new window and doesn't let our handler work, if we manage to add `new_window_handler` on Android, we can refactor this later.

plugins/opener/src/open.rs
amrbashir: I don't know tbh, I think I will change to remove this enum altogether.

plugins/opener/src/open.rs
amrbashir: I think I will change to remove this enum altogether.

amrbashir:
> Also a small thing, when opening folders with the current shell execute open-rs implementation, the left panel of the file explorer will expand the folder to the current location, and I find it being quite annoying, the old implementation using `open` as operation (it's `explorer` now) doesn't have this problem
> After seeing you use `SHOpenFolderAndSelectItems` I found that passing in the folder like this `SHOpenFolderAndSelectItems(folder, Some(&[folder]), 0)` will open that folder without expanding the left panel, also I don't think this will cause the explorer to freeze, do you think we can bring this upstream?

maybe, I don't have time to do it atm.

plugins/opener/src/open.rs
Legend-Master: This enum seems to support some commonly used programs cross-platforms, I agree we remove it for now since obviously it doesn't do what it supposed to do right now, we can bring it back in the future when we actually made it work

plugins/opener/guest-js/init.ts
Legend-Master: Yeah sure, as long as we can do it in a non breaking way I'm fine with it

Legend-Master:
> maybe, I don't have time to do it atm.

Don't worry, I can work on this, I'll open an [issue](https://github.com/Byron/open-rs/issues/103) to ask the owner first

plugins/opener/guest-js/init.ts
Legend-Master: About middle click, you should be able to use [`onauxclick`](https://developer.mozilla.org/en-US/docs/Web/API/Element/auxclick_event)?

Legend-Master:
From testing, when I click on the links in the example, the link gets opened twice, but from #2051, it seems like it's not a problem from this PR?

amrbashir:
> From testing, when I click on the links in the example, the link gets opened twice, but from #2051, it seems like it's not a problem from this PR?

yeah, this was a bug I noticed with old loop implementation, it is fixed in this PR and I will open a PR to fix #2051 later

plugins/opener/guest-js/index.ts
```diff
@@ -0,0 +1,112 @@
...
+ * On the configuration object, `open: true` means that the {@link open} API can be used with any URL,
+ * as the argument is validated with the `^((mailto:\w+)|(tel:\w+)|(https?://\w+)).+` regex.
+ * You can change that regex by changing the boolean value to a string, e.g. `open: ^https://github.com/`.
+ *
+ * @module
+ */
+
+import { invoke } from '@tauri-apps/api/core'
+
+export type Program =
```
Legend-Master: Need to update these as well

plugins/opener/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+
+pub use error::Error;
+type Result<T> = std::result::Result<T, Error>;
+
+pub use open::{open_path, open_url};
+pub use reveal_item_in_dir::reveal_item_in_dir;
+
+pub struct Opener<R: Runtime> {
+    #[allow(dead_code)]
+    app: AppHandle<R>,
```
Legend-Master: Do we need this?

Legend-Master (changes requested):
> it is fixed in this PR

It still seem to happen in this PR to me

plugins/opener/guest-js/init.ts
amrbashir: not supported on Safari, and so not supported on macOS wkwebview

plugins/opener/src/lib.rs
amrbashir: not really, can be removed

amrbashir:
> > it is fixed in this PR
> 
> It still seem to happen in this PR to me

weird, I can reproduce in this PR, anyways I am looking into it, seems not related to the implementation, might be a bug in tauri, still searching...

plugins/opener/guest-js/init.ts
Legend-Master: Right, we can roll with this for now

Legend-Master:
Seems like there're some problems about the permissions?

image

amrbashir:
fixed

plugins/opener/src/scope_entry.rs
```diff
@@ -0,0 +1,35 @@
+// Copyright 2019-2023 Tauri Programme within The Commons Conservancy
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-License-Identifier: MIT
+
+use std::path::PathBuf;
+
+use serde::Deserialize;
+
+#[derive(Deserialize, Debug)]
+pub enum Application {
```
Legend-Master, line 11: This needs to be untagged or the scope can't be parsed

```suggestion
#[serde(untagged)]
pub enum Application {
```

Legend-Master (changes requested):

Legend-Master:
Need to get some sleep, will get back to this tomorrow

amrbashir:
double events are fixed in https://github.com/tauri-apps/wry/pull/1418, so this PR can move forward.

plugins/opener/src/reveal_item_in_dir.rs
```diff
@@ -0,0 +1,192 @@
...
+                        lpClass: if is_dir { w!("folder") } else { PCWSTR::null() },
+                        lpFile: PCWSTR(file_h.as_ptr()),
+                        ..std::mem::zeroed()
+                    };
+
+                    ShellExecuteExW(&mut info).inspect_err(|_| {
+                        ILFree(Some(dir_item));
+                        ILFree(Some(file_item));
+                    })?;
+                }
```
Legend-Master: After a closer look, I find that this fallback doesn't quite make sense, this opens the file instead of revealing it

Legend-Master (changes requested):

plugins/opener/src/reveal_item_in_dir.rs
amrbashir: nice catch, fixed

plugins/opener/src/lib.rs
```diff
@@ -0,0 +1,134 @@
...
+impl<R: Runtime, T: Manager<R>> crate::OpenerExt<R> for T {
+    fn opener(&self) -> &Opener<R> {
+        self.state::<Opener<R>>().inner()
+    }
+}
+
+/// Initializes the plugin.
+pub fn init<R: Runtime>() -> TauriPlugin<R> {
+    Builder::new("opener")
+        .js_init_script(include_str!("init-iife.js").to_string())
```
Legend-Master: I still feel like this should be opt-in or at least be optional, and in that case, we'll need this to be a builder or it will be a breaking change when we add it

amrbashir: added

Legend-Master (approved):
Looks good to me! Only tested on Windows though

amrbashir:
@FabianLars can you give this a test on macOS? (and Linux if you want)

FabianLars (approved):
works well on macos and wsl (don't have an actual linux system on hand rn)

PR #2021, opened 2024-11-08
fix(log): inconsistent webview log target

### LINKED ISSUE #1058: [log] target filter doc string example doesn't actually work
This filter will not work

> `|metadata| metadata.target() == WEBVIEW_TARGET`

https://github.com/tauri-apps/plugins-workspace/blob/753c7be0a6a78121d2e88ea0efc3040580c885b4/plugins/log/src/lib.rs#L362-L373

This overrides the target to be `webview:unknown` or something else from `location` parameter instead of `WEBVIEW_TARGET`

https://github.com/tauri-apps/plugins-workspace/blob/753c7be0a6a78121d2e88ea0efc3040580c885b4/plugins/log/src/lib.rs#L233

Legend-Master:
Make webview log target more consistent that it always starts with `"webview"`

The `WEBVIEW_TARGET` target (the value used to be `"Webview"`) is never actually being used before, and the location info was only generated for the webkit based before (also they seem to be the current location instead of the caller's location)

So this PR changes the behavior to use the caller's location, format it like `webview:bar@filename.js:6:3`, fallback to `webview`

Fix #1058


Legend-Master:
Note: I didn't test the `JavaScriptCore` stack trace parser, just used the old one with the third line instead the of the first line

FabianLars (approved):
(didn't test this sry - looks nice though)

PR #2027, opened 2024-11-10
feat(websocket): add ability to remove listener

twlite:
This pr changes listeners array to use a set instead, which prevents addition of duplicated listeners which could possibly lead to memory leaks.


FabianLars:
i agree that a way to remove listeners is needed (that's why i didn't close the issue) but i'm not a fan of the Set here (or how it was used). I can very much imagine that someone wants to register the same function twice for whatever reason. Buuuut i'm totally fine with doing it this way and waiting to see if anyone actually complains.

Also, changing the return value is technically a breaking change but because it didn't return anything before and because this plugin doesn't have _that_ many users i think we can get away with it.

FabianLars (dismissed):

twlite:
> Also, changing the return value is technically a breaking change but because it didn't return anything before and because this plugin doesn't have that many users i think we can get away with it.

Yes, that's exactly what I thought 😅

> i'm not a fan of the Set here (or how it was used). I can very much imagine that someone wants to register the same function twice for whatever reason.

Yeah but it doesn't make sense to do so because same function gets invoked multiple times which can cause weird behaviors.

jffaust:
The ability to remove a listener is pretty important in my opinion. Otherwise, you'd have to close the socket but you might still need the connection for other purposes. I say let's proceed with these changes.  

@twlite do you think you can look at the merge conflicts?

twlite:
> @twlite do you think you can look at the merge conflicts?

done

twlite:
Looks like we got an error saying `Missing "minor" bump for Rust package "websocket" in .changes/cleanup-fn-websocket.md.`. I have not touched the rust part here, do I still have to add it?

FabianLars:
yes. since 2.2.0 the rust and js package must have matching versions (added the workflow at the same time to train ourselves to remember that)

jffaust:
I'm not sure if you received a notification, but it seems the changes are now ready to be re-reviewed @FabianLars :)

jffaust:
Are we good to go @FabianLars ? I looked around the documentation but I couldn't find any details about the release schedule/mechanism. Will merging this PR automatically create a new version of the package on NPM?

jffaust:
@FabianLars a gentle reminder :)

FabianLars:
ay sry, will merge in a second

> Will merging this PR automatically create a new version of the package on NPM?

Not quite, i also have to merge the versioning PR. I tend to do that regularly in this repo though.

FabianLars (approved):

jffaust:
@FabianLars a gentle reminder :)

PR #2033, opened 2024-11-11
feat(upload): add `progressTotal` to event payload

514sid:
This pull request addresses an issue with progress tracking during downloads. Previously, progress was updated using `chunk.len()`, which only reflected the size of each individual chunk rather than the cumulative total.

Changes:

Progress Calculation: Replaced `chunk.len()` with `stats.total_transferred` for progress, ensuring it accurately reflects the cumulative download progress from start to end.

TransferStats Update: Added `total_transferred` to `TransferStats` to maintain an ongoing count of all transferred bytes without resetting per granularity interval.

### DIFF
--- a/.changes/add-total-transferred-field.md
+++ b/.changes/add-total-transferred-field.md
@@ -3,4 +3,4 @@
 "upload-js": "minor"
 ---
 
-Added a new field `total_transferred` (or `progressTotal` in JS) to track the total amount of data transferred during the upload/download process.
+Added a new field `progressTotal` to track the total amount of data transferred during the upload/download process.
--- a/.changes/add-total-transferred-field.md
+++ b/.changes/add-total-transferred-field.md
@@ -1,0 +1,6 @@
+---
+"upload": "minor"
+"upload-js": "minor"
+---
+
+Added a new field `total_transferred` (or `progressTotal` in JS) to track the total amount of data transferred during the upload/download process.
--- a/plugins/upload/guest-js/index.ts
+++ b/plugins/upload/guest-js/index.ts
@@ -6,6 +6,7 @@ import { invoke, Channel } from '@tauri-apps/api/core'
 
 interface ProgressPayload {
   progress: number
+  progressTotal: number
   total: number
   transferSpeed: number
 }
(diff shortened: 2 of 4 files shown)


FabianLars:
This is not a bug and has always been the expected behavior. (Expected from us at least, there were quite a few users confused by this)

Since we're now keeping track of much more data than before the transfer_speed update i wouldn't mind adding a new field for this (keeping progress to represent the chunk size).

In v3 we could consider renaming the progress field but now this would break too many apps.

514sid:
> This is not a bug and has always been the expected behavior. (Expected from us at least, there were quite a few users confused by this)
> 
> Since we're now keeping track of much more data than before the transfer_speed update i wouldn't mind adding a new field for this (keeping progress to represent the chunk size).
> 
> In v3 we could consider renaming the progress field but now this would break too many apps.

How about adding a new field, for example, `total_progress` or `total_transferred`, to track cumulative progress while keeping `progress` for the chunk size?

FabianLars:
Well, i just said the same so yes, i agree :D

i'd go for `total_progress` or `progress_total` and leave `transfer*` for just the speed stuff.

FabianLars:
Thanks, could you also add a changefile like this one? https://github.com/VirtualPirate/plugins-workspace/blob/aa7b4c39ce1d3893f6fb72a06046f738f956fe52/.changes/add-transfer-speed.md

514sid:
> `progress_total`

It's probably more convenient that way

.changes/add-total-transferred-field.md
```diff
@@ -0,0 +1,6 @@
+---
+"upload": "minor"
+"upload-js": "minor"
+---
+
+Added a new field `total_transferred` (or `progressTotal` in JS) to track the total amount of data transferred during the upload/download process.
```
FabianLars:
```suggestion
Added a new field `progressTotal` to track the total amount of data transferred during the upload/download process.
```
The rust side is not exposed outside of the crate so our users don't have to know about this here.

FabianLars (approved):
thank you

514sid:
> thank you

thank you too
I really appreciate all the work you’re doing — it’s awesome

PR #2038, opened 2024-11-12
fix(sql) Allow tauri-plugin-sql to work when Tauri is running async

johncarmack1984:
At present, `tauri-plugin-sql` only works when the main Tauri thread is using the default, synchronous runtime. 

In certain configurations, running Tauri asynchronously can be necessary, ie when using TaurPC or other dependencies that require it.

With the current configuration, running Tauri asynchronously while trying to use `tauri-plugin-sql` yields this error:

```
thread 'main' panicked at ~/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.40.0/src/runtime/scheduler/multi_thread/mod.rs:86:9:
Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

By changing the `block_on` function call in `sql/lib.rs` to the macro provided, the plugin works flawlessly, in both synchronous and asynchronous runtimes.

Please let me know if more information is needed!

### DIFF
--- a/.changes/sql-allow-blocking-without-nested-runtime.md
+++ b/.changes/sql-allow-blocking-without-nested-runtime.md
@@ -1,0 +1,5 @@
+---
+"sql": "patch"
+---
+
+Allow blocking on async code without creating a nested runtime.
--- a/plugins/sql/src/lib.rs
+++ b/plugins/sql/src/lib.rs
@@ -106,7 +106,11 @@ impl MigrationSource<'static> for MigrationList {
 
 /// Allows blocking on async code without creating a nested runtime.
 fn run_async_command<F: std::future::Future>(cmd: F) -> F::Output {
-    tokio::task::block_in_place(|| tokio::runtime::Handle::current().block_on(cmd))
+    if tokio::runtime::Handle::try_current().is_ok() {
+        tokio::task::block_in_place(|| tokio::runtime::Handle::current().block_on(cmd))
+    } else {
+        tauri::async_runtime::block_on(cmd)
+    }
 }
 
 /// Tauri SQL plugin builder.
--- a/plugins/sql/src/lib.rs
+++ b/plugins/sql/src/lib.rs
@@ -104,10 +104,9 @@ impl MigrationSource<'static> for MigrationList {
     }
 }
 
-macro_rules! run_async_command {
-    ($cmd:expr) => {
-        tokio::task::block_in_place(|| tokio::runtime::Handle::current().block_on($cmd))
-    };
+/// Allows blocking on async code without creating a nested runtime.
+fn run_async_command<F: std::future::Future>(cmd: F) -> F::Output {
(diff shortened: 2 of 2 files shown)


FabianLars:
Makes me wonder why we didn't make https://github.com/tauri-apps/tauri/blob/dev/crates%2Ftauri%2Fsrc%2Fasync_runtime.rs#L288 the default 🤔

FabianLars:
Actually, I think having a centralized approach would be better, this could happen in every plugin because they kinda as expect a sync main fn

FabianLars:
@amrbashir do you think pub exposing safe_block_on in core makes sense?

amrbashir:
maybe, idk, I don't use async that much

FabianLars:
okay no problem, let's keep it plugin-local for now and see if this gets more common.

@johncarmack1984 would you mind checking if the safe_block_on approach also works here? I assume there's a reason it doesn't use block_in_place.

Also, i think a code comment explaining why this was added (so nobody removes it by accident) would be valuable.

And idk if it's just me but i'd rather not solve this with a macro but just a simple function call if there's no good reason for a macro.

Lastly, we'll also need a changefile.

johncarmack1984:
Sure thing! If I clone Tauri locally and modify `crates/tauri/src/async_runtime.rs` to make `block_in_place` public, then use `tauri::async_runtime::safe_block_on` in `tauri-plugin-sql` instead of the macro, I get:
```
`dyn StdError` cannot be sent between threads safely
the trait `Send` is not implemented for `dyn StdError`, which is required by `Result<(), _>: Send`
required for `Unique<dyn StdError>` to implement `Send`
```

I'll add an explainer comment and get started on a changefile, hopefully will have it later today!

Will also see if functionality is preserved when using a function vs a macro and report back if not.

johncarmack1984:
Success converting macro to function, have added changefile and explainer comment! Please let me know if I can provide anything else! :)

FabianLars:
Okay now i remember why safe_block_on was bigger than that. Your change currently panics in sync contexts.

johncarmack1984:
Oof... yikes, I'll take another crack at it. Any initial thoughts on approaches that might satisfy both? If not I'll troubleshoot it today.

FabianLars:
a modified version of safe_block_on. Using Handle::try_current to check if there's a runtime and then using block_in_place or block_on directly.

johncarmack1984:
> a modified version of safe_block_on. Using Handle::try_current to check if there's a runtime and then using block_in_place or block_on directly.

Worked a charm; sync and async both operating locally. Thanks for catching that!

FabianLars (approved):

johncarmack1984:
(just want to pop in here and give credit to @jodavaho for the initial macro, thanks for the collaboration everybody)

Issue #1578, opened 2024-07-23
[dialog] [iOS] open([options]) brings up photo picker and not file select


bjbk:
Install plugin-dialog using manual install
JS API installed with pnpm

Import plugin-dialog and then use the following code:
```js
await open({
      multiple: false,
      directory: false,
      properties: [{ mimeTypes: ['text/csv'] }],
      filters: [{ name: 'CSV', extensions: ['csv'] }],
    })
```

Running `pnpm tauri dev` for desktop, the code brings up a file picker dialog as expected allowing the selection of a CSV file.

## iOS issue

Running `pnpm tauri ios dev` choosing a simulator then when the app is running, tap the button to bring up the file picker.  The photo app picker is what comes up.  No other means to select a file from the Files app.  A photo can be selected and returned.  What is needed though is a means to select a file not a photo.

__Expected__ open file picker dialog

Tied setting permissions in capabilities individually.  No change

### Other information

capabilities/default.json
```json
{
  "$schema": "../gen/schemas/desktop-schema.json",
  "identifier": "default",
  "description": "Capability for the main window",
  "windows": [
    "main"
  ],
  "permissions": [
    "store:default",
    "store:allow-get",
    "store:allow-set",
    "store:allow-save",
    "store:allow-load",
    "store:allow-clear",
    "store:allow-delete",
    "store:allow-has",
    "store:allow-keys",
    "store:allow-length",
    "store:allow-reset",
    "store:allow-values",
    "path:default",
    "event:default",
    "window:default",
    "app:default",
    "image:default",
    "resources:default",
    "menu:default",
    "tray:default",
    "shell:allow-open",
    "fs:default",
    "dialog:default"
  ]
}
```
`package.json`
```json
...
"@tauri-apps/plugin-dialog": "2.0.0-beta.7",
...
```
`lib.rs`
```rust
use tauri::Manager;

#[cfg_attr(mobile, tauri::mobile_entry_point)]
pub fn run() {
    let mut builder = tauri::Builder::default()
        .setup(|app| {
            #[cfg(debug_assertions)]
            {
                let window = app.get_webview_window("main").unwrap();
                window.open_devtools();
                window.close_devtools();
            }
            Ok(())
        })
        .plugin(tauri_plugin_dialog::init())
        .plugin(tauri_plugin_fs::init())
        .plugin(tauri_plugin_shell::init())
        .plugin(tauri_plugin_os::init())
[...]

Issue #1984, opened 2024-10-30
[dialog][v2] Should the `name` of the `filters` parameter of the `open` method be made optional?


ayangweb:
Should the `name` of the `filters` parameter of the `open` method be made optional?

```ts
import { open } from "@tauri-apps/plugin-dialog";

open({
  filters: [{ extensions: [extname()] }],
})
```

<img width="387" alt="image" src="https://github.com/user-attachments/assets/9b614a4f-1020-4c42-93cd-62bb1cc2312f">

<img width="711" alt="image" src="https://github.com/user-attachments/assets/6dc707ce-169a-48fb-ab6f-9cab8d41e94c">

Issue #1596, opened 2024-07-30
[feat] [dialog] How to determine whether to open photos picker or files picker on iOS


winjeysong:
using `open` imported from `@tauri-apps/plugin-dialog` will open files picker on iOS 17.5 simulator. 
Is there a way to determine whether to open photos picker or files picker? I just want to open photos picker by `open`.

tauri-plugin-dialog: 2.0.0-beta.11

FabianLars:
Considering we have the opposite behavior in https://github.com/tauri-apps/plugins-workspace/issues/1578 i have to admit that this is quite confusing to me. Anyway, just wrote this to link to 1578 so that when someone works one of those issues the other one will be looked at & resolved as well.

Thanks for the report.

ecmel:
From my testing the following opens the file dialog:
```ts
await open({ filters: [{ extensions: ["pdf", "txt"], name: "Documents" }]});
```
but interestingly the following opens photo chooser:
```ts
await open({ filters: [{ extensions: ["pdf"], name: "Documents" }]});
```
Checking the [source code](https://github.com/tauri-apps/plugins-workspace/blob/v2/plugins/dialog/ios/Sources/DialogPlugin.swift#L50), `open` method can show multiple dialogs depending on the parsing of the `filters` attribute so both issues are valid. Parsing logic should be fixed.

bjbk:
Confirmed.  Adding the second extension to the extensions array opens the Files app.  However, for some reason the target files are greyed out and not selectable.  
```js
file = await open({
        multiple: false,
        directory: false,
        filters: [{ name: 'CSV', extensions: ['txt', 'csv'] }]
      })
```

bjbk:
I wonder if this may be of use:

PLEASE USE CAUTION WITH THE CODE BELOW, I am not a Swift developer.  I was looking to see what could be the issue with parsing the code and so asked Claude AI.  I realize that asking AI may be bad form, but sometimes it proves to be useful.

The code is using `kUTTagClassMIMEType` to create the UTI, but it's passing in the file extension. Instead, should it use `kUTTagClassFilenameExtension`?

```swift
import MobileCoreServices
import Photos
import PhotosUI
import SwiftRs
import Tauri
import UIKit
import WebKit

// ... (previous code remains the same)

class DialogPlugin: Plugin {
  var filePickerController: FilePickerController!
  var pendingInvoke: Invoke? = nil
  var pendingInvokeArgs: FilePickerOptions? = nil

[...]

PR #2061, opened 2024-11-16
fix(dialog): improve file type handling and media type checks

garyhai:
Fixes bug #1578, allowing different file types on iOS to correctly use the image picker and document picker. This update also addresses issues discussed in #1596 and #1984.


nab138:
Any updates on this?

FabianLars:
implemented in https://github.com/tauri-apps/plugins-workspace/pull/3034 instead, sorry. Thanks for the pr though :)

PR #2070, opened 2024-11-18
fix(fs): Escape paths in Scope methods

### LINKED ISSUE #11707: [bug] Paths with open an close square brackets [] in the path are not correctely added to accessible scope
### Describe the bug

If a path has square brackets in it, e.g. `/path/[to]/a.txt` or `/path/to/[a].txt`, then that path will not work with fs plugin methods. Even if the path is selected with dialog.open, which should add selected path to allowed scope. 

### Reproduction

Example project: https://github.com/lendle/tauri-app-bad-filenames

1. Run the app in dev mode
2. Click the button, select any file that does not have square brackets in the path.
You will see something like 
```
path: /path/to/a.txt
msg: /path/to/a.txt exists: true
error: 
```
3. Click the button, select a file with matching brackets in the path, like `[a].txt`. You will see
```
path: /path/to/[a].txt
msg: 
error: "forbidden path: /path/to/[a].txt"
  
```

### Expected behavior

Paths with square brackets should be correctly added to allowed scope

### Full `tauri info` output

```text
npm run tauri info

> tauri-app-bad-filenames@0.1.0 tauri
> tauri info


[✔] Environment
    - OS: Mac OS 14.6.1 arm64 (X64)
    ✔ rustc: 1.79.0 (129f3b996 2024-06-10)
    ✔ cargo: 1.79.0 (ffa9cf99a 2024-06-03)
    ✔ Rust toolchain: 1.79.0-aarch64-apple-darwin (default)
    - node: 22.5.1
    - npm: 10.8.3


[-] Packages
    - tauri 🦀: 2.1.1
    - tauri-build 🦀: 2.0.3
    - wry 🦀: 0.47.0
    - tao 🦀: 0.30.8
    - @tauri-apps/api : 2.1.1
    - @tauri-apps/cli : 2.1.0

[-] Plugins
    - tauri-plugin-fs 🦀: 2.0.3
    - @tauri-apps/plugin-fs : 2.0.2
    - tauri-plugin-shell 🦀: 2.0.2
    - @tauri-apps/plugin-shell : 2.0.1
    - tauri-plugin-persisted-scope 🦀: 2.0.3
    - @tauri-apps/plugin-persisted-scope : not installed!
    - tauri-plugin-dialog 🦀: 2.0.3
    - @tauri-apps/plugin-dialog : 2.0.1

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../build
    - framework: Svelte
    - bundler: Vite
```

### Stack trace

_No response_

### Additional context

_No response_

### LINKED ISSUE #11708: [bug] A single, unclosed open bracken in a path will corrupt the filesystem scope permanently
### Describe the bug

When tauri tries to add a path the filesystem scope, if that path has an unmatched square open bracket in it, (`[a.txt`) the filesystem scope gets corrupted, rendering most or all fs plugin methods useless. If the persisted-scope plugin is installed, this will persist even when the app restarts.  This happens with paths selected with dialog.open or if the user drag and drops a badly named file into a tauri window, **even if no drag and drop event listener is registered**. 

Deleting $APPLOCALDATA/.persisted-scope fixes this, but does lose persisted scope info. 

### Reproduction

Example project: https://github.com/lendle/tauri-app-bad-filenames

1. Run the app in dev mode
2. Click the button, select any file that does not have square brackets in the path. You will see something like
```
path: /path/to/a.txt
msg: /path/to/a.txt exists: true
error: 
```
3. Either drag and drop a file with an unmatched open bracket in the path into the window, or selected one with in the dialog when clicking the "Click Me" button.
4. Any time you select any file after that, you will see something like
```
path: /some/path/which/may/or/may/not/have/brackets
msg: 
error: "invalid glob pattern: Pattern syntax error near position 11: invalid range pattern"
  ```
  

### Expected behavior

1. Paths with unmatched open brackets are usable
2. Even if not, they don't permanently corrupt the scope

### Full `tauri info` output

```text
npm run tauri info

> tauri-app-bad-filenames@0.1.0 tauri
> tauri info


[✔] Environment
    - OS: Mac OS 14.6.1 arm64 (X64)
    ✔ rustc: 1.79.0 (129f3b996 2024-06-10)
    ✔ cargo: 1.79.0 (ffa9cf99a 2024-06-03)
    ✔ Rust toolchain: 1.79.0-aarch64-apple-darwin (default)
    - node: 22.5.1
    - npm: 10.8.3


[-] Packages
    - tauri 🦀: 2.1.1
    - tauri-build 🦀: 2.0.3
    - wry 🦀: 0.47.0
    - tao 🦀: 0.30.8
    - @tauri-apps/api : 2.1.1
    - @tauri-apps/cli : 2.1.0

[-] Plugins
    - tauri-plugin-fs 🦀: 2.0.3
    - @tauri-apps/plugin-fs : 2.0.2
    - tauri-plugin-shell 🦀: 2.0.2
    - @tauri-apps/plugin-shell : 2.0.1
    - tauri-plugin-persisted-scope 🦀: 2.0.3
    - @tauri-apps/plugin-persisted-scope : not installed!
    - tauri-plugin-dialog 🦀: 2.0.3
    - @tauri-apps/plugin-dialog : 2.0.1

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../build
    - framework: Svelte
    - bundler: Vite
```

### Stack trace

_No response_

### Additional context

There seems to be no way to guard against this because both drag and drop and dialog open attempt to add the path to the scope immediately, so you cannot inspect the path and reject it.

FabianLars:
opening as a draft because it doesn't compile locally, complaining about glob not being a dependency but only if it's uses in scope.rs - if it fails in ci too i'll just move it to lib.rs.

if anyone spots my mistake here, please enlighten me...

fixes https://github.com/tauri-apps/tauri/issues/11707
fixes https://github.com/tauri-apps/tauri/issues/11708


amrbashir:
this is because the scope module is used in build.rs, see https://github.com/tauri-apps/plugins-workspace/blob/b42064dc70b799c6777ca3a9420968d53a7a04ef/plugins/fs/build.rs#L10-L12

you can just add `glob` to the plugin `build-dependencies` and it should work

FabianLars:
Ahhhhhh yes, forgot about that...

FabianLars:
Oh and thanks. I'll fix it when I'm back

FabianLars:
this seems to work for now but i'm less than happy.

i think it may make sense to release this is a temp fix/workaround and wait for lucas to get back to discuss further steps.

Alternatively i'm thinking about storing a `tauri::scope::fs::Scope` inside of `tauri_plugin_fs::Scope` and using that to keep track of the actual scope instead of the PathBuf Vec (keeping the latter around to not have a breaking change...). But ideally really we could deduplicate the scope definitions themselves but i'm sure that was already tried.

idk, i guess i'll take another look tomorrow.

plugins/fs/Cargo.toml
```diff
@@ -14,7 +14,7 @@ rustc-args = ["--cfg", "docsrs"]
 rustdoc-args = ["--cfg", "docsrs"]
 
 [package.metadata.platforms.support]
-windows = { level = "full", notes = "" }
+windows = { level = "full", notes = "No write access to `$RESOURCES` folder with MSI installer and NSIS installers in `perMachine` or `both` mode" }
```
amrbashir: I think running the app as admin would give you write access.

plugins/fs/src/scope.rs
```diff
@@ -129,3 +137,27 @@ impl Scope {
         id
     }
 }
+
+// taken from https://github.com/rust-lang/glob/blob/master/src/lib.rs#L717C5-L737C6
+/// Escape metacharacters within the given string by surrounding them in
+/// brackets. The resulting string will, when compiled into a `Pattern`,
+/// match the input string and nothing else.
+pub fn escape(s: &str) -> PathBuf {
```
amrbashir: we can use `glob::Pattern::escape` directly, `glob` is already a dependency of tauri, so it doesn't matter if we include it as a dependency here as well

plugins/fs/src/scope.rs
```diff
@@ -56,8 +56,9 @@ impl Scope {
 
         {
             let mut allowed = self.allowed.lock().unwrap();
-            allowed.push(path.to_path_buf());
-            allowed.push(path.join(if recursive { "**" } else { "*" }));
+            let p = path.to_string_lossy();
+            allowed.push(escape(&p));
+            allowed.push(PathBuf::from(if recursive { "**" } else { "*" }));
```
amrbashir: shouldn't this be `escape(&p).join(if recursive { "**" } else { "*" })`, otherwise I think we are allowing access to all and any directory

plugins/fs/src/scope.rs
```diff
@@ -82,8 +86,9 @@ impl Scope {
 
         {
             let mut denied = self.denied.lock().unwrap();
-            denied.push(path.to_path_buf());
-            denied.push(path.join(if recursive { "**" } else { "*" }));
+            let p = path.to_string_lossy();
+            denied.push(escape(&p));
+            denied.push(PathBuf::from(if recursive { "**" } else { "*" }));
```
amrbashir: ditto

amrbashir (changes requested):

plugins/fs/Cargo.toml
FabianLars: yeah likely. totally makes sense to include it, didn't think about it.

amrbashir:
> Alternatively i'm thinking about storing a `tauri::scope::fs::Scope` inside of `tauri_plugin_fs::Scope` and using that to keep track of the actual scope instead of the PathBuf Vec (keeping the latter around to not have a breaking change...). But ideally really we could deduplicate the scope definitions themselves but i'm sure that was already tried.

actually I think we maybe able to reuse the one from `tauri::scope::fs::Scope` and looking at the fs plugin, it doesn't actually make sense to have a separate implementation at all, but let's refactor that in another PR.

plugins/fs/src/scope.rs
FabianLars: i changed it to not escape `\\?\` because that's what caused memory leaks in v1 (with the persisted scope plugin?). It may work but i didn't even try cause of ptsd.

plugins/fs/src/scope.rs
FabianLars: i have no words

plugins/fs/src/scope.rs
amrbashir: Allowing `?` not to be escaped will cause a small security vulnerability, we can look into the memory leak separately.

And don't push yourself, take it slow. I wish you a speedy recovery.

FabianLars: > Allowing ? not to be escaped will cause a small security vulnerability, we can look into the memory leak separately.

Just sounds like a reason to dismiss this pr and try to work on a proper fix instead.

FabianLars: oh btw, idk if you were joking or not but just to be safe, with ptsd i meant the memory leak, not actual ptsd (~~i only have that from #support~~).

amrbashir: > Just sounds like a reason to dismiss this pr and try to work on a proper fix instead.

I will let you be the judge of that

amrbashir: > oh btw, idk if you were joking or not but just to be safe, with ptsd i meant the memory leak, not actual ptsd (~i only have that from #support~).

lol, my bad 😂

FabianLars:
i already miss the time when it was ok to break user facing apis 😂 

what do you think about the new approach?

FabianLars:
also thought about a new FsExt where fs_scope() returns a new scope based on tauri's fs::Scope and maybe making the old apis no-op but idk, everything i can think of is ugly

plugins/fs/src/commands.rs
```diff
@@ -1002,37 +1002,88 @@ pub fn resolve_path<R: Runtime>(
...
+        .unwrap_or(false)
+        || scope.is_allowed(&path)
+    {
         Ok(path)
     } else {
         Err(CommandError::Plugin(Error::PathForbidden(path)))
     }
 }
 
+fn is_forbidden<P: AsRef<Path>>(
```
amrbashir, line 1015: why is this implemented separately here?

plugins/fs/src/scope.rs
```diff
@@ -39,10 +25,8 @@ pub enum Event {
 
 #[derive(Default)]
 pub struct Scope {
-    pub(crate) allowed: Mutex<Vec<PathBuf>>,
-    pub(crate) denied: Mutex<Vec<PathBuf>>,
-    event_listeners: Mutex<HashMap<EventId, EventListener>>,
-    next_event_id: AtomicU32,
+    // TODO: Remove Option in v2, just used to keep Default
```
amrbashir: let's just remove this whole scope implementation and expose tauri's internal scope and break the APIs of this plugin in a minor version.

amrbashir (changes requested):

plugins/fs/src/scope.rs
FabianLars: that's what i wanted to hear

FabianLars: or wait, what do you mean with the "expose tauri's internal scope" part?

plugins/fs/src/commands.rs
FabianLars: i can implement it in tauri::fs::Scope too and re-use that but wanted to keep the plugin compatible with all tauri v2 versions for now.

not too happy about the approach itself but merging instances of `tauri::fs::Scope`s isn't possible rn so this was the next best thing to make the draft work.

plugins/fs/src/scope.rs
amrbashir: nvm I thought we have something similar to `Manager::asset_protocol_scope` but for general fs operations.

FabianLars:
this should be done now apart from the (ongoing?) discussion about is_forbidden

plugins/persisted-scope/src/lib.rs
```diff
@@ -269,7 +195,7 @@ pub fn init<R: Runtime>() -> TauriPlugin<R> {
 
                     for allowed in &scope.allowed_paths {
                         let allowed = fix_pattern(&ac, allowed);
-                        allow_path(fs_scope, &allowed);
+                         allow_path(fs_scope, &allowed);
```
amrbashir: weird formatting

plugins/dialog/src/commands.rs
```diff
@@ -143,7 +143,7 @@ pub(crate) async fn open<R: Runtime>(
                     for folder in folders {
                         if let Ok(path) = folder.clone().into_path() {
                             if let Some(s) = window.try_fs_scope() {
-                                s.allow_directory(&path, options.recursive);
+                                let _ = s.allow_directory(&path, options.recursive);
```
amrbashir: we should propagate the error, similar to how `tauri_scope` is propagated

plugins/fs/src/commands.rs
amrbashir: let's ship with this for now, but I think we should still add it in `tauri` and then migrate the `fs` plugin later to that.

amrbashir (changes requested):

plugins/fs/src/commands.rs
FabianLars: ok, will open a pr later ~~if i don't forget~~

plugins/persisted-scope/src/lib.rs
FabianLars: weird that no tools picked that up

plugins/dialog/src/commands.rs
FabianLars: ah yes, of course

plugins/fs/src/commands.rs
```diff
@@ -999,40 +1001,81 @@ pub fn resolve_path<R: Runtime>(
...
-    if scope.is_allowed(&path) {
+    let require_literal_leading_dot = fs_scope.require_literal_leading_dot.unwrap_or(cfg!(unix));
+
+    if is_forbidden(&fs_scope.scope, &path, require_literal_leading_dot)
+        || is_forbidden(&scope, &path, require_literal_leading_dot)
+    {
+        return Err(CommandError::Plugin(Error::PathForbidden(path)));
+    }
+
+    if fs_scope.scope.is_allowed(&path) || scope.is_allowed(&path) {
```
amrbashir, line 1008: do really need to manually call `is_forbidden`, doesn't `scope.is_allowed` check for forbidden/denied patterns?

FabianLars: the problem is that is_allowed returns false for paths that are explictly denied _and_ paths that are _neither_ allowed nor denied.
When we're checking for 2 scopes at the same time we only care about the explicitly denied paths because those take precedence of explictly allowed paths.
A path that's neither allowed nor denied in the first scope could still be explicitly allowed by the second scope.

Alternatively a way to create a new Scope from 2 existing scopes would be cool too, then is_allowed would work for us here.

amrbashir (dismissed):

plugins/fs/src/commands.rs
amrbashir: I guess we can improve this a bit by adding the `Scope::is_forbidden` in tauri

amrbashir (approved):

PR #2076, opened 2024-11-19
(Positioner) Allow for Screen constrained positioning when using tray-icon feature

jakobwesthoff:
**positioner** plugin related:

This PR implements a new trait fn on the `Window` called `move_window_constrained`, which is only available with the already existent ´tray-icon` feature. It allows to position a window based on the tray icon position, as well as the given `Tray*` variants from the `Position` enum, while constraining the window to the screen, where the tray icon lives. This provides a way of positioning, which ensures the window is not clipped by the borders of the screen, when being positioned relative to the tray icon.

Introducing the new method, allows the user to explicitly choose the new behavior over the old one, while maintaining BC without cluttering the `Position` enum with further options.

I refactored the code a little to ensure minimal duplication and to ease the maintenance in the future of both features.

Please let me know if you want anything changed with regards to general realization or coding style.

before

after


amrbashir (approved):
Thank you

jbolda:
🚀

PR #2079, opened 2024-11-20
feat(http): add request and response tracing behind feature flag

### LINKED ISSUE #2055: [http] can't debug requests
with fetch i can see plain request/response body
with tauri fetch i can't
does it possible? and if not how can we improve it?

amrbashir:
closes #2055

Users can use tauri-plugin-log or CrabeNebula devtools to inspect the requests and responses.

### DIFF
--- a/.changes/http-tracing.md
+++ b/.changes/http-tracing.md
@@ -1,0 +1,5 @@
+---
+"http": "patch"
+---
+
+Add tracing logs for requestes and responses behind `tracing` feature flag.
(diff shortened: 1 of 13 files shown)


FabianLars:
iirc in tauri we're using tracing in tauri (and log in cli stuff). i think that makes sense here too? I also still use log in my projects but i hear again and again that tracing should be used instead especially in async contexts.

FabianLars (approved):

PR #2095, opened 2024-11-25
feat(fs): add the `size` method to get the size of a file or directory

ayangweb:
@FabianLars Help me refine it a bit more, thanks!


plugins/fs/guest-js/index.ts
```diff
@@ -1322,6 +1322,28 @@ async function watchImmediate(
...
+ * Get the size of a file or directory.
+ * @example
+ * ```typescript
+ * import { size, BaseDirectory } from '@tauri-apps/plugin-fs';
+ * // Get the size of the `$APPDATA/tauri` directory.
+ * const dirSize = await size('tauri', { baseDir: BaseDirectory.AppData });
+ * console.log(dirSize); // 1024
+ * ```
+ *
+ * @since 2.0.0
```
FabianLars:
```suggestion
 * @since 2.1.0
```

plugins/fs/src/commands.rs
```diff
@@ -3,6 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-License-Identifier: MIT
 
+use anyhow::Ok;
```
FabianLars:
```suggestion
```

Do we really need this? If so, which line does rust complain about if it's not there?

plugins/fs/guest-js/index.ts
```diff
@@ -1322,6 +1322,28 @@ async function watchImmediate(
   }
 }
 
+/**
+ * Get the size of a file or directory.
```
FabianLars:
```suggestion
 * Get the size of a file or directory. For files, the `stat` functions can be used as well.
 *
 * If `path` is a directory, this function will recursively iterate over every file and every directory inside of `path` and therefore will be very time consuming if used on larger directories.
```

Maybe we should also add a note on the stat functions that it does not return the expected dir size and that this function should be used if that is required.

FabianLars (changes requested):
thanks! just a few small thinks. Also, we'll need a changefile like https://github.com/tauri-apps/plugins-workspace/blob/v2/.changes/fix-fs-scope-escape-paths.md

plugins/fs/src/commands.rs
ayangweb: Haha, I didn't even notice using it

ayangweb:
Thank you for your guidance, I've revised it, and you can look at it again!

.changes/add-fs-size.md
```diff
@@ -0,0 +1,6 @@
+---
+fs: minor
+persisted-scope: minor
+---
+
+**Breaking Change:** add the `size` method to get the size of a file or directory.
```
FabianLars:
```suggestion
Add the `size` method to get the size of a file or directory.
```
My bad, i gave you a stupid example file. This change is not breaking.

.changes/add-fs-size.md
```diff
@@ -0,0 +1,6 @@
+---
+fs: minor
+persisted-scope: minor
```
FabianLars:
```suggestion
```

Same

FabianLars (commented):
I'll review the rest later, I'm just on a phone rn

ayangweb:
```md
---
fs: minor
---

Add the `size` method to get the size of a file or directory.
```

Change it to this?

plugins/fs/guest-js/index.ts
```diff
@@ -1322,6 +1322,30 @@ async function watchImmediate(
   }
 }
 
+/**
+ * Get the size of a file or directory. For files, the `stat` functions can be used as well.
+ *
+ * If `path` is a directory, this function will recursively iterate over every file and every directory inside of `path` and therefore will be very time consuming if used on larger directories.
+ * @example
```
FabianLars, line 1330:
```suggestion
 *
 * @example
```

FabianLars:
the errors should be gone once you add the command to the invoke_handler list https://github.com/ayangweb/plugins-workspace/blob/fs-size/plugins/fs/src/lib.rs#L396

ayangweb:
> the errors should be gone once you add the command to the invoke_handler list https://github.com/ayangweb/plugins-workspace/blob/fs-size/plugins/fs/src/lib.rs#L396

ok, I added it!

amrbashir (dismissed):

amrbashir:
looks like you also need to run `pnpm build` and potentially `pnpm format`

ayangweb:
> looks like you also need to run `pnpm build` and potentially `pnpm format`

Okay, I executed and submitted it!

amrbashir (approved):

FabianLars (approved):

PR #2123, opened 2024-12-02
Incorrect plugin registration path in the usage section of every plugin's README

Muhammed-Rahif:


FabianLars (approved):
thank you!

Muhammed-Rahif:
I actually found this mistake while reading readme on npm; better to update it there too.

FabianLars:
that should be fixed on the next release then (hopefully)

PR #2130, opened 2024-12-03
ci: Force the same version bumps for rs and js packages

FabianLars:
I just fucked up the fs plugin and can't unpublish the js package because it has dependencies.

what i actually wanted to have was to remove the `-js` bumps and have the "rust" bumps work for both packages but i don't think that's possible with covector? idk covector is too hard for me so i went with this for now.

To be clear, the idea is to have matching versions for the rust and javascript packages of a plugin but _not_ across plugins.

Once CI is happy i'll remove the test files and add a changefile for all plugins, to then bump _all_ plugins to the same minor in the covector PR to mark the start of the synced version.

### DIFF
--- a/.changes/bad.md
+++ b/.changes/bad.md
@@ -1,9 +1,0 @@
----
-upload: patch
-upload-js: minor
-fs: patch
-fs-js: patch
-localhost: patch
----
-
-test
--- a/.changes/bad.md
+++ b/.changes/bad.md
@@ -1,0 +1,5 @@
+---
+upload: patch
+---
+
+test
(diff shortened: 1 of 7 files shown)


FabianLars:
Okay, CI is stuck on other workflows. I'll mark this as a draft because i have to leave for a few hours.

FabianLars:
Copying my discord message:

This approach fixes a couple issues:
1) makes it easy to identify versions that work together
2) makes the version specifier in `tauri add` make sense
3) makes it possible to have an automatic check in tauri dev or tauri info to confirm js == rs

PR #2140, opened 2024-12-04
Add support for HTTP Streaming

### LINKED ISSUE #2129: [bug] plugin-http is unable to perform streaming HTTP requests
### Describe the bug

Here is demo:

```typescript
const response = await fetch(`http://xxx:port/xxx`, {
  method: 'POST',
  headers: {
    'Content-Type': 'application/json',
    'Accept': 'text/event-stream',
  },
  body: {/* Some body */},
});

if (!response.body) {
  throw new Error('No response body');
}

const decoder = new TextDecoder();

while (true) {
  const {done, value} = await reader.read();

  if (done) break;

  const chunk = decoder.decode(value);
  const lines = chunk.split('\n');

  for (const line of lines) {
    if (line.startsWith('data: ')) {
      const data = line.slice(6).trim();

      if (data === '[DONE]') {
        console.log('Stream completed');
        break;
      } else if (data) {
        console.log('Received chunk:', data);
      }
    }
  }
}
```

With the same code, when I use the browser's native fetch, it runs normally and outputs results in a streaming manner. However, when using Tauri's plugin-http, it blocks until the server outputs all content and then outputs everything at once, which means it blocks at `const = await fetch()`

### Reproduction

_No response_

### Expected behavior

_No response_

### Full `tauri info` output

```text
[✔] Environment
    - OS: Mac OS 15.1.1 arm64 (X64)
    ✔ rustc: 1.82.0 (f6e511eec 2024-10-15)
    ✔ cargo: 1.82.0 (8f40fc59f 2024-08-21)
    - node: 22.8.0
    - pnpm: 9.14.2
    - yarn: 1.22.21
    - npm: 10.9.0

[-] Packages
    - tauri 🦀: 2.0.3
    - tauri-build 🦀: 2.0.1
    - wry 🦀: 0.46.0
    - tao 🦀: 0.30.3
    - @tauri-apps/api : 2.0.2 (outdated, latest: 2.1.1)
    - @tauri-apps/cli : 2.0.0-rc.13 (outdated, latest: 2.1.0)

[-] Plugins
    - tauri-plugin-dialog 🦀: 2.0.1
    - @tauri-apps/plugin-dialog : 2.0.1
    - tauri-plugin-fs 🦀: 2.0.1
    - @tauri-apps/plugin-fs : 2.0.2
    - tauri-plugin-http 🦀: 2.0.1
    - @tauri-apps/plugin-http : 2.0.1
    - tauri-plugin-sql 🦀: 2.0.1
    - @tauri-apps/plugin-sql : 2.0.1
    - tauri-plugin-single-instance 🦀: 2.0.1
    - tauri-plugin-log 🦀: 2.0.1
    - @tauri-apps/plugin-log : 2.0.0
    - tauri-plugin-os 🦀: 2.0.1
    - @tauri-apps/plugin-os : 2.0.0
    - tauri-plugin-updater 🦀: 2.0.2
    - @tauri-apps/plugin-updater : 2.0.0
    - tauri-plugin-notification 🦀: 2.0.1
    - @tauri-apps/plugin-notification : 2.0.0
    - tauri-plugin-autostart 🦀: 2.0.1
    - @tauri-apps/plugin-autostart : 2.0.0
    - tauri-plugin-process 🦀: 2.0.1
    - @tauri-apps/plugin-process : 2.0.0
    - tauri-plugin-shell 🦀: 2.0.1
    - @tauri-apps/plugin-shell : 2.0.1
    - tauri-plugin-global-shortcut 🦀: 2.0.1
    - @tauri-apps/plugin-global-shortcut : 2.0.0

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - devUrl: http://localhost:10002/
    - framework: Vue.js
    - bundler: Vite
```

### Stack trace

_No response_

### Additional context

_No response_

### LINKED ISSUE #2155: [feat] Progress reporting for HTTP API in JS
### Describe the problem

When using the Tauri HTTP API, I'd like to know how much of a file is downloaded so that I can create a progress bar in my app.

It would be nice if the HTTP api's fetch function could take a progress callback.

### Alternatives considered

Another option is to provide access to a `ReadableStream` object as a response type, allowing the developer to access data as it becomes available.

### Additional context

_No response_

robertlong:
This PR changes the internals of the fetch implementation such that `fetch_read_body` uses the `Channel` API to stream http responses instead of waiting for the entire response to finish and then returning the final ArrayBuffer.

Two things to note:

1. The `Channel<&[u8]>` type is serializing as a `number[]` instead of `ArrayBuffer` or `Uint8Array`. I'm not certain that the Channel supports serializing as `ArrayBuffer`s? I think this could be a possible performance improvement.

2. The `Channel` API implementation does not guarantee that all events are received by the time the `invoke` promise is resolved. This is unfortunate, but I worked around it by returning an empty array over the `Channel` which is used to close the `ReadableStream`. I think this design works alright, but maybe there's a cleaner way to do this.

Fixes https://github.com/tauri-apps/plugins-workspace/issues/2129#


plugins/http/guest-js/index.ts
```diff
@@ -206,24 +206,44 @@ export async function fetch(
...
-      ? body
-      : body instanceof Array && body.length > 0
-        ? new Uint8Array(body)
-        : null,
-    {
-      status,
-      statusText
+      // If the promise fails, make sure the stream is closed
+      readPromise.catch((e) => {
+        console.error('error reading body', e)
```
amrbashir, line 233: do we really need to console.error here? won't controller.error be enough?

plugins/http/src/commands.rs
```diff
@@ -386,13 +387,34 @@ pub async fn fetch_send<R: Runtime>(
...
+    channel: tauri::ipc::Channel<&[u8]>,
+) -> crate::Result<()> {
     let res = {
         let mut resources_table = webview.resources_table();
         resources_table.take::<ReqwestResponse>(rid)?
     };
+
     let res = Arc::into_inner(res).unwrap().0;
-    Ok(tauri::ipc::Response::new(res.bytes().await?.to_vec()))
+    let mut stream = res.bytes_stream();
```
amrbashir, line 398: can't we do the same streaming behavior using [`Response::chunk`](https://docs.rs/reqwest/latest/reqwest/struct.Response.html#method.chunk) method, instead of pulling another crate to iterate over the stream?

amrbashir (changes requested):
Thank you, could you also add a change file in `.changes` directory?

FabianLars:
positive results on discord: https://discord.com/channels/616186924390023171/1344666371316908063 (not rushing anyone, i still think the review comments need to be acknowledged)

livefantasia:
hello, is there any progress on this PR? Waiting for it to enable LLM streaming

FabianLars:
still waiting on reactions to the review comments 🤷

Issue #2092, opened 2024-11-25
[feature request] upload plugin, add option: skip certificate check


HuakunShen:
My app needs to download from a https server hosted locally with self signed certificate.

Can a `skipCertificate` option be added to the `download` API? or add an option to trust a given certificate?

PR #2165, opened 2024-12-09
feat(upload): added ssl cert options to upload and download function

HuakunShen:
Original Issue: https://github.com/tauri-apps/plugins-workspace/issues/2092

Add skip_ssl_cert_check and trust_ssl_cert options to upload and download function, to skip ssl check or trust a given self signed ssl cert.

Sample Code: 

```ts
 download(
      "https://localhost:9559/stream-file",
      "/Users/hk/Downloads/test.mp4",
      (progress) => {
        console.log(progress);
      },
      undefined,
      undefined,
      {
        skipSslCertCheck: true,
      }
    ))
```

My app (https://github.com/kunkunsh/kunkun) includes a local file transfer server. Self-signed cert is used for data encryption, but this plugin doesn't support self-signed cert or skipping SSL check.

Function signatures:

```ts
async function download(
  url: string,
  filePath: string,
  progressHandler?: ProgressHandler,
  headers?: Map<string, string>,
  body?: string,
  options?: {
    skipSslCertCheck?: boolean
    trustSslCert?: string
  }
): Promise<void>


async function upload(
  url: string,
  filePath: string,
  progressHandler?: ProgressHandler,
  headers?: Map<string, string>,
  options?: {
    skipSslCertCheck?: boolean
    trustSslCert?: string
  }
): Promise<string> 
```


HuakunShen:
Another suggestion: an `options` object field is added to the JS API `download()` and `upload()`, which prevents having to use undefined for not used variables. `headers` and `body` field should also be added to the `options` object. 
But to keep backward compatibility, I didn't move them in the PR to avoid breaking the API.

PR #2253, opened 2025-01-02
fix(opener): return error if path not exists

betamos:
From discussion on Discord. This PR checks that the path exists before shelling out to the "launcher". This helps with the common error that a file doesn't exist, and thus cannot be opened.

This is needed because we open files & urls "detached" in a separate thread, so we can't alert users if the action failed.

### DIFF
--- a/.changes/fix-opener-open-path-error.md
+++ b/.changes/fix-opener-open-path-error.md
@@ -1,0 +1,5 @@
+---
+"opener": patch
+"opener-js": patch
+---
+`open_path` now returns an error if the file does not exist
--- a/.changes/fix-opener-open-path-error.md
+++ b/.changes/fix-opener-open-path-error.md
@@ -2,4 +2,5 @@
 "opener": patch
 "opener-js": patch
 ---
-`open_path` now returns an error if the file does not exist
+
+Return an error in `open_path` if the file does not exist when opening with default application.
--- a/plugins/opener/src/open.rs
+++ b/plugins/opener/src/open.rs
@@ -53,5 +53,9 @@ pub fn open_url<P: AsRef<str>, S: AsRef<str>>(url: P, with: Option<S>) -> crate:
 /// ```
 pub fn open_path<P: AsRef<Path>, S: AsRef<str>>(path: P, with: Option<S>) -> crate::Result<()> {
     let path = path.as_ref();
+    if with.is_none() {
+        // Returns an IO error if not exists, and besides `exists()` is a shorthand for `metadata()`
+        _ = path.metadata()?;
+    }
     open(path, with)
 }


plugins/opener/src/open.rs
```diff
@@ -53,5 +53,7 @@ pub fn open_url<P: AsRef<str>, S: AsRef<str>>(url: P, with: Option<S>) -> crate:
 /// ```
 pub fn open_path<P: AsRef<Path>, S: AsRef<str>>(path: P, with: Option<S>) -> crate::Result<()> {
     let path = path.as_ref();
+    // Returns an IO error if not exists, and besides `exists()` is a shorthand for `metadata()`
+    _ = path.metadata()?;
```
amrbashir: Shouldn't we have two checks?
1. If `with` is `Some`, we check that one instead, as that's the one we are executing
2. If `with` is `None`, we check `path` as that's what we are trying to open

We could also apply the first check for `open_url`

However when we are checking for `with`, we need to take into account the `PATH`, right?

Legend-Master: > However when we are checking for with, we need to take into account the PATH, right?

It's not that simple actually, Windows's `ShellExecute` looks for installed apps from registry, so things like `firefox` would work even it's not in the `PATH`

betamos: Let's see. I think that when `with` is used, you still want to check `path`, which needs to exist to be meaningful at all. As such, checking path is still useful, and should not have false positives. So this would still help, even if it's not solving all problems.

To get full coverage of all meaningful errors (including something bad with `with`) we need to propagate the real error back, which isn't available with this detached design. This PR isn't fool-proof, but just addressing a common issue. For instance, my file transfer app references and creates files but the user very frequently move or remove those files out of bounds. I imagine it's similar for file editors, photo albums and similar software.

amrbashir: I can see a use-case where we shouldn't be checking the existence of a file, say we have an app called `generate-random-avatar` and it takes a non-existing path to save the generated avatar into. 

This can be solved by checking for path existence only if `with` is `None`.

betamos: Didn't think of that! Sounds like the conservative way to do it. Fixed.

amrbashir (dismissed):

betamos:
Thanks. Should be good to go. Let me know if there's something I need to do.

.changes/fix-opener-open-path-error.md
```diff
@@ -0,0 +1,5 @@
+---
+"opener": patch
+"opener-js": patch
+---
+`open_path` now returns an error if the file does not exist
```
amrbashir:
```suggestion

Return an error in `open_path` if the file does not exist when opening with default application.
```

amrbashir (approved):

PR #2322, opened 2025-01-14
feat(dialog): add `showFolderPicker` to iOS and android targets

Mo0nbase:
Add the folder picker to iOS and android. Ready for review!


FabianLars:
Awesome, thanks so much! I don't have any complains other than the CI. Your changes seem to work in my testing (though i didn't connect it to file reading/writing yet), nice work :)

Mo0nbase:
@FabianLars  Just tested file writing. The uri from the folder picker on ios works perfectly with the URI returned 

e.g. if there is a folder already created called Test in Download

```
const filePath = `${selected}/test.txt`;
await writeTextFile(filePath, 'This is a test file created by the app!');
```

In android however the URI does not work directly. If for example we select the same folder and do as we did before we get an invalid URI error.

e.g. content://com.android.externalstorage.documents/tree/primary%3ADownload%2FTest

In order to write to that folder this would need to be transformed to the following.

```
await writeTextFile("/storage/emulated/0/Download/Test/test.txt", 'This is a test file created by the app!');
```

This is more of a stylistic choice IMO and Im not sure if this is already an existing quirk of the android filesystem so I'm looking for some insight as to whether I should correct this.

Here is a repository if you are looking to reproduce.

https://github.com/Mo0nbase/test-dialog

FabianLars:
@lucasfernog Are you available in the near future to take a look at this? Really unsure on how to best connect it to the fs plugin and i'm afraid that we end up having to change the return value of this plugin so i'm a bit hesitant to merge it rn.

IT-ess:
Thank you for your work ! I hope we can merge it soon enough :innocent:

Issue #2277, opened 2025-01-07
Support multiple installer/bundle types per platform


FabianLars:
Now with the introduction of .deb updater support our current json schema really shows its limits. It would also be a valuable addition for Windows though because support msi and exe sometimes makes sense (or at least so i was told).

About the Static JSON response changes required:
Duplicate object keys are seemingly not well supported (understandibly since the specs highly discourage it) so i guess that's sadly out.

I feel like overloading the current object keys like `x86_64-windows-nsis` is a bit ugly but honestly not the worst.

Alternatively we could provide basically schema version 2 and use an array instead of an object, example:
```json
{
  // Maybe rename platforms to better reflect what the array actually contains
  "platforms": [
    {
      "os": "windows",
      "arch": "x86_64",
      "bundleType": "nsis",
    },
    {
      "os": "windows",
      "arch": "x86_64",
      "bundleType": "wix",
    }
  ]
}
```

FabianLars:
Starting to work on this. After a meeting with chip and tillmann it looks like the `x86_64-windows-nsis` approach makes the most sense. The logic i have in mind looks like this:

1) If dev sets `target` use this
2) If response only contains `x86_64-windows`, use this
3) If response contains x86_64-windows-nsis, use this for nsis installs
4) If response contains x86_64-windows-nsis and x86_64-windows use the first one for nsis and the latter one for wix
5) ~~I forgot~~

Any feedback on that?

One open question is: should it straight up fail if the response only contains x86_64-windows-nsis with the app being installed with wix or should the updater ignore the `-nsis` part if there's no other x86_64-windows/-* key and just install nsis over wix?
- imo it should fail with TargetNotFound

Legend-Master:
> If response contains `x86_64-windows-nsis` and `x86_64-windows` use the first one for nsis and the latter one for wix

Shouldn't it be NSIS checks for `x86_64-windows-nsis` and fallback to `x86_64-windows` while Wix checks for `x86_64-windows-wix` and fallback to `x86_64-windows`?

> should it straight up fail if the response only contains x86_64-windows-nsis with the app being installed with wix or should the updater ignore the -nsis part if there's no other x86_64-windows/-* key and just install nsis over wix?

An extra field containing a condition so the decision can be made from the new update instead of relying on the older updater would be nice, but I honestly hate logics written in custom languages so I don't know

[...]

PR #2325, opened 2025-01-15
feat(updater): Collect unknown response fields

### LINKED ISSUE #2275: custom key for updater  plugin
I found that it is impossible to add custom fields to the json file. Only those fields in the document are supported. This makes it impossible to do some custom functions, such as configuring forced updates.

FabianLars:
closes #2275

i was actually starting to work on #2277 🙃 

I'm not sure about the `other_response_fields` name - suggestions are very welcome


amrbashir (commented):
instead of calling it `other_response_fields`, let's call it `raw_json` and embed the whole JSON response and not just extra fields.

amrbashir (approved):

PR #2330, opened 2025-01-16
feat(window-state): add filter callback for excluding windows from tracking

### LINKED ISSUE #2329: [window-state] pattern based denylist
Hey!
I use window state in tauri v2 app.
In the app I have dynamic windows with dynamic labels since the app can have multiple duplicate windows (infinite).
Currently the plugin save the state of every dynamic window. I would like to disable it based on pattern rule.

For instance, I would like to disable state saving for every window with the label pattern `dialog_*`.
However, currently the window state plugin check by `contains()` instead of pattern matching. 

Can you add support for pattern based denylist, similar to how we have it in capabilities->windows property?

Thanks!

thewh1teagle:
Add filter callback for excluding windows from tracking for window state plugin 

Usage:

1. setup the plugin with

```rust
pub fn run() {
    /// in tauri.conf.json set window label to 'main' -> state saved -> change to 'main_deny' -> it won't be saved
    let plugin = tauri_plugin_window_state::Builder::new()
        .with_filter(|label| !label.ends_with("deny"))
        .build();
    ...

}
```

2. Check the state file in https://v2.tauri.app/reference/javascript/api/namespacepath/#configdir

Resolve
-  fix https://github.com/tauri-apps/plugins-workspace/issues/2329.

### DIFF
--- a/.changes/window-state-filter-callback.md
+++ b/.changes/window-state-filter-callback.md
@@ -3,4 +3,4 @@ window-state: patch
 window-state-js: patch
 ---
 
-Introduces a filter callback to exclude specific windows from saving their state. This allows for more flexibility by enabling dynamic exclusion of windows based on custom logic.
+Add `Builder::with_filter` callback to exclude specific windows from saving their state. This allows for more flexibility by enabling dynamic exclusion of windows based on custom logic.
--- a/.changes/window-state-filter-callback.md
+++ b/.changes/window-state-filter-callback.md
@@ -1,0 +1,6 @@
+---
+window-state: patch
+window-state-js: patch
+---
+
+Introduces a filter callback to exclude specific windows from saving their state. This allows for more flexibility by enabling dynamic exclusion of windows based on custom logic.
--- a/plugins/window-state/Cargo.toml
+++ b/plugins/window-state/Cargo.toml
@@ -29,4 +29,5 @@ serde_json = { workspace = true }
 tauri = { workspace = true }
 log = { workspace = true }
 thiserror = { workspace = true }
+glob = { workspace = true }
 bitflags = "2"
--- a/plugins/window-state/Cargo.toml
+++ b/plugins/window-state/Cargo.toml
@@ -29,5 +29,4 @@ serde_json = { workspace = true }
 tauri = { workspace = true }
 log = { workspace = true }
 thiserror = { workspace = true }
-glob = { workspace = true }
 bitflags = "2"
--- a/plugins/window-state/src/lib.rs
+++ b/plugins/window-state/src/lib.rs
@@ -352,7 +352,7 @@ impl Builder {
 
         let mut denylist_patterns = Vec::new();
         for pattern in denylist {
-            denylist_patterns.push(glob::Pattern::new(&pattern)?);
+            denylist_patterns.push(glob::Pattern::new(pattern)?);
         }
         self.denylist = denylist_patterns;
         Ok(self)
@@ -420,7 +420,6 @@ impl Builder {
                     .as_ref()
                     .map(|map| map(window.label()))
                     .unwrap_or_else(|| window.label());
-
 
                 for pattern in &self.denylist {
                     if pattern.matches(label) {
--- a/plugins/window-state/src/lib.rs
+++ b/plugins/window-state/src/lib.rs
@@ -321,7 +321,7 @@ impl<R: Runtime> WindowExtInternal for Window<R> {
 #[derive(Default)]
 pub struct Builder {
     denylist: HashSet<String>,
-    filter_callback: Option<Arc<Mutex<FilterCallbackFn>>>,
+    filter_callback: Option<Box<FilterCallbackFn>>,
     skip_initial_state: HashSet<String>,
     state_flags: StateFlags,
     map_label: Option<Box<LabelMapperFn>>,
@@ -358,7 +358,7 @@ impl Builder {
     where
         F: Fn(&str) -> bool + Send + Sync + 'static,
     {
-        self.filter_callback = Some(Arc::new(Mutex::new(filter_callback)));
+        self.filter_callback = Some(Box::new(filter_callback));
         self
     }
 
@@ -432,7 +432,6 @@ impl Builder {
 
                 // Check deny list callback
                 if let Some(filter_callback) = &self.filter_callback {
-                    let filter_callback = filter_callback.lock().unwrap();
                     // Don't save the state if the callback returns false
                     if !filter_callback(label) {
                         return;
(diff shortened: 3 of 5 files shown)


plugins/window-state/src/lib.rs
```diff
@@ -344,10 +346,16 @@ impl Builder {
     }
 
     /// Sets a list of windows that shouldn't be tracked and managed by this plugin
-    /// for example splash screen windows.
-    pub fn with_denylist(mut self, denylist: &[&str]) -> Self {
-        self.denylist = denylist.iter().map(|l| l.to_string()).collect();
-        self
+    /// For example, splash screen windows. It also supports glob patterns for flexible window matching.
+    pub fn with_denylist(mut self, denylist: &mut [&str]) -> Result<Self> {
```
amrbashir: this is a breaking change, let's add the globbing into a different function or simply ignore the glob errors and store and match against it as is

amrbashir (changes requested):

plugins/window-state/src/lib.rs
thewh1teagle: > this is a breaking change, let's add the globbing into a different function or simply ignore the glob errors and store and match against it as is

I'll change into

```rust
denylist_patterns.push(glob::Pattern::new(pattern).expect("Failed to parse glob pattern"));
```

Should I also change back the &mut [&str] denylist into &[&str]? (not sure how then sort it nicely)

thewh1teagle: Maybe

```rust
pub fn with_denylist(mut self, denylist: &[&str]) -> Self {
    let mut denylist = denylist.to_vec();
    denylist.sort();

    let mut denylist_patterns = Vec::new();
    for pattern in denylist {
        denylist_patterns.push(glob::Pattern::new(pattern).expect("Failed to parse glob pattern"));
    }
    self.denylist = denylist_patterns;
    self
}
```

Legend-Master:
Just a suggestion, instead of expending on the pattern, what about we allow an user provided function to customize this?

Maybe it's fine to just expand with glob patterns in this use case though

amrbashir:
> Just a suggestion, instead of expending on the pattern, what about we allow an user provided function to customize this?

We have a similar function `map_label` that is used when saving. But I think globs make sense here as well as it is similar to how you configure allowed/denied windows in the capability files.

plugins/window-state/src/lib.rs
amrbashir: Why do you need to sort?

Also please don't call `.expect()`, that's just a breaking change in disguise.

Instead, make two deny lists, one for globs and one for normal labels.  if the pattern is built successfully then append to patterns list, otherwise append to normal list.

thewh1teagle: > why do you need to sort?

In second thought, we don't need.


> also please don't call `.expect()`, that's just a breaking change in disguise.

What do you suggest? If returning a `Result` isn't an option, I'm not sure what else to do.

amrbashir: > Instead, make two deny lists, one for globs and one for normal labels. if the pattern is built successfully then append to patterns list, otherwise append to normal list.

amrbashir: actually I think we should go with two functions:

```rs
pub fn with_denylist(mut self, denylist: &[&str]) -> Self {
  // save to normal list
}

pub fn with_denylist_glob(mut self, denylist: &[&str]) -> Result<Self> {
  // save to glob list 
}
```

This would allow us to keep the old behavior and add a new API that can propagate the errors.

thewh1teagle: Added

Legend-Master: To be honest, I feel like if we're adding it in this way, a function might work better, the error handling here just feels wrong to me

amrbashir: why so?

thewh1teagle: > To be honest, I feel like if we're adding it in this way, a function might work better, the error handling here just feels wrong to me

The denylist patterns will run during every reload of the app in development. The chances of the production build having an error with it are small.

Legend-Master: I was thinking they would re-use the same function but then I realized they can't as that will be a breaking change

If we're adding in another function just for the globs, I feel like it's enough reason to let the user provide a function instead, this will also allow some more advanced use cases like only skip for next window creation as well

Also in most cases, I believe a check like `window_label.starts_with("some-name")` would probably be enough

Another thing is I find functions returning results like this is quite confusing in general, I always wondered about why would this function ever fail? It's an extra mental burden

I'm not against this though, just find myself preferring a function more

> ```rust
> tauri_plugin_window_state::Builder::new()
>     .with_denylist_glob(&["window-*"])
>     .unwrap()
>     .build()
> ```

thewh1teagle: > If we're adding in another function just for the globs, I feel like it's enough reason to let the user provide a function instead, this will also allow some more advanced use cases like only skip for next window creation as well

sounds better
How about 


```rust
    .with_filter(|label| true) // return true to save the state
    .build()
```

plugins/window-state/src/lib.rs
```diff
@@ -320,6 +321,7 @@ impl<R: Runtime> WindowExtInternal for Window<R> {
 #[derive(Default)]
 pub struct Builder {
     denylist: HashSet<String>,
+    filter_callback: Option<Arc<Mutex<FilterCallbackFn>>>,
```
Legend-Master: I think just `Option<Box<FilterCallbackFn>>` would be enough here?

plugins/window-state/Cargo.toml
```diff
@@ -29,4 +29,5 @@ serde_json = { workspace = true }
 tauri = { workspace = true }
 log = { workspace = true }
 thiserror = { workspace = true }
+glob = { workspace = true }
```
Legend-Master: We can remove this now

Legend-Master (changes requested):

plugins/window-state/Cargo.toml
thewh1teagle: Removed

plugins/window-state/src/lib.rs
thewh1teagle: Yes. Fixed.

.changes/window-state-filter-callback.md
```diff
@@ -0,0 +1,6 @@
+---
+window-state: patch
+window-state-js: patch
+---
+
+Introduces a filter callback to exclude specific windows from saving their state. This allows for more flexibility by enabling dynamic exclusion of windows based on custom logic.
```
amrbashir:
```suggestion
Add `Builder::with_filter` callback to exclude specific windows from saving their state. This allows for more flexibility by enabling dynamic exclusion of windows based on custom logic.
```

amrbashir (approved):

Legend-Master (approved):

Issue #2234, opened 2024-12-23
[bug] Open URL fails on Android/iOS with exception


velocitysystems:
When attempting to use `plugin-opener`to open a URL on Android/iOS, the following error(s) are thrown:

**iOS:**
`Unhandled Promise Rejection: No such file or directory (os error 2)`

**Android:**
`Uncaught (in promise) Permission denied (os error 13)`

All installation/usage instructions have been followed [here](https://www.npmjs.com/package/@tauri-apps/plugin-opener/v/2.2.2) and the below capabilities added:
- "opener:default",
- "opener:allow-open-url"

betamos:
Confirmed. This happens because the command calls `crate::open_url(url, with)` instead of the mobile plugin. It works on iOS if you use the rust bindings directly.

As a temp workaround I created a custom command which just calls `app.opener().open_url(..)` (can share if needed).

Not familiar with the guidelines, because there are a couple of ways of doing it, so won't bother with a PR until someone gives me a goahead, but here's what I'd suggest:

Simply do as my workaround and access the `app.opener()` directly in the command. That way, the cfg-directives can be kept to a minimum and `lib.rs` is left alone. Let me know, I'll send a PR.

FabianLars:
> Simply do as my workaround and access the app.opener() directly in the command. That way, the cfg-directives can be kept to a minimum and lib.rs is left alone. Let me know, I'll send a PR.

sounds reasonable, pr appreciated :)

velocitysystems:
Fixed with #2332.

satya1395:
i am having issue with opening remote video url with vlc. Is this realated ? 
Getting error: Failed to open video: "Not allowed to open url https://cdn.plyr.io/static/demo/View_From_A_Blue_Moon_Trailer-576p.mp4 with org.videolan.vlc"

NextJs 15 , Tauri 2 on MacOS 15, M4 processor.

```
// pages/index.js
"use client"
import { useState } from 'react'
import Head from 'next/head'

export default function Home() {
  const [isLoading, setIsLoading] = useState(false)
  const [error, setError] = useState(null)
  
  // Function to open video URL in VLC
  const openVideoInVLC = async () => {
    try {
      setIsLoading(true)
      setError(null)
      
        // Dynamically import the plugin only when needed
        const { openUrl } = await import('@tauri-apps/plugin-opener')
        
        // Static video URL - replace with your desired URL
[...]

PR #2332, opened 2025-01-19
fix(opener): invoke the mobile plugin in the commands too (#2234)

betamos:
Please double check this -- is accessing the plugin from commands the right way? Thinking if the plugin is not initialized..


betamos:
Btw, this is imo serious enough to deserve a version roll. It seems like it usually happens quickly, but just in case..

FabianLars:
> Thinking if the plugin is not initialized..

Then you can't invoke the plugin's commands anyway.

FabianLars (approved):

PR #2343, opened 2025-01-21
[opener] Prepare for objc2 frameworks v0.3

madsmtm:
The next version will have a bunch of default features enabled, see https://github.com/madsmtm/objc2/issues/627, so this PR pre-emptively disables them (so that your compile times down blow up once you upgrade to the next version).

I did not include a changelog entry here, since there are no user-facing changes, and since it doesn't need to be shipped to users.


amrbashir (approved):

amrbashir:
Thank you

PR #2377, opened 2025-01-31
Support BYO Logging

3lpsy:
# Changes
This feature adds a `.is_skip_logger` flag to the Log Plugin `Builder`, a `skip_logger()` method to the `Builder`, and logic to avoid acquiring (creating) a logger and attaching it to the global logger. Since `acquire_logger` is `pub`, a `LoggerNotInitialized` is added and returned if it's called when the `is_skip_looger` flag is set. 

# Why
This permits a crude BYO logging implementation as it avoids `attach_logger`. I have tracing setup with `tracing-log` and would just prefer to send the logs there. Additionally, without this change, you cannot create multiple Tauri mock apps for testing. Additionally, the end user can handle the logger but still use the plugin for easy access to the command and JS.

Please let me know if you do not think this approach is useful or if you would like me to make changes. Thanks!


plugins/log/src/lib.rs
```diff
@@ -339,6 +343,16 @@ impl Builder {
         self
     }
 
+    /// Skips logger acquisition and attachment. Certain functions such as split will be unusable.
```
FabianLars: can this contain a bit more info from the changelog? I think as is it's hard to understand what it does or rather why/when you'd want to use it.

3lpsy: I have updated the documentation.

3lpsy: @FabianLars Let me know if the changes are satisfactory. Happy to add more detail / reword it.

FabianLars: i think that's good, thanks

FabianLars (approved):

PR #2408, opened 2025-02-09
Update README.md

do-me:
Unfortunately it seems like fs doesn't work on iOS at the moment, see: 
- Github https://github.com/tauri-apps/tauri/issues/8843
- Discord https://discord.com/channels/616186924390023171/1293735921497276418


FabianLars:
This is not true. It's shit to work with, i give you that, but it still works at least a little bit. The apps i know of are all closed source with me not having access anymore so i can't give you an example right now, but this repo's example _almost_ works (almost because the capabilities aren't correct).

If the file isn't in the application specific directories you'll have to combine it with the dialog plugin though. I'll see if i can push a minimal fix (no time for more) to the plugin.

do-me:
Thanks a lot for taking the time, really appreciate it! 
It's just that so far I didn't see a single working repo with fs on iOS yet, so it's kind of frustrating to poke around in the dark.

FabianLars:
yeah i totally get it. it's frustrating for me as well because i have no clue about iOS. I should have pushed the people who worked on mobile support more to work on docs/examples as well 🙃

FabianLars:
oh damn that's a huge screenshot - obviously i don't have a clue about macOS as well 😂  - Edit: made it smaller...

do-me:
Awesome, thanks a lot! Could you share the link to this example source code?

FabianLars:
https://github.com/tauri-apps/plugins-workspace/tree/v2/examples/api or more specifically, this is what i used in the screenshot: https://github.com/tauri-apps/plugins-workspace/pull/2410

do-me:
I reproduced it on my physical device, so fs indeed seems to work! Thank you very much for taking the time.
However, against this comment, out of all App* paths on iOS, AppCache is the only path that is not writable (forbidden path) when trying to mkdir (https://github.com/tauri-apps/tauri/issues/8843#issuecomment-2306572951). 
AppData seems to be the safer bet here.

By the way, the UX on mobile is a little inconvenient. A resize function for the console height would be great. Also, it'd be better if there was no automatic zoom on text input. I fixed this before in another app but don't remember what kind of setting I had to tweak.

FabianLars:
> Also, it'd be better if there was no automatic zoom on text input. 

100% - that drove me crazy. And as you can see in the screenshot, i couldn't fix it either 🙃

do-me:
> > Also, it'd be better if there was no automatic zoom on text input.
> 
> 100% - that drove me crazy. And as you can see in the screenshot, i couldn't fix it either 🙃

If I remember correctly, a simple `event.preventDefault();` (see here https://github.com/do-me/tauri-basemap/blob/main/photon_geocode.js#L112) did the trick for me in this mini app: https://github.com/do-me/tauri-basemap

PR #2424, opened 2025-02-15
fix: Fix the issue of inconsistency between selected files and returned files.

### LINKED ISSUE #2089: [bug] Using plugin-dialog and plugin-fs to write file on iOS is ineffective.
Replicate the code：https://github.com/shell360/tauri-issue
The replication steps：
1. `git clone https://github.com/shell360/tauri-issue.git`
2. `cd tauri-issue && pnpm install`
3. run `pnpm tauri ios dev`，And select the iOS simulator.
4. click `save file` button
5. When checking the file in iOS file management, it was found that the content was not successfully written.

```tsx
import { useCallback, useState } from 'react'
import { save, open } from '@tauri-apps/plugin-dialog'
import { writeFile, readFile } from '@tauri-apps/plugin-fs'

function App() {
  const onSave = useCallback(async () => {
    const path = await save()
    if (!path) {
      return
    }

    let encoder = new TextEncoder()
    let data = encoder.encode('Hello World')
    await writeFile(path, data) // The content was not successfully written.
  }, [])

  const onOpen = useCallback(async () => {
    const path = await open()
    if (!path) {
      return
    }

    let content = await readFile(path)
    console.log(content)
  }, [])
  
  return (
    <main className="container">
      <button onClick={onSave}>save file</button>
      <button onClick={onOpen}>open file</button>
    </main>
  )
}

export default App

```

nashaofu:
* fix: #2089


plugins/dialog/ios/Sources/FilePickerController.swift
```diff
@@ -119,8 +119,7 @@ public class FilePickerController: NSObject {
 extension FilePickerController: UIDocumentPickerDelegate {
 	public func documentPicker(_ controller: UIDocumentPickerViewController, didPickDocumentsAt urls: [URL]) {
 		do {
-			let temporaryUrls = try urls.map { try saveTemporaryFile($0) }
```
nashaofu, line 122: The file path returned by saveTemporaryFile is inconsistent with $0, so the correct file cannot be obtained in the end.

PR #2428, opened 2025-02-15
chore: regenerate api-iife

Legend-Master:
I messed this up in #2420

@FabianLars should we put `pnpm-lock.yaml` into `.github/workflows/check-generated-files.yml` as well so we could prevent something like this in the future?


amrbashir (approved):

FabianLars:
> should we put pnpm-lock.yaml into .github/workflows/check-generated-files.yml as well so we could prevent something like this in the future?

sounds reasonable, yes

PR #2430, opened 2025-02-16
add `on_before_request` to `tauri-plugin-updater`

goenning:
This hook gives us much more flexibility when creating the Client for the HTTP request.

As an example, I personally need to do this:

```rust
.on_before_request(|b| b.use_preconfigured_tls(ClientConfig::with_platform_verifier()))
```


FabianLars:
Can we add a comment in the same spirit as this one? https://docs.rs/tauri/latest/tauri/webview/struct.WebviewWindow.html#method.with_webview
> Note that webview2-com, webkit2gtk, objc2_web_kit and similar crates may be updated in minor releases of Tauri. Therefore it’s recommended to pin Tauri to at least a minor version when you’re using with_webview.

I don't want us to be stuck with 0.11

goenning:
@FabianLars can you have a look now? I had to do some force-push juggling due to missing signed commits, but it should be ok now

Dmitry-Titov-K:
[FabianLars](https://github.com/FabianLars) check pls

FabianLars:
You forgot the `@` in front of my name. That's needed to actually ping me (assuming that's what you wanted to do judging by the link), without it it's just a normal message.

goenning:
It seems to have worked regardless 😅

@FabianLars any thoughts on the PR as is it? Am I missing something?

FabianLars:
> It seems to have worked regardless 😅

I read every message, just don't tell anyone that i'm actually just ignoring everything and everyone 🙃 

PR looks good i think and i meant to merge it already but ended up forgetting about it. Will take another look at merging it tomorrow (i don't trust the mobile preview so can't do it now)

plugins/updater/src/updater.rs
```diff
@@ -239,6 +242,19 @@ impl UpdaterBuilder {
         self
     }
 
+    /// Allows you to modify the `reqwest`` client builder before the HTTP request is sent.
+    ///
+    /// Note that `reqwest` crate may be updated in minor releases of tauri-plugin-updater.
+    /// Therefore it's recommended to pin the plugin to at least a minor version when you're using `on_before_request`.
+    ///
+    pub fn on_before_request<F: Fn(ClientBuilder) -> ClientBuilder + Send + Sync + 'static>(
```
amrbashir: I think this can be named `configure_client` and we should call it when downloading the update as well  not just when checking.

goenning:
Renamed to `configure_client` + added to download.

plugins/updater/src/updater.rs
```diff
@@ -239,6 +242,19 @@ impl UpdaterBuilder {
         self
     }
 
+    /// Allows you to modify the `reqwest`` client builder before the HTTP request is sent.
```
```suggestion
    /// Allows you to modify the `reqwest` client builder before the HTTP request is sent.
```

.changes/updater-add-on-before-request.md
```diff
@@ -0,0 +1,6 @@
+---
+"updater": minor
+"updater-js": minor
+---
+
+Add `configure_client` hook to the updater plugin.
```
amrbashir:
```suggestion
Add `UpdaterBuilder::configure_client` method on Rust side, to configure rhe `reqwest` client used to check and download the update.
```

amrbashir:
Should we include the same method on the plugin Builder as well?

.changes/updater-add-on-before-request.md
```diff
@@ -0,0 +1,6 @@
+---
+"updater": minor
+"updater-js": minor
+---
+
+Add `UpdaterBuilder::configure_client` method on Rust side, to configure rhe `reqwest` client used to check and download the update.
```
lucasfernog:
```suggestion
Add `UpdaterBuilder::configure_client` method on Rust side, to configure the `reqwest` client used to check and download the update.
```

amrbashir (approved):

PR #2434, opened 2025-02-17
fix(updater): panic when updating with empty current_exe_args

### LINKED ISSUE #2335: [bug] Update fails on Windows when constructing the updater using the UpdaterBuilder directly
I'm using the `UpdaterBuilder` to construct an `Updater` and initiate an update but I'm getting a panic error: `range start index 1 out of range for slice of length 0` in `tauri-plugin-updater-2.3.1\src\updater.rs:616:52`.

It seems to be because the `UpdaterBuilder::current_exe_args` is empty and it's not possible to change it from outside the crate.

The reason I'm using the `UpdaterBuilder` directly and not using `UpdaterExt` is because I only want to use the update functionality but I'm otherwise not using tauri and I don't want to construct a `Runtime`. Is this kind of use not supported?

talzion12:
fix: https://github.com/tauri-apps/plugins-workspace/issues/2335

### DIFF
--- a/plugins/updater/src/updater.rs
+++ b/plugins/updater/src/updater.rs
@@ -632,12 +632,11 @@ impl Update {
 
         let install_mode = self.config.install_mode();
         let current_exe_args = self.current_exe_args();
-        let current_args =
-            current_exe_args
-                .split_first()
-                .map(|(_, args_without_exe)| args_without_exe)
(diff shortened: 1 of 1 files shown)


amrbashir (commented):
Is there more work need to be done in this PR and that is why it is in draft?

PR #2440, opened 2025-02-20
fix(barcode-scanner): fix the cameraView is not removed after scanning in iOS

kingsword09:
On `iOS`, when `windowed = false`, the `cameraView` is not removed from the superView when the scanning is done, making the interface unclickable.

image

When destroy, only the `removePreviewLayer` was done, not the `removeFromSuperview`.
image


zepyrshut:
Thank you so much!!! Fixed my problem!! The barcode scanner fully working with iPhone with `windowed: false`

I copied the plugin folder (barcode-scanner) and fetched it locally

kingsword09:
@lucasfernog Is it possible to review the code and release a version if no problem, otherwise everyone who uses this plugin will have to copy a copy locally.

FabianLars (approved):
counting the new duplicate pr as some sort of approval

Issue #2306, opened 2025-01-11
[feat] [biometric] Allow using biometric authentication to encrypt/decrypt sensitive data


charlesschaefer:
It would be great to biometric plugin to allow to encrypt/decrypt sensitive data using the biometric authentication. 
I.e. let's think of an app that encrypts some user's sensitive data using a secret key generated based on a password provided by the user. (password-based-key) To access the data, the user must always type the password (that we can use to regenerate the key and decrypt the data).

To improve the experience, the app can store the password-based-key somewhere and, when the user authenticates using the biometric plugin, it reads the key and uses it to decrypt the needed data. But this is insecure because the password-based-key would be stored in a way that someone could read it.

In android (I don't know about iOS), we can create a [CryptoObject](https://developer.android.com/identity/sign-in/biometric-auth#auth-per-use-keys) that can use biometric data to encrypt some data. We could, then, use the `cipher` provided by Android to encrypt our password-based-key (ensuring it won't be readable by anyone) and we would be able to decrypt it only when the biometric authentication is successful.  And if the user cancels the biometric authentication (or registers a new biometric credential), we can just delete the key (because we can regenerate it if the user provides the correct password).

To allow the code to be consistent between platforms, the plugin could receive a modified AuthOptions with fields to configure to use the encryption and a flag to tell if it must encrypt (and the data to be encrypted) or decrypt the data. And the `Result` returned by the `authenticate()` could be filled with the data (encrypted or decrypted). Thus, the plugin consumer won't need to deal with different APIs depending on the platform.

Do you think this can be implemented in a way that can be used both on Android and iOS?

charlesschaefer:
Ok, so I've started to check if what I thought to use in android side would work. And it is working pretty well, just needing to change the possible authentication types from WEAK to STRONG when using the biometric auth to get a key to encrypt/decrypt some data.

Is it ok if I finish the work on the plugin and open a PR to implement this feature on Android, without implementing it on iOS? I know that having different features on each platform, even when the platform has what is needed to implement them (I'm not sure if iOS has, although) is a very bad thing. But I don't have any kind of Mac equipment and never worked with iOS environment/techs before. 

If it is not acceptable, I can create an unofficial plugin that provides the feature.

FabianLars:
> Is it ok if I finish the work on the plugin and open a PR to implement this feature on Android, without implementing it on iOS?

[...]

PR #2454, opened 2025-02-23
[feat] [biometric] Adding the capability of encrypting/decrypting data using biometric authentication on Android (Ref. #2306)

charlesschaefer:
I've used the current structure of the biometric plugin to use Android's capability of encrypting/decrypting data using a key that can only be accessible when the user successfully authenticates with their registered biometric methods.

More details on the need of this change can be found on the #2306 issue. 

Summary of changes: 
**Kotlin**
- Created the command `biometricCipher` that will be responsible to encrypt/decrypt data by asking the user for their biometric authentication and, when successful, get a key from android device to make the cryptography process.
- Extracted some common code to their own functions.
- Changed the AuthOptions class to include the fields dataToEncrypt and dataToDecrypt
- Changed the force of authentication methods when using cryptography. Android allows using this feature only when using STRONG methods. When not using cryptography, we'll keep allowing the use of WEAK methods.

**JS**
- Added the method biometricCipher
- Changed the AuthOptions interface to include the fields dataToEncrypt and dataToDecrypt

**Rust**
- Changed the AuthOptions struct to include the fields data_to_encrypt and data_to_decrypt
- Added the CipherResult struct
- Added the method Biometric.biometric_cipher() to access the similar method on Kotlin's side.

Some implementation notes:

- I've used a fixed crypto algorythm instead of allowing the consumer to define it (AED with GCM block mode).
- I detect that the consumer wants to encrypt/decrypt data only by the presence of the fields dataToEncrypt/dataToDecrypt on the AuthOptions object sent by them.
- The flow is: the consumer can use the authenticate command as usual (if doesn't want to use cryptography) or can use the biometricCipher command (both after checking with the status command) passing the raw data to encrypt or the encrypted data to decrypt.
- If the key used before to encrypt some raw data is not valid anymore, it will return an error (using the same authenticateResult method that is used with authenticate command).

I'll need some guidance about where to change the site's documentation to include this new method.


charlesschaefer:
Hello guys, anyone available to review this?

surkaa:
I am looking forward to this feature! Currently, the biometric plugin lacks the ability to handle cryptographic operations, which is essential for secure credential storage. This PR solves a major pain point for mobile security in Tauri.

surkaa (approved):

zz1254 (approved):

PR #2479, opened 2025-03-02
HTTP add stream support

adrieljss:
This PR was meant to add stream support for `fetch`. 

What was changed: 

Rust:
- Removed `fetch_read_body`, because we're already sending response body chunks through the IPC.
- Using `reqwest::chunk` to send an array of bytes (`Vec<u8>`) to JavaScript using the IPC `Channel`

JavaScript:
- return a `ReadableStream` as `InitBody` in `Response`, this should be automatically collected and function as normal even if not used for streaming responses.

## Example Usage
```js
const res = await fetch('...')
const rd = res.body.getReader()

while (true) {
  const { done, value } = await rd.read()
  if (done) break
  const txt = new TextDecoder().decode(value)
  console.log(txt)
}
```

or (TypeScript might not like this)
```js
const res = await fetch('...')

for await (const chunk of res.body) {
  const txt = new TextDecoder().decode(chunk)
  console.log(txt)
}
```

closes #2140


FabianLars:
Thanks for opening a PR! Tiny bit more context for reviewers: https://discord.com/channels/616186924390023171/1344666371316908063 This is supposed to be an alternative to #2140

@adrieljss Can you please revert the changes to Cargo.toml, Cargo.lock, package.json, and pnpm-lock.yaml ? Versioning is handled in CI and this would break stuff.

plugins/http/src/commands.rs
```diff
@@ -126,6 +129,12 @@ pub struct BasicAuth {
     password: String,
 }
 
+#[derive(Clone, Serialize)]
+pub struct StreamMessage {
+    value: Option<Vec<u8>>,
+    done: bool,
+}
+
```
amrbashir: This is not performant because of serialization overhead. 

To avoid this, you can send a `tauri::Response` which is just a `Vec<u8>` and then you could append a single byte at the end that is either 1 or 0, indicating the status of done. Or even an empty array would suffice.

I am not sure if the channel is as performant as invoke (will have to check that later) but would be better to avoid using the channel.

adrieljss: Yes you're right, I will change the struct, how about using [serde_bytes](https://docs.rs/serde_bytes/latest/serde_bytes/)? Since serde could be serializing the Vec<u8> as an array of numbers.

I don't think there's any way to do streaming using invoke(?) so maybe a solution would be having a seperate function for seperate usecases, one for streaming and one for collecting all the contents at the end, but the user has to specify that and that will not be `fetch`-like.

adrieljss: Though, if the server doesn't want to stream, the channel will only receive 1 chunk and that should be (around) the same performance since both uses IPC anyways.

amrbashir: this can be done with `invoke` and an `AsyncIterator` in JS:
- The removed `fetch_read_body` instead of using `.bytes()` it will use `.chunk()`, effictively returning a single chunk each time it is called
- On JS side, you can create an `AsyncIterator` that will be used to create the `ReadableStream`, on each iteration it will call `fetch_read_body` and append the data to the stream

adrieljss: Would that be better though? In [channels](https://tauri.app/develop/calling-rust/#channels), it said that is the recommended way for streaming data (like HTTP streaming, which is our case) to the frontend. I think the `invoke` approach will block operations until it has returned something. While `channels` is just an event listener.

In my opinion, I feel like it's more intuitive rather than spamming invokes

amrbashir: The `channel` calls `invoke` under the hood. Channels are more ergonomic but there is room for improvement in their implementation as it uses `tauri::State` with a `Mutex` unlike what I am proposing which would use a Mutex only until the request is sent, after that the body could be read without any Mutexes.

For now, let's use the channel approach and I can improve on it later (once I confirm that there is mutex overhead involved).


My main blocker for this PR is the use of serialization which we can avoid even with channels.

adrieljss: Yes okay, I will fix that 👍

adrieljss: @amrbashir can you check the new commit I did? I can't use `tauri::Response` inside the channel because it doesn't impl Clone and the compiler doesn't allow me to. I used [serde_bytes](https://github.com/serde-rs/bytes) to try and make it as efficient as possible, instead of just sending an array of numbers.

amrbashir: That clone bound shouldn't be needed, I will open a PR to fix it later. 
Edit: here it is https://github.com/tauri-apps/tauri/pull/12876

For now, you can use `InvokeResponseBody::Raw` which is `Clone` and implements `IpcResponse` trait and so can be used with channels.

adrieljss: > That clone bound shouldn't be needed, I will open a PR to fix it later.
> 
> For now, you can use `InvokeResponseBody::Raw` which is `Clone` and implements `IpcResponse` trait and so can be used with channels.

done! 👍

plugins/http/src/commands.rs
```diff
@@ -22,6 +23,8 @@ use crate::{
 
 const HTTP_USER_AGENT: &str = concat!(env!("CARGO_PKG_NAME"), "/", env!("CARGO_PKG_VERSION"),);
 
+// reqwest::Response is never read, but might be needed for future use.
+#[allow(dead_code)]
```
amrbashir: let's remove, no need to keep it around anymore

plugins/http/guest-js/index.ts
```diff
@@ -186,6 +186,27 @@ export async function fetch(
     throw new Error(ERROR_REQUEST_CANCELLED)
   }
 
+  const streamChannel = new Channel<Uint8Array>()
+
+  const readableStreamBody = new ReadableStream({
+    start: (controller) => {
+      streamChannel.onmessage = (res: Uint8Array) => {
```
amrbashir: this should be an `ArrayBuffer` iirc, and sometimes in the case where IPC fails to use the borwser `fetch`, it fallsback to Json serialization so this would be `Array<number>`

So we need to do some checks here, similar as what was done before

adrieljss: Yes, you're right. I'll update that.

adrieljss:
Happy to say the stream that was implemented worked:
image

Something is wrong when i do content processing (like `.json(), .text(), etc`) so i'm trying to figure it out.

adrieljss:
@amrbashir I added a new commit, can you check? I also fixed the content conversion bug said above, by converting the `ArrayBuffer | number[]` to `Uint8Array` (the content reader/consumer on `Response` doesn't like non-`Uint8Array`s). So far it works fine when I tested it.

adrieljss:
I also removed ReqwestResponse completely [acafdd8](https://github.com/tauri-apps/plugins-workspace/pull/2479/commits/acafdd84d87f33379876e8a85c09690be8d5246b), it's never read on Rust and is not needed in guest-js.

amrbashir:
LGTM, now you just need to run `cargo fmt` and `pnpm format`, and a change file in `.changes` directory and we should be good to go.

adrieljss:
Done!

just a heads up,
image
`Channel` might not be supported on the browser (`http://localhost:xxxx`), but when ran as an application it was fine.

Maybe should just fallback to `fetch` when in a browser setting?

amrbashir:
that's out of the plugin scope users will need to add check and fallback if they wish so.

amrbashir:
looks like you need a final `pnpm build` to build the JS API

DresAaron:
Really appreciate the great work! Any plans to release this feature soon?

letscagefdn:
good job guys, when do you think this will be released ?

lucasfernog (approved):
@amrbashir i'll approve and merge this one, it works well
i think any further change should come after a change in tauri core too right?

livefantasia:
@adrieljss @amrbashir 

Thanks for the update. But while I can see the chunks by putting logs in the  streamChannel.onmessage callback in the js http client, but the fetch call doesn't return until the full response is collected.
I have a look at the rust code: 
- In the fetch function, the future ( fut ) processes the entire response stream, sending chunks through the IPC channel
- The fetch_send function waits for this future to complete with fut.as_mut() before returning the response metadata

On the client side, the client js is blocked here until the full response is received 
const {
    status,
    statusText,
    url,
    headers: responseHeaders
  } = await invoke<FetchSendResponse>('plugin:http|fetch_send', {
    rid
  })

So you can't use the standard pattern like to get streaming object: 
const { value, done } = await reader.read()

The implementation in https://github.com/tauri-apps/plugins-workspace/pull/2140 doesn't have this issue, as the you fetch_send only needs the header to return and fetch_read_body can be invoked async

adrieljss:
@mobilexiang yes you're right, going to send out a fix ASAP

PR #2483, opened 2025-03-03
fix(deep-link): local machine registry handling

### LINKED ISSUE #2462: [bug][deep-link] `is_registered` does not behave as expected for windows
I build my software in multi-user mode (`bundle.windows.nsis.installMode` = `both`). After installation, when the program starts (launched by the installer with administrator privileges), the JavaScript code executes `register("opener")`;.

I checked the registry and found that the key `HKEY_CLASSES_ROOT\opener\shell\open\command` was registered. However, when another user starts the program and calls `isRegistered("opener")`;, it fails to detect that "opener" has already been registered.

After reading the relevant code, I noticed that all registry operations related to deep-link registration are performed under `HKEY_CURRENT_USER`. I am not sure why this happens, but there may be an issue with custom protocol registration in multi-user mode.

Legend-Master:
- Fix `unregister` not being able to remove deep link registered in `HKEY_LOCAL_MACHINE` on Windows
- Fix `is_registered` not being able to pickup deep link registered in `HKEY_LOCAL_MACHINE` on Windows (fixes #2462)

### DIFF
--- a/plugins/deep-link/src/lib.rs
+++ b/plugins/deep-link/src/lib.rs
@@ -172,7 +172,7 @@ mod imp {
     use tauri::Manager;
     use tauri::{AppHandle, Runtime};
     #[cfg(windows)]
-    use windows_registry::CURRENT_USER;
+    use windows_registry::{CLASSES_ROOT, CURRENT_USER, LOCAL_MACHINE};
 
     /// Access to the deep-link APIs.
     pub struct DeepLink<R: Runtime> {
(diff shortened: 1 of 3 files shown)


plugins/deep-link/src/lib.rs
```diff
@@ -398,10 +407,8 @@ mod imp {
         pub fn is_registered<S: AsRef<str>>(&self, _protocol: S) -> crate::Result<bool> {
             #[cfg(windows)]
             {
-                let cmd_reg = CURRENT_USER.open(format!(
-                    "Software\\Classes\\{}\\shell\\open\\command",
-                    _protocol.as_ref()
-                ))?;
+                let protocol = _protocol.as_ref();
+                let cmd_reg = CLASSES_ROOT.open(format!("{protocol}\\shell\\open\\command"))?;
```
lucasfernog:
```suggestion
                let Ok(cmd_reg) = CLASSES_ROOT.open(format!("{protocol}\\shell\\open\\command"))
                else {
                    return Ok(false);
                };
```

lucasfernog (approved):

PR #2505, opened 2025-03-07
fix(window-state): The window size state needs to be related to the screen's scale_factor.

muwoo:
The window size state needs to be related to the screen's `scale_factor`. Otherwise, if the external monitor resolution is inconsistent, the restored window size will be scaled relative to the previous state.

### DIFF
--- a/.changes/window-state-size-scale-factor.md
+++ b/.changes/window-state-size-scale-factor.md
@@ -1,0 +1,5 @@
+---
+window-state: patch
+---
+
+fix: The window size state needs to be related to the screen's scale_factor, Otherwise, the restored window size will be scaled relative to the previous state.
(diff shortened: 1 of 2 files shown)


Legend-Master:
This was changed in #1743 deliberately for changing scale factors on the same resolution screen

I feel like there's no right answer to this when the resolution and scaling are changing, we either break on resolution change or on scaling change 😥

I'm thinking maybe we should check for the monitor and see if we're still on the same monitor, and don't restore the size if that have changed, like what we did for the position

muwoo:
Thank you for your answer. I just completed the migration from version 1.x to 2.x and found that the window size status was inconsistent with the 1.x version. I mistakenly thought that the code was not implemented.😂

PR #2507, opened 2025-03-07
fix(clipboard-manager): Wayland support

### LINKED ISSUE #2498: tauri_plugin_clipboard_manager: Unavailable Wayland support
On Linux with Wayland, the clipboard content cannot be read or wrote, or even shared between most apps.  This is related to Wayland, using a different clipboard protocol than x11, resulting in an incompatibility. If this is not a desired behavior with the plugin, we can fix this by enabling a feature from `arboard` to automagically support the Wayland protocol. 

I'm able to make the PR if needed.

Squitch1:
* fix: #2498

### DIFF
--- a/plugins/clipboard-manager/Cargo.toml
+++ b/plugins/clipboard-manager/Cargo.toml
@@ -35,5 +35,8 @@ thiserror = { workspace = true }
 [target.'cfg(target_os = "ios")'.dependencies]
 tauri = { workspace = true, features = ["wry"] }
 
-[target."cfg(any(target_os = \"macos\", windows, target_os = \"linux\", target_os = \"dragonfly\", target_os = \"freebsd\", target_os = \"openbsd\", target_os = \"netbsd\"))".dependencies]
+[target."cfg(any(target_os = \"macos\", windows))".dependencies]
 arboard = "3"
+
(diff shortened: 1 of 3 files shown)


plugins/clipboard-manager/Cargo.toml
```diff
@@ -35,5 +35,8 @@ thiserror = { workspace = true }
 [target.'cfg(target_os = "ios")'.dependencies]
 tauri = { workspace = true, features = ["wry"] }
 
-[target."cfg(any(target_os = \"macos\", windows, target_os = \"linux\", target_os = \"dragonfly\", target_os = \"freebsd\", target_os = \"openbsd\", target_os = \"netbsd\"))".dependencies]
+[target."cfg(any(target_os = \"macos\", windows))".dependencies]
 arboard = "3"
+
+[target."cfg(any(target_os = \"linux\", target_os = \"dragonfly\", target_os = \"freebsd\", target_os = \"openbsd\", target_os = \"netbsd\"))".dependencies]
+arboard = { version = "3", features = ["wayland-data-control"] }
```
FabianLars: is there any real world reason to split it up per platform for feature flags?

Squitch1: Apart from highlighting the feature's target, there is no real reason to split it up. I can merge the two lines if it seems more clear for you.

FabianLars: yeah, i think i'd prefer it as one dependency tbh

FabianLars (approved):

PR #2513, opened 2025-03-08
feat(updater): improve tracing and error logging

lucasfernog:


amrbashir (approved):

plugins/updater/src/updater.rs
```diff
@@ -391,24 +394,38 @@ impl Updater {
...
-                        raw_json = Some(res.json().await?);
-                        match serde_json::from_value::<RemoteRelease>(raw_json.clone().unwrap())
+                        let update_response: serde_json::Value = res.json().await?;
+                        log::debug!("update response: {update_response:?}");
+                        raw_json = Some(update_response.clone());
+                        match serde_json::from_value::<RemoteRelease>(update_response)
                             .map_err(Into::into)
                         {
                             Ok(release) => {
+                                println!("parsed release response {release:?}");
```
Legend-Master, line 408: This should be `log::debug` as well?

lucasfernog: ohhh

plugins/updater/src/updater.rs
```diff
@@ -391,24 +394,38 @@ impl Updater {
...
-                            Err(err) => last_error = Some(err),
+                            Err(err) => {
+                                log::error!("failed to deserialize update response: {err}");
+                                last_error = Some(err)
+                            }
                         }
+                    } else {
+                        log::error!(
+                            "update endpoint did not respond with a successful status code"
+                        );
```
beanow-at-crabnebula, line 422: Is this case missing a `last_error`?

Users are reporting that 400 errors are shown as "invalid json", even though the response should be hitting this success status check, and the body is still a valid json string.

FabianLars: looks like it

PR #2522, opened 2025-03-12
fix(http): Fixes Stream Bug

adrieljss:
The problem was that the stream while sending was blocking the await function, not returning `ReadableStream` and the headers after sending.

This was fixed by (kind of) bringing back the `fetch_read_body`, I actually wanted to name it `fetch_stream_body`, but I don't know how to set up the permissions.

On ReadableStream start, it will call a different function (`fetch_read_body`) as to not block the return of the headers on `fetch_send`.

Fixes https://github.com/tauri-apps/plugins-workspace/pull/2479#issuecomment-2712762909

@amrbashir


amrbashir:
Can you please fix the conflicts? 


And for future PRs, it is recommended not make PRs from the default branch as it most certainly will cause conflicts. You should always make a new branch for each PR.

adrieljss:
Whoops
sorry I didn't know, I will resolve them soon. Thank you for the tip👍.

plugins/http/src/commands.rs
```diff
@@ -370,7 +359,31 @@ pub fn fetch_cancel<R: Runtime>(webview: Webview<R>, rid: ResourceId) -> crate::
     Ok(())
 }
 
-#[tauri::command]
+#[command]
+pub async fn fetch_read_body<R: Runtime>(
+    webview: Webview<R>,
+    rid: ResourceId,
+    stream_channel: Channel<tauri::ipc::InvokeResponseBody>,
+) -> crate::Result<()> {
```
amrbashir: Let's move this command below fetch_send so it matches the order they are called on the JS side

plugins/http/src/commands.rs
```diff
@@ -370,7 +359,31 @@ pub fn fetch_cancel<R: Runtime>(webview: Webview<R>, rid: ResourceId) -> crate::
     Ok(())
 }
 
-#[tauri::command]
+#[command]
+pub async fn fetch_read_body<R: Runtime>(
```
amrbashir: This also needs to be add in lib.rs

amrbashir (changes requested):

amrbashir:
Also you need to run pnpm build

adrieljss:
> Also you need to run pnpm build

I already ran it in the previous commits so it didn't change anything

FabianLars:
is your local `node_modules` instance outdated maybe? (or did you maybe just not commit the dist-js files?)

adrieljss:
> is your local `node_modules` instance outdated maybe? (or did you maybe just not commit the dist-js files?)

image
I think the compiled js is already correct, the build function doesnt change anything, I must have built it and commited while I was testing.

adrieljss:
I think I'm just going to make a new branch and a new PR with the latest fork, this has been really messy, sorry.

FabianLars:
i can push the diff for you if you prefer? then you don't have to recreate the branch

adrieljss:
> i can push the diff for you if you prefer? then you don't have to recreate the branch

Nevermind, I ran it twice and then it changed this time, so weird.

amrbashir (dismissed):

amrbashir (approved):

PR #2524, opened 2025-03-12
enable TargetKind::LogDir on mobile

elwerene:
It's not possible to do a workaround on mobile with `TargetKind::Folder` as it's not possible to access the `AppHandle` to get a writable location and I didn't want to copy that code into my project.

I tested it on Android, but it should also work on iOS.

Do you see this as a breaking change because it generates logs on mobile apps with the default setting when it didn't do that before? I could change the defaults for mobile, if that change is considered breaking.

### DIFF
--- a/.changes/change-pr-2524.md
+++ b/.changes/change-pr-2524.md
@@ -1,0 +1,6 @@
+---
+"log": patch
+"log-js": patch
+---
+
+enable TargetKind::LogDir on mobile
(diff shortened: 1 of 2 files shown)


FabianLars:
> as it's not possible to access the AppHandle to get a writable location and I didn't want to copy that code into my project.

it is if you register the plugin in the setup hook instead of the tauri builder but for a logger that's a bit weird i guess.

> Do you see this as a breaking change

no

FabianLars (approved):
tested on ios as well

elwerene:
@FabianLars Thanks for the very quick test, merge and release!

PR #2548, opened 2025-03-18
fix(dialog): do not create file copy for save file picker on iOS

### LINKED ISSUE #2089: [bug] Using plugin-dialog and plugin-fs to write file on iOS is ineffective.

nashaofu:
On iOS the file picker returns a security scoped resource file path on the save() file picker: https://developer.apple.com/documentation/uikit/uidocumentpickerviewcontroller?language=objc#Work-with-external-documents

this means we can't directly access it without calling [startAccessingSecurityScopedResource](https://developer.apple.com/documentation/foundation/nsurl/1417051-startaccessingsecurityscopedreso?language=objc)

this PR changes the plugin to not access the save file early, leaving that to the user and returning its actual path

fix #2089


lucasfernog:
nice catch. I think we still need some improvements here, for instance now I couldn't access the file without https://developer.apple.com/documentation/foundation/nsurl/1417051-startaccessingsecurityscopedreso?language=objc

I think we should add that to the FS plugin too

nashaofu:
Add two methods to fs `startAccessingScopedResource` and `stopAccessingScopedResource`? Or I hope to automatically add these two methods before and after the fs call

nashaofu:
The open and save methods now return Path objects. These objects must be manually destroyed after use.

PR #2550, opened 2025-03-19
fix(fs): convert async iterator syntax to manual read

tjramage:
The `writeFile` function accepts a `ReadableStream`, however, under macOS this does not work since the built-in webview does not have async iterator support. See Safari's [compatibility](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream#browser_compatibility) for more info.

This PR converts the code to a manual read while-loop instead, fixing support for macOS.

_**Note:**_ I have not tested this across all platforms. Perhaps @amrbashir could check this, if possible? 🙏


plugins/fs/guest-js/index.ts
```diff
@@ -1075,9 +1075,18 @@ async function writeFile(
...
+      while (true) {
+        const { done, value } = await reader.read()
+        if (done) break
+        await file.write(value)
+      }
+    } finally {
+      reader.releaseLock()
     }
+  
     await file.close()
```
amrbashir: I think we should propagate the errors and not ignore them
```suggestion
    try {
      while (true) {
        const { done, value } = await reader.read()
        if (done) break
        await file.write(value)
      }
    } catch(e) {
      reader.releaseLock()
      await file.close()
      throw e
    }
```

amrbashir (changes requested):

amrbashir:
Also could you add a change file in `.changes` directory? see https://github.com/tauri-apps/plugins-workspace/blob/v2/.changes/persist-cookies.md?plain=1 for an example.

Also you need to run `pnpm build` and `pnpm format`

Thank you

plugins/fs/guest-js/index.ts
tjramage: @amrbashir - with a `try / finally` block, errors will still propagate as expected, no? Also, I think we should release the lock and close the file handle, regardless of an error, so it seems like `finally` is the right place for this. Let me know your thoughts.

```js
function test() {
  try {
    throw new Error("Oops!");
  } finally {
    console.log("Finally block runs!");
  }
}

test(); // Logs "Finally block runs!", then throws "Oops!"
```

amrbashir: this does work indeed, thank you

amrbashir (approved):

amrbashir:
Thank you

Issue #1705, opened 2024-08-29
Sql plugin - Sqlite — unable to load an extension


ezzaouia:
How to load an extension for sqlite?

For example the sql below does not seem to work, I'm getting `not authorized` (in the console):

```
SELECT load_extension ('spellfix');
```

lutfi-haslab:
Same, i am trying to load extension vector with sqlite-vec, not work. It support rust & js, maybe tauri can considering adding loadExtension property

```
const initDB = async () => {
   console.log('initDB')
   const db = await Database.load('sqlite:test.db');
   sqliteVec.load(db);
   console.log('db', db)
}
```

```
Argument of type 'Database' is not assignable to parameter of type 'Db'.
  Property 'loadExtension' is missing in type 'Database' but required in type 'Db'.ts(2345)
index.d.ts(10, 5): 'loadExtension' is declared here.
```

PR #2553, opened 2025-03-20
feat(sql): support for SQLite pragma and encryption pragmas

### LINKED ISSUE #7: [sql] SQLCipher Encryption
Hi and thank you for everything you are doing. I just started using Tauri (+ this plugin) for a side Angular project and results were amazing so far.

One thing left to figure out, my app will be downloadable for free, with the condition to only use it in a single machine. My logical thinking for it is to encrypt the database using some per-device unique ID. So I was wondering if is it possible to use SQLCipher. Digging a bit into sqlx, I found this thread linking a great sample repo: https://github.com/launchbadge/sqlx/issues/2009

It seems possible by overriding libsqlite3-sys dependencies to use bundled SQLCipher: 

```rust
# optional = false - force to build the package with sqlcipher feature enabled
libsqlite3-sys = { version = "0.24.1", optional = false, default-features = false, features = [
     "bundled-sqlcipher"
] }
```

But I know nothing about RUST or SQLCipher and may be totally wrong. Is it possible to use SQLCipher through this library? can it be initialized with a key from the main.rs file? 

Any help on choosing the right direction will be appreciated. Thank you.

HuakunShen:
resolve https://github.com/tauri-apps/plugins-workspace/issues/7

- Introduced `libsqlite3-sys` dependency for SQLite support.
- Updated the `load` method to accept optional pragmas for database connections.
- Enhanced the JavaScript API to demonstrate loading databases with encryption keys and custom pragmas.
- Added VSCode settings for Rust analyzer to enable SQLite feature, to facilitate development.
- Updated Rust code to handle SQLite options and pragmas in the database connection logic.

Now sqlite encryption can be enabled like this.

```ts
const db = await Database.load("sqlite:encrypted.db", {
  sqlite: { pragmas: { "key": "encryption_key" } }
});
```

I already tested on Mac, further testing needed on Windows and Linux.

If you want to try this PR, here is a demo video https://youtu.be/xizlTiX4zTY


HuakunShen:
Preload doesn't support pragma yet, because that will introduce breaking APIs here
https://github.com/tauri-apps/plugins-workspace/blob/43f0f95da6d0232e5d9e7e7290749d27733bea84/plugins/sql/src/lib.rs#L55

If we have to support pragma in preload, then maybe something like this?
```json
{
  "plugins": {
    "sql": {
      "preload": ["sqlite:mydatabase.db"],
      "sqlite_options": {
        "sqlite:mydatabase.db": {
          "pragmas": {
            "journal_mode": "WAL",
            "foreign_keys": "ON"
          }
        }
      }
    }
  }
}
```

HuakunShen:
~I realized one problem, there are two load commands now, one for sqlite one for others. So the current implementation doesn't support multiple features. Either sqlite only or turn off sqlite feature. 
Will need a more robust design.~

Problem fixed, the latest version supports connecting to sqlite, mysql and postgres at the same time

haexhub:
I'm desperately waiting for this feature

HuakunShen:
> I'm desperately waiting for this feature

No one is reviewing this PR.

In the meanwhile, if anyone wants to try this, you can set dependency to my branch in `Cargo.toml`.
And for JS dependency, you may want to clone my branch and link to it with `pnpm/npm/bun link`.

roniemartinez:
+1 on this! I am also waiting to move from my custom implementation to fully adapting to this plugin. Thanks for the PR!

HuakunShen:
> +1 on this! I am also waiting to move from my custom implementation to fully adapting to this plugin. Thanks for the PR!

I was also using my custom implementation with cipher using rusqlite.
While I made this PR, it will be a lot of work to migrate from my implementation to this.
And I want to at least wait til this PR get merged to migrate.

BTW I also figured out how to use drizzle to work with this sql plugin and also rusqlite. 
Working with sqlite in Tauri now feels like in Electron.

plugins/sql/guest-js/index.ts
```diff
@@ -42,12 +42,31 @@ export default class Database {
...
+   *   sqlite: { pragmas: { "journal_mode": "WAL", "foreign_keys": "ON" } }
+   * });
    * ```
    */
-  static async load(path: string): Promise<Database> {
+  static async load(
+    path: string,
+    options?: {
+      sqlite?: {
+        pragmas?: Record<string, string>
```
c6p, line 63: Since you are already adding `options`. Could you add support for extensions, to fix #1705
```
extensions?: Record<string, Option<string>>  // extension, entry_point
```

plugins/sql/src/wrapper.rs
```diff
@@ -82,13 +121,23 @@ impl DbPool {
...
                 create_dir_all(&app_path).expect("Couldn't create app config dir");
-
                 let conn_url = &path_mapper(app_path, conn_url);
+                let filename = conn_url.split_once(':').unwrap().1;
 
-                if !Sqlite::database_exists(conn_url).await.unwrap_or(false) {
-                    Sqlite::create_database(conn_url).await?;
+                let mut sqlite_options = SqliteConnectOptions::new()
+                    .filename(filename)
+                    .create_if_missing(true);
```
c6p, line 129: [Load extensions](https://docs.rs/sqlx/latest/sqlx/sqlite/struct.SqliteConnectOptions.html#method.extension) with entry points.

c6p (changes requested):
Maybe out of scope of this PR, but with minimal change, it can support loading sqlite extensions.

roniemartinez:
Had the chance to test this PR.... but I added my changes for my use: https://github.com/roniemartinez/plugins-workspace/commit/7ca825dc901114fae477e76a8597c2dd9c3b65e0

fokx:
Hope this PR could be merged soon!

FabianLars:
Sorry i don't have much time for plugins at the moment and this also comes a bit in the middle of a refactor of the sql plugin (which is my bad since that should've been long done).

One thing that sure would help is to do what Huakun also asked for and try this PR yourself before it's merged so we get a bit of feedback on the API and issues.

> Preload doesn't support pragma yet, because that will introduce breaking APIs here

We can do it without a breaking change by creating an untagged enum for use in that Vec with string and struct variants.

fokx:
> try this PR yourself

I've tried this PR and can confirm `pragmas` works on Linux.

To build for Android, you need to use `bundled-sqlcipher-vendored-openssl` feature of `libsqlite3-sys` instead of `bundled-sqlcipher`, which cause openssl related build error.

```
libsqlite3-sys = { version = "0.30", features = ["bundled-sqlcipher-vendored-openssl"] }
```
It works on Android after makes such modification. Test method: pull the db file from Android device and open it in Sqlite DB Browser.

roniemartinez:
Is it safe to put “PRAGMA key” in the config file? Putting it in the Rust code like what I did in https://github.com/tauri-apps/plugins-workspace/pull/2553#issuecomment-2848166499 looks a lot more safe because I can get the key dynamically from a server.

FabianLars:
> Is it safe to put “PRAGMA key” in the config file?

Not reallyyy. But since we consider the js side unsafe as well a config option is not much worse than that. Assuming of course that it's not committed to git but injected at build time (true for config, js, and rust). That's a bit less obvious for the config option so maybe you're right and we shouldn't add it (yet).

FabianLars:
> I've tried this PR and can confirm pragmas works on Linux.

Thanks for testing :)

> To build for Android, you need to use bundled-sqlcipher-vendored-openssl feature of libsqlite3-sys instead of bundled-sqlcipher, which cause openssl related build error.

That's a bit problematic for those that build their apps on Windows but i guess there's no way around that. But with that both feature flags should be exposed as plugin features and users should enable what they need (making sqlcipher as a whole option again as well).

salemdomain:
> > I've tried this PR and can confirm pragmas works on Linux.
> 
> Thanks for testing :)
> 
> > To build for Android, you need to use bundled-sqlcipher-vendored-openssl feature of libsqlite3-sys instead of bundled-sqlcipher, which cause openssl related build error.
> 
> That's a bit problematic for those that build their apps on Windows but i guess there's no way around that. But with that both feature flags should be exposed as plugin features and users should enable what they need (making sqlcipher as a whole option again as well).

hey everything works fine with npm run tauri dev, but I encounter issues when building a Windows package from macOS using this command ` npm run tauri build -- --target x86_64-pc-windows-msvc --runner cargo-xwin` and here is the error `error: failed to run custom build command for 'libsqlite3-sys v0.30.1' ` and here is my cargo.toml 
```
[dependencies.tauri-plugin-sql]
features = ["bundled-sqlcipher"] # or "postgres", or "mysql"
version = "^2.0.0-beta"
# alternatively with Git
git = "https://github.com/eleroy/plugins-workspace"
branch = "sql/sqlcipher"
```

oom-:
Any news?

accaolei:
Is anyone continuing to handle this PR?

BIightning:
Any news on this would be appreciated!
Is there maybe something one can help with to get this merged?

FabianLars:
1) Try this PR in your project
2) Preferably provide some feedback on the api
3) address my earlier comment, it must be behind a feature flag (this is the hard blocker)
4) a changefile would be cool but we can add it ourselves if we ever merge it
5) we need more maintainers or there will be a high chance we may remove the plugin from this repo

i'll also be afk for the next 3+ weeks.

plugins/sql/guest-js/index.ts
HuakunShen: I don't have experience using sqlite extension, but I could definitely try

HuakunShen:
I could record a video to show you the process to add this PR to your project. 
Please vote if you want this.

plugins/sql/guest-js/index.ts
FabianLars: But please not in this PR unless you want it to be merged even later x)

oSethoum:
Merging this PR will be helpfull.

oSethoum:
> I could record a video to show you the process to add this PR to your project. Please vote if you want this.

Please do the video.

HuakunShen:
@oSethoum 
Here is the demo recording https://youtu.be/xizlTiX4zTY

HuakunShen:
For people who are waiting for this PR, here is an even better solution: 
https://github.com/HuakunShen/tauri-plugin-libsql
libsql + encryption + drizzle orm + auto migration + local first turso sync

oSethoum:
@HuakunShen did you test it on Windows?

HuakunShen:
> @HuakunShen did you test it on Windows?

~Not yet, I don’t have a windows right now. will test it soon~

Just verified, works on Windows and ubuntu 24

skorphil:
Hello from 2026. Is there any news of implementing encryption to tauri sql? Looks like Tauri died. PRs hanging for years

PR #2569, opened 2025-03-25
feat(auto-start): introduce a builder

Legend-Master:
Introduce a `Builder` for more flexible settings and also for the preparation of newer version of auto launch which introduced a new option to change the registry entry it uses to enable auto start

### DIFF
--- a/.changes/autostart-builder.md
+++ b/.changes/autostart-builder.md
@@ -1,5 +1,6 @@
 ---
 autostart: minor
 autostart-js: minor
+---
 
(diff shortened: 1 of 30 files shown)


.changes/autostart-builder.md
```diff
@@ -0,0 +1,6 @@
+---
+autostart: minor
+autostart-js: minor
+---
+
+Introduce a `Builder` for more flexible settings and also for the preparation of newer version of auto launch which introduced a new option to change the registry entry it uses to enable auto start
```
amrbashir:
```suggestion
Add a `Builder` struct for the plugin, for more flexible configuration.
```

I think that extra information should wait until we actually add that option.

amrbashir (approved):

PR #2572, opened 2025-03-28
fix(updater): download reuse check timeout

### LINKED ISSUE #2372: [bug][updater] Timeout passed to check() is reused for downloadAndInstall()
For quite a while I have been debugging an issue where the updater often failed on one device while there were no issues at all on other devices. I would never have expected that the timeout I pass to `check()` (JS API) would be reused later when downloading the update by calling `downloadAndInstall()`.

~~I don't think there is a reasonable default value for downloading the update because it depends on the app size and connection speed, so I would suggest to remove it. In case of the updater, timeouts IMHO only make sense for the update check and the connection handshakes.~~

Reproduction repo: https://github.com/dgerhardt/tauri2-updater-test/ (downloads but does not start install for security reasons)

Legend-Master:
Fix #2372

Fix `timeout` passed to `check` gets re-used by `download` and `downloadAndinstall` 

Also did some other cleanups

### DIFF
--- a/plugins/updater/src/commands.rs
+++ b/plugins/updater/src/commands.rs
@@ -40,6 +40,8 @@ pub(crate) struct Metadata {
 struct DownloadedBytes(pub Vec<u8>);
 impl Resource for DownloadedBytes {}
 
+// TODO: Align this with the result of `updater.check` to Result<Option<Metadata>>
+// and remove `available` instead of handling this in the js side
 #[tauri::command]
 pub(crate) async fn check<R: Runtime>(
     webview: Webview<R>,
--- a/plugins/updater/src/updater.rs
+++ b/plugins/updater/src/updater.rs
@@ -124,8 +124,7 @@ pub struct UpdaterBuilder {
 impl UpdaterBuilder {
     pub(crate) fn new<R: Runtime>(app: &AppHandle<R>, config: crate::Config) -> Self {
         let app_ = app.clone();
-        let run_on_main_thread =
-            move |f: Box<dyn FnOnce() + Send + Sync + 'static>| app_.run_on_main_thread(f);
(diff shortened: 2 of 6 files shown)


plugins/updater/src/updater.rs
```diff
@@ -478,10 +475,10 @@ impl Updater {
                 version: release.version.to_string(),
                 date: release.pub_date,
                 download_url: release.download_url(&self.json_target)?.to_owned(),
-                body: release.notes.clone(),
                 signature: release.signature(&self.json_target)?.to_owned(),
+                body: release.notes,
                 raw_json: raw_json.unwrap(),
-                timeout: self.timeout,
+                timeout: None,
```
Legend-Master, line 481: This is the actual change 🙃

plugins/updater/src/commands.rs
```diff
@@ -40,6 +40,8 @@ pub(crate) struct Metadata {
 struct DownloadedBytes(pub Vec<u8>);
 impl Resource for DownloadedBytes {}
 
+// TODO: Align this with the result of `updater.check` to Result<Option<Metadata>>
+// and remove `available` instead of handling this in the js side
```
Legend-Master, line 44: I don't know if it's considered breaking if we change this from returning `Result<Metadata>` to `Result<Option<Metadata>>`, if not then I would like to do it before v3

amrbashir: We can change it, if you want since we now require users to sync their rust and js versions

amrbashir (approved):

plugins/updater/src/commands.rs
Legend-Master: That's nice, I'll do it later

PR #2573, opened 2025-03-28
fix(updater): format `Update.date` to RFC 3339

### LINKED ISSUE #2416: Invalid RFC 3339 date string returned from plugin-updater.check()
I have set this `pub_date` value in my API endpoint, which is valid according to RFC 3339:

```
2025-02-11T09:15:00+00:00
```

But this is the `date` value returned from the `check()` call from @tauri-apps/plugin-updater:

```
2025-02-11 9:15:00.0 +00:00:00
```

This is not a valid RFC 3339 date string and it happens because of a naive date to string conversion here:

https://github.com/tauri-apps/plugins-workspace/blob/v2/plugins/updater/src/commands.rs#L75

I believe this should be an explicit conversion according to RFC 3339, as the docs request:
https://v2.tauri.app/plugin/updater/#static-json-file

Opening a PR.

Legend-Master:
Closes #2416
Closes #2417

Also aligned `check`'s return to match the js side's structure


Legend-Master:
Didn't test it locally yet, will mark as ready then

Tested, seems to work

amrbashir (approved):

PR #2576, opened 2025-03-29
Support emitting traces behind feature flag

3lpsy:
**Short Description:**
This PR adds an optional "tracing" feature that will emit log messages to `tracing`. It does not affect current logging. To keep things simple, by enabling the feature, you're opting into emitting to `tracing`. This was to avoid adding state to the Tauri app on behalf of the plugin that would hold some sort of "is_using_tracing" value. 

**Problem it's addressing**
To currently leverage tracing without this PR, applications must call `skip_logger()` and setup of the `tracing-log` logger themselves. This is overall fine. However, I was encountering issues with the `tracing-log` crate on iOS so wanted to a way to avoid the bridge. Additionally, since I only use tracing myself within my code, I only had the adapter to support this plugin and therefore this feature removes that requirement. 

So overall, this PR would support:
1) Allowing both the emission of logs and events. For example if users had different systems in place. 
2) Only emit traces via `skip_logger()` which would nop on the logger invocation and emit the trace.

Implementation is fairly basic in order to avoid unnecessary complexity. Please let me know if you have any feedback.


Legend-Master (approved):

Legend-Master:
I'm not the most familiar with tracing but it looks good to me in general, and since it doesn't affect existing code anyways, this is a good start, I'll merge it

Nice work, thanks!

PR #2581, opened 2025-03-31
chore(process): use `request_restart`

Legend-Master:
Migrate restart to use tauri's new `AppHandle::request_restart` method

Reference: https://github.com/tauri-apps/tauri/pull/12874

### DIFF
--- a/.changes/request-restart.md
+++ b/.changes/request-restart.md
@@ -1,0 +1,6 @@
+---
+process: patch
+process-js: patch
+---
+
+Migrate restart to use tauri's new `AppHandle::request_restart` method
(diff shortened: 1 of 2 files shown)


FabianLars:
shouldn't this be a second command?

Legend-Master:
The JavaScript API doesn't care about the diverging signature and the new `request_restart` actually does what we have documented previously on `restart` so I think it's fine to just replace it (currently `restart` is just `request_restart` but with a diverging return type and doesn't trigger the events if run on main thread)

FabianLars (approved):

PR #2592, opened 2025-04-03
refactor(updater): replace `log` crate with `tracing` for improved logging

thep0y:


Legend-Master:
Is there any reason in particular to use `tracing` instead of `log`? Like we didn't use `span` or anything, wouldn't `log` be enough?

thep0y:
If logs are recorded using the `log` crate, when a user's project uses `tracing` to record logs, it becomes difficult to log messages in the updater due to the difference in logging frameworks. This can be resolved using [tracing_log](https://docs.rs/tracing-log/latest/tracing_log/), but it is not the optimal solution. Currently, all implemented plugins mix the use of `log` and `tracing`. In this scenario, using `tracing` as the default logger is a better choice. Of course, the best implementation would be to make `tracing` an optional or default feature for each plugin, rather than specifying it as a dependency.

Legend-Master:
I see, I feel like we probably want to rethink about where to use `log` where to use `trace`, I didn't even realize that we're mixing the two all over the place 😅

cc @lucasfernog since you added the logs

lucasfernog:
why is tracing_log not optimal? there will always be a bunch of crates using either log or tracing (or in our case, incorrectly mixing both). I see `log` as the common factor between them, and we use tracing mostly when it's marked as an optional feature (which I agree is a bit weird).

thep0y:
All plugins should have `tracing` as an optional dependency, allowing users to conveniently choose either `log` or `tracing` as the logger.

PR #1130, opened 2024-03-30
[v2][log]: Add generic fern::Dispatch TargetKind to log

venkr:
Closes #1113. I've tested this out in our app, and it works as expected.

PR #2600, opened 2025-04-06
[v2][log]: Add generic fern::Dispatch TargetKind to log

### LINKED ISSUE #1113: [log][feature] Allow adding additional `fern::Dispatch`s in the log builder
Currently, the usage of `fern::Dispatch` forces Tauri to be the only source of logging - that's because it runs both `fern::Dispatch::new` and then `self.dispatch.apply()`. Ideally, `fern` is designed to let several different log consumers consume the logs in different manners.

One concrete example we're running into is: we'd like Tauri's strong x-plat console logs as well as Sentry's error monitoring behavior. Ideally, we would be able to simply call `self.dispatch.chain( /* create a new fern::Dispatch for Sentry here */)`.

Two possible implementations:
1- The Builder exposes a `externalDispatch` field which takes a `fern::Dispatch` and chains it into the logger before `apply`ing. Likely without applying any Tauri-specific filtering.
2- The Builder takes as an input a `fern::Dispatch` and is ultimately not responsible for calling `apply`. In this case, some other user code would call `apply`.

### LINKED ISSUE #2525: [Feature Request: log] Expose `chain` method for `dispatch` field in tauri_plugin_log
### Motivation

I want to achieve colored text output in the terminal while maintaining clean text in log files. 

The existing `Builder::dispatch`'s `format` method appears insufficient for implementing conditional formatting logic required for this dual-output scenario.

By exposing the `chain` method of `Builder::dispatch`, developers could leverage fern to implement custom formatting

### Additional context

I previously submitted PR #2451 but had to close it due to missing GPG signature per the Contributing Guide. I can reopen it if needed.

### LINKED ISSUE #2530: [Feature Request: log] Send logs to the Internet (e.g. Google Cloud Logging)
# Motivation

My Tauri app is running in multiple locations and those are NOT connected to WAN.
So I want to send logs into a SaaS (Google Cloud Logging for my case) via Internet.

# Actual

- `Target` and `TargetKind` are not expandable by developers.
- `dispatch` member of`Builder` struct is private
- `Builder.acquire_logger` function accept custom `dispatch`, but the function is private 

# Solution ideas

- Adding custom/generic `TargetKind` as in https://github.com/tauri-apps/plugins-workspace/pull/1130 
    - This seems the most appropriate and versatile approach to me  
- Exposing `Dispatch.chain` as in https://github.com/tauri-apps/plugins-workspace/issues/2525
- Adding `pub dispatch(...)` to `impl Builder`
- Making `Builder.acquire_logger` public.

exoego:
This is a conflict-resolved version of https://github.com/tauri-apps/plugins-workspace/pull/1130
Because I need this feature very soon 

Closes https://github.com/tauri-apps/plugins-workspace/issues/1113
Closes #2525 
Closes https://github.com/tauri-apps/plugins-workspace/issues/2530

### DIFF
--- a/.changes/fern-dispatch-log-target.md
+++ b/.changes/fern-dispatch-log-target.md
@@ -1,0 +1,6 @@
+---
+log: minor
+log-js: minor
+---
+
+Adds a new varient `TargetKind::Dispatch` that allows you to construct arbitrary log targets
--- a/plugins/log/src/lib.rs
+++ b/plugins/log/src/lib.rs
@@ -170,7 +170,7 @@ pub enum TargetKind {
     ///
     /// This requires the webview to subscribe to log events, via this plugins `attachConsole` function.
     Webview,
-    /// Send logs to a fern::Dispatch
+    /// Send logs to a [`fern::Dispatch`]
     ///
     /// You can use this to construct arbitrary log targets.
     Dispatch(fern::Dispatch),
--- a/plugins/log/src/lib.rs
+++ b/plugins/log/src/lib.rs
@@ -170,6 +170,10 @@ pub enum TargetKind {
     ///
     /// This requires the webview to subscribe to log events, via this plugins `attachConsole` function.
     Webview,
+    /// Send logs to a fern::Dispatch
+    ///
+    /// You can use this to construct arbitrary log targets.
+    Dispatch(fern::Dispatch),
 }
 
 /// A log target.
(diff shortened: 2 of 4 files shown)


Legend-Master:
@lucasfernog This is technically a breaking change since we didn't mark `TargetKind` as `non_exhaustive`, but we don't have any public facing API that returns it so it's unlikely an user would use `match` on it, do you think we can merge this or do we wait for v3?

lucasfernog:
i think we can merge it - maybe with a minor bump

Legend-Master (dismissed):

Legend-Master (approved):

Legend-Master:
Thank you both

PR #2609, opened 2025-04-08
single-instance: fix `cwd` in single instance on macOS

Simon-Laux:
which was the `cwd` of the first instance, instead of the second how it should be and is  on windows and linux.


lucasfernog (dismissed):
nice catch, thanks!

PR #2613, opened 2025-04-10
refactor(fs): reduce overhead of `watch`

Legend-Master:
This PR changes a few things

- Migrate the watcher in js side to use `Resource`
- Remove `WatcherResource` and `unwatch` as they're not needed, `notify`'s watcher will clean itself up on drop
- Remove the new thread and channel redirect on every `watch`, as they're not really needed

Also:

- Removed `close` command as it's not used anymore 
- Updated `metadata` to `stat` in readme
- Removed `uuid` as dependency since it's not used anymore


examples/api/src/views/FileSystem.svelte
```diff
@@ -2,16 +2,18 @@
...
   export let insecureRenderHtml;
 
   let path = "";
   let img;
+  /** @type {fs.FileHandle} */
   let file;
   let renameTo;
   let watchPath = "";
-  let watchDebounceDelay = 0;
+  let watchDebounceDelay = "0";
```
amrbashir, line 16: is this right?

plugins/fs/guest-js/index.ts
```diff
@@ -1283,15 +1279,33 @@ async function watch(
...
+    options,
     onEvent
   })
 
-  return () => {
-    void unwatch(rid)
-  }
+  const watcher = new Watcher(rid)
+
+  return watcher.unwatch
```
amrbashir: wouldn't it be better to do 
```js

return () => { watcher.close() }
```

examples/api/src/views/FileSystem.svelte
Legend-Master: Yeah it's bind to an input like this, it it's actually a string

```tsx
<input
  class="input"
  id="watch-debounce-delay"
  bind:value={watchDebounceDelay}
/>
```

plugins/fs/guest-js/index.ts
Legend-Master: That's also fine

amrbashir (approved):

examples/api/src/views/FileSystem.svelte
amrbashir: kinda weird ngl, shouldn't it be bound to an input of number type?

Legend-Master: Yeah probably, but that'll still be a string I think, if svelte didn't do any conversions

PR #2624, opened 2025-04-11
feat(updater): support bundle-specific targets

kandrelczyk:
Here are the changes for updater to handle #2277

I'm using the patched __TAURI_BUNDLE_TYPE variable to download the proper installer with fallback as described in #2277.

I've added some additional test on both Linux and Windows. Not sure if something needs to be done on other platforms.


kandrelczyk:
Windows and Linux tests are failing because they require tauri-cli with https://github.com/tauri-apps/tauri/pull/13209 change.

MacOS test is failing because of some compliation issue. I will have access to a mac next week so I will fix it.

FabianLars:
> MacOS test is failing because of some compliation issue. I will have access to a mac next week so I will fix it.

It looks like you ran `cargo update` - if you revert the Cargo.lock file changes it should compile again.

plugins/updater/src/lib.rs
```diff
@@ -29,6 +29,12 @@ pub use config::Config;
 pub use error::{Error, Result};
 pub use updater::*;
 
+/// Variable holding the type of bundle the executable is stored in. This is modified by binary
+/// patching during build
+#[unsafe(no_mangle)]
+#[link_section = ".data.ta"]
+pub static __TAURI_BUNDLE_TYPE: &str = "UNK_BUNDLE";
```
FabianLars: hmm, if possible it may make sense to move that into the main tauri crate (or utils idk) because a few people asked for this in the past regardless of the updater

kandrelczyk: I had it in the main tauri crate originally but build (`cargo test`) was failing during linking. I will try to move it to utils.

FabianLars: ah yeah i remember you mentioning this

plugins/updater/src/updater.rs
```diff
@@ -342,11 +383,21 @@ pub struct Updater {
 }
 
 impl Updater {
+    fn get_updater_installer(&self) -> Result<Option<Installer>> {
+        match __TAURI_BUNDLE_TYPE {
```
lucasfernog: `__TAURI_BUNDLE_TYPE` shouldn't be exported from Tauri, but we should map it to a `PackageFormat` type directly from tauri instead of doing the check here, it would make it type safe (instead of relying on string matching from tauri-cli)

kandrelczyk: What do we want to do for Mac, iOS and Android? Or in general case when the binary was not patched and we don't know the bundle type? I think the simplest thing to do would be to return PackageFormat::Unknown in case the binary was not patched and use `#[cfg(target_os = "macos")]` to return PackageFormat::Dmg for Mac. 

We have BundleType in tauri::utils::config but if we want to return Unknown I would have to create new type. What would be the best place to put it? tauri::utils::config? some new module?

plugins/updater/src/updater.rs
```diff
@@ -71,26 +102,39 @@ pub struct RemoteRelease {
 
 impl RemoteRelease {
     /// The release's download URL for the given target.
-    pub fn download_url(&self, target: &str) -> Result<&Url> {
+    pub fn download_url(&self, target: &str, installer: Option<Installer>) -> Result<&Url> {
+        let fallback_target = installer.map(|installer| format!("{target}-{}", installer.suffix()));
```
Legend-Master: Isn't this reversed? It should be we check the update with the current installer target first and then fallback to the one without the suffix (e.g. `x86_64-windows-nsis` -> `x86_64-windows`)

kandrelczyk: This part is pretty confusing. I think it's correct, only the names of the variables are wrong. I will try to rewrite it or a least change the names of the variables to make it clearer.

lucasfernog: yeah it's reversed

Cargo.toml
```diff
@@ -39,3 +39,6 @@ codegen-units = 1
 lto = true
 incremental = false
 opt-level = "s"
+
+[patch.crates-io]
+tauri = { git = "https://github.com/tauri-apps/tauri", rev = "232265c70e1c213bbb3f84b5541ddc07d330fce1" }
```
FabianLars: we must not forget to configure the minimum version once tauri was released. sadly we can't set a min cli version.

plugins/updater/src/updater.rs
kandrelczyk: pushed a fix

plugins/updater/src/updater.rs
```diff
@@ -71,26 +102,36 @@ pub struct RemoteRelease {
 
 impl RemoteRelease {
     /// The release's download URL for the given target.
-    pub fn download_url(&self, target: &str) -> Result<&Url> {
+    pub fn download_url(&self, fallback_target: &str, installer: Option<Installer>) -> Result<&Url> {
```
Legend-Master: Honestly, this name doesn't make much sense either, can we change the whole chain to something like:

1. User provided custom target (no fallback)
2. `{OS}-{architecture}-{installer}` with a fallback to `{OS}-{architecture}`

Essentially, we keep the old `download_url` and `signature`, and change the `download_url: release.download_url(&self.json_target)?.to_owned()` part in `check` instead with these logic

kandrelczyk: Moved this logic to check. I think it's a bit clearer now.

plugins/updater/tests/app-updater/tests/update.rs
```diff
@@ -48,7 +49,7 @@ struct Update {
 fn build_app(cwd: &Path, config: &Config, bundle_updater: bool, target: BundleTarget) {
     let mut command = Command::new("cargo");
     command
-        .args(["tauri", "build", "--debug", "--verbose"])
+        .args(["tauri", "build", "--verbose"])
```
kandrelczyk, line 52: Since we had issues with the bundle type variable being stripped in release mode but not in debug mode I changed this. We test this in tauri crate as well but not we only have a warning there, not an error. The test will take longer but it may catch some issues that otherwise would slip through.  What do you think?

Legend-Master: Good idea to me

kandrelczyk:
Ok, I've updated tauri version for updater plugin to 2.7 and the tests now pass with latest tauri-cli. Anything missing here?

plugins/updater/src/error.rs
```diff
@@ -39,9 +39,12 @@ pub enum Error {
...
     #[error(transparent)]
     Reqwest(#[from] reqwest::Error),
-    /// The platform was not found on the updater JSON response.
-    #[error("the platform `{0}` was not found on the response `platforms` object")]
+    /// The platform was not found in the updater JSON response.
+    #[error("the platform `{0}` was not found in the response `platforms` object")]
     TargetNotFound(String),
+    /// Neither the platform not the fallback platform was not found in the updater JSON response.
+    #[error("the platform `{0}` and `{1}` were not found in the response `platforms` object")]
+    TargetsNotFound(String, String),
```
beanow-at-crabnebula: The error message reads like both are required to be present. But it's either, right?

plugins/updater/src/updater.rs
```diff
@@ -466,6 +510,31 @@ impl Updater {
             None => release.version > self.current_version,
         };
 
+        let mut download_url = release.download_url(&self.json_target);
+        let mut signature = release.signature(&self.json_target);
+
+        let installer = self.get_updater_installer();
+        if installer.is_none() && (download_url.is_err() || signature.is_err()) {
+            return Err(Error::TargetNotFound(self.json_target.clone()));
+        }
```
beanow-at-crabnebula: Because the updater is getting additional logic and decisions, can we add a few logs about those decisions?

plugins/updater/src/updater.rs
```diff
@@ -477,9 +546,10 @@ impl Updater {
                 extract_path: self.extract_path.clone(),
                 version: release.version.to_string(),
                 date: release.pub_date,
-                download_url: release.download_url(&self.json_target)?.to_owned(),
-                signature: release.signature(&self.json_target)?.to_owned(),
-                body: release.notes,
+                download_url: download_url?.to_owned(),
+                body: release.notes.clone(),
+                signature: signature?.to_owned(),
+                installer,
```
beanow-at-crabnebula: It looks like the installer would return `Some` even when the corresponding entry does not get used for it's download_url or signature.

plugins/updater/src/updater.rs
```diff
@@ -466,6 +510,31 @@ impl Updater {
...
+                    .or(download_url.or(Err(Error::TargetsNotFound(
+                        self.json_target.clone(),
+                        target.clone(),
+                    ))));
+            signature = release
+                .signature(target)
+                .or(signature.or(Err(Error::TargetsNotFound(
+                    self.json_target.clone(),
+                    target.clone(),
+                ))));
```
beanow-at-crabnebula: You can't fall back separately for the download URL and signature.
Either we select one target's signature + url, or the other.

Currently if the target download URL exists, but the signature is missing, it will use the wrong signature.

plugins/updater/src/updater.rs
kandrelczyk: I believe this is correct. We use this information later in `install_inner` to determine installation method. So for example if the current binary was packaged as Deb but we didn't find specific download_url for deb and used the fallback we still assume that it points to deb package and will try to install it as deb. If we know the current binary was packaged as deb it doesn't make sense to try to install the update as AppImage. Only in case we don't know what the current bundle type is will we fallback to AppImage.

kandrelczyk:
@beanow-at-crabnebula I've addressed the comments. Please take a look and let me know if anything else needs fixing.

Cargo.toml
Legend-Master: We can remove these now

Legend-Master (commented):
You probably want to revert the change in `Cargo.lock`, and change the `tauri` version in root/updater's `Cargo.toml` file to 2.7 instead

Cargo.toml
kandrelczyk: Reverted Cargo.lock and removed this. I hope it's fine now. Sorry, I'm not familiar with this way of patching.

Legend-Master: No worries, I meant something like 0a495cc, it should be fixed now (bump the tauri in lock file to 2.7 and the updater's manifest to ^2.7 while keeping the workspace tauri manifest at ^2.6)

Legend-Master:
Honestly, I feel like the entire `target` thing is a bit confusing, `tauri_plugin_updater::target` returns `{os}-{arch}` while we replace `{{target}}` inside the URL with just the OS, also with the introduction of `installer` to the mix, this gets even more confusing, should `tauri_plugin_updater::target` return with the installer info or not?

Also feel like we should probably remove the `json_target` field in `Updater` and handle this logic in `check` instead

kandrelczyk:
@Legend-Master moved the logic to `check`.

Legend-Master (dismissed):
The main code looks good to me, didn't get a chance to look at the tests yet, also we'll need a change file 🙃

Thanks for the patience on this ❤️

kandrelczyk:
Added change file.

And thank **you**. This should go way smoother but I'm hitting a lot of Rust related things that are new to me. 

BTW what about `latest.json` file? Generating latest.json with new version of `target` is something we still have to do, right? Is that part of github actions?

Legend-Master:
> And thank you. This should go way smoother but I'm hitting a lot of Rust related things that are new to me.

No worries, we all came from there 🙃

> BTW what about latest.json file? Generating latest.json with new version of target is something we still have to do, right? Is that part of github actions?

Yeah, that's either made by the user (we need to update the docs after this) or through https://github.com/tauri-apps/tauri-action, and in that case, we need to give it an update as well

Legend-Master (dismissed):
Looks good to me!

Not entirely sure if we should also change the Windows updater detection to use it instead of infer, but I guess that could be for another day

@FabianLars @lucasfernog Maybe you guys want to also give it a review in case I missed something 🙃 (especially about the dynamic server part since I didn't quite follow the discussions about that part)

plugins/updater/src/updater.rs
```diff
@@ -1201,14 +1273,14 @@ impl Update {
 
 /// Gets the target string used on the updater.
 pub fn target() -> Option<String> {
```
Legend-Master, line 1262: Maybe we could document the current behavior somewhere, this is not the only default target now

> https://github.com/tauri-apps/plugins-workspace/pull/2624#issuecomment-3116202953

kandrelczyk: Shouldn't this return Vec<String> instead now? If the installer info is present it will return ["{os}-{arch}-{installer}, "{os}-{arch}"]. If installer is None return ["{os}-{arch}"]? Or if we want to keep this backward compatible we can mention `tauri::utils::platform::bundle_type` here. I don't really know what people use this for.

Legend-Master: Don't know why it's public to be honest, maybe used in unit test for the dynamic server?
I think just some docs would be nice, I wouldn't add/change the fallback thing to it unless someone asks for it

kandrelczyk: Ok, updated the docs. Take a look.

FabianLars:
> Maybe you guys want to also give it a review in case I missed something 🙃 (especially about the dynamic server part since I didn't quite follow the discussions about that part)

give me a day or two.

also don't worry about tauri-action (unless you want to of course)

plugins/updater/src/error.rs
```diff
@@ -39,9 +39,14 @@ pub enum Error {
     /// `reqwest` crate errors.
     #[error(transparent)]
     Reqwest(#[from] reqwest::Error),
-    /// The platform was not found on the updater JSON response.
-    #[error("the platform `{0}` was not found on the response `platforms` object")]
+    /// The platform was not found in the updater JSON response.
+    #[error("the platform `{0}` was not found in the response `platforms` object")]
     TargetNotFound(String),
+    /// Neither the platform not the fallback platform was not found in the updater JSON response.
```
FabianLars:
```suggestion
    /// Neither the platform nor the fallback platform was found in the updater JSON response.
```

plugins/updater/src/updater.rs
```diff
@@ -511,8 +587,11 @@ pub struct Update {
     pub version: String,
     /// Update publish date
     pub date: Option<OffsetDateTime>,
-    /// Target
+    /// The `{{target}}` variable we replace in the endpoint and serach for in the JSON,
```
```suggestion
    /// The `{{target}}` variable we replace in the endpoint and search for in the JSON,
```

plugins/updater/src/updater.rs
```diff
@@ -511,8 +587,11 @@ pub struct Update {
...
+    /// this is either the user provided target or the current operating system by default
     pub target: String,
+    /// Current installer
```
```suggestion
    /// Current installer type
```

plugins/updater/src/updater.rs
```diff
@@ -381,11 +421,11 @@ impl Updater {
                 .to_string()
                 // url::Url automatically url-encodes the path components
                 .replace("%7B%7Bcurrent_version%7D%7D", &encoded_version)
-                .replace("%7B%7Btarget%7D%7D", &self.target)
+                .replace("%7B%7Btarget%7D%7D", target)
```
FabianLars, line 412:
```suggestion
                .replace("%7B%7Btarget%7D%7D", target)
                .replace("%7B%7Binstaller%7D%7D", installer.name())
```
needs the let installer moved above this of course. ignore this if this was planned for another pr i guess

plugins/updater/src/updater.rs
```diff
@@ -381,11 +421,11 @@ impl Updater {
...
                 .replace("%7B%7Barch%7D%7D", self.arch)
                 // but not query parameters
                 .replace("{{current_version}}", &encoded_version)
-                .replace("{{target}}", &self.target)
+                .replace("{{target}}", target)
```
FabianLars, line 416:
```suggestion
                .replace("{{target}}", target)
                .replace("{{installer}}", installer.name())
```

plugins/updater/src/updater.rs
```diff
@@ -342,6 +365,18 @@ pub struct Updater {
...
 impl Updater {
+    fn get_updater_installer(&self) -> Option<Installer> {
+        match bundle_type()? {
+            BundleType::Deb => Some(Installer::Deb),
+            BundleType::Rpm => Some(Installer::Rpm),
+            BundleType::AppImage => Some(Installer::AppImage),
+            BundleType::Msi => Some(Installer::Msi),
+            BundleType::Nsis => Some(Installer::Nsis),
+            BundleType::App => Some(Installer::App),
+            BundleType::Dmg => None,
```
FabianLars: should this return Installer::App as well? (the updater does use the `.app` bundle for apps installed via .dmg)

FabianLars (commented):
not a proper review yet, just a few things i've noticed half asleep x)

plugins/updater/src/updater.rs
```diff
@@ -342,6 +365,18 @@ pub struct Updater {
 }
 
 impl Updater {
+    fn get_updater_installer(&self) -> Option<Installer> {
+        match bundle_type()? {
```
beanow-at-crabnebula: This does not read from `&self`.
I'd suggest making this a pure function along these lines.

```rust
pub fn installer_for_bundle_type(bundle: &BundleType) -> Option<Installer>
```

plugins/updater/src/updater.rs
beanow-at-crabnebula: I think that conflates two things:
- Detecting information about the current environment.
- Determining what update to install from remote release information.

It also mixes up responsibilities. Should this function reject the update that's being served when it's for the incorrect installer? Or should the consumer of this function handle that?

`tauri_utils::platform::bundle_type()` already addresses the environment detection aspect.
Especially when converting that bundle is a pure function.

beanow-at-crabnebula (commented):
Thanks for the effort :]
Just suggestions from my end.

plugins/updater/src/updater.rs
kandrelczyk: I think this is covered in `tauri::platform::utisl::bundle_type`:

```
    match __TAURI_BUNDLE_TYPE {
      "DEB" => Some(BundleType::Deb),
      "RPM" => Some(BundleType::Rpm),
      "APP" => Some(BundleType::AppImage),
      "MSI" => Some(BundleType::Msi),
      "NSS" => Some(BundleType::Nsis),
      _ => {
        if cfg!(target_os = "macos") {
          Some(BundleType::App)
        } else {
          None
        }
      }
```

On mac we also get BundleType::App.. I will change BundleTYpe::Dmg to default case as it better explains what happens.

kandrelczyk:
Addressed the comments.

Legend-Master (dismissed):
Looks good on my part

plugins/updater/src/updater.rs
FabianLars: > On mac we also get BundleType::App.. I will change BundleTYpe::Dmg to default case as it better explains what happens.

I guess that didn't happen but maybe a comment would help prevent others from asking the same question i asked.

FabianLars:
> Addressed the comments.

Not all of them yet :)
In case it's github's UI being weird again, here are the ones i saw:
https://github.com/tauri-apps/plugins-workspace/pull/2624#discussion_r2235237479
https://github.com/tauri-apps/plugins-workspace/pull/2624#discussion_r2234117536
https://github.com/tauri-apps/plugins-workspace/pull/2624#discussion_r2234118038
https://github.com/tauri-apps/plugins-workspace/pull/2624#discussion_r2259881222
https://github.com/tauri-apps/plugins-workspace/pull/2624#discussion_r2235218319

plugins/updater/src/updater.rs
kandrelczyk: I prefer to leave it for another PR

plugins/updater/src/updater.rs
kandrelczyk: Same, here. I would leave it for another PR

plugins/updater/src/updater.rs
kandrelczyk: I changed this to a default case and added a comment. This switch was moved to a method, probably that's why it's not showing here.

plugins/updater/src/error.rs
```diff
@@ -67,8 +72,8 @@ pub enum Error {
     TempDirNotFound,
     #[error("Authentication failed or was cancelled")]
     AuthenticationFailed,
-    #[error("Failed to install .deb package")]
-    DebInstallFailed,
+    #[error("Failed to install package")]
+    PackageInstallFailed,
```
FabianLars, line 78: this is a breaking change. i doubt that anyone uses that enum variant though so i can act like i didn't see it if you prefer :D

plugins/updater/src/updater.rs
FabianLars: hmm okay

plugins/updater/src/error.rs
kandrelczyk: Added it back, no problem.

kandrelczyk:
@FabianLars checked all the comments again. Some of them I would rather leave for another PR (better to finally merge this and keep working in a clean branch). I moved the bundle type to a pure function. The only unaddressed comment is the one about  mixed up responsibilities but I don't know what specifically should be changed there. Any ideas?

lucasfernog:
I think we can release this one next week

Legend-Master:
I have reverted the tauri bumps to align with the main branch, and since we have already bumped the workspace tauri requirement to 2.8 in main branch, we can remove the separate bump in the updater plugin now

lucasfernog (approved):

FabianLars:
after merging this, please wait with the release until we support the `{{}}` url thingy too :) i feel like it doesn't make sense to not release them together.

FabianLars (approved):

kandrelczyk:
@FabianLars can please explain what needs to be done for the `{{}}`? Is it for the dynamic URL? What exactly is missing and how it should work?

I will also take a look at latest.json generation.

FabianLars:
it's for the dynamic urls, yes and i'm pretty sure that it would require really just https://github.com/tauri-apps/plugins-workspace/pull/2624#discussion_r2234117536 and https://github.com/tauri-apps/plugins-workspace/pull/2624#discussion_r2234118038

edit: nevermind, with some of the changes since then those suggestions don't work as-is anymore but the gist is the same.

kandrelczyk:
Ok, I'm happy to do it. I can work on it this week.

FabianLars:
went ahead and opened https://github.com/tauri-apps/plugins-workspace/pull/2960 sorry 😅 wanted to get to testing asap

PR #2633, opened 2025-04-14
fix(biometric) prompt for fallback credentials when enabled and biometry unavailable on iOS (fix #2632)

### LINKED ISSUE #2632: [bug] [biometric] Fallback logic ignored when biometry is unavailable on iOS
## Describe the problem
When biometry isn't enrolled or available, but fallback credentials are allowed, calls to the `authenticate` command fail without prompting for fallback credentials on iOS platforms. This differs from the behavior on Android where the plugin does prompt for fallback credentials when biometry isn't enrolled or available.

## Context
Device: Physical iPhone 15 Pro Max (`iPhone(16,2)`)
OS: `iOS 18.3`
tauri-plugin-biometric version: `2.2.0`

## Steps to reproduce
In my case, I tested this using the `examples/api` app, but general reproduction steps are as follows:
  - Run a Tauri app that uses the biometric plugin on an iOS device or simulator; ensure biometry is unenrolled but a passcode is enabled.
  - Call the `authenticate` command with `allowDeviceCredential` set to `true`
  - Observe authentication failure, despite fallback being allowed.

pjf-dev:
## Description
Fix #2632: This fixes the issue described in #2632 where authentication would fail without prompting for fallback credentials when biometry wasn't available or enrolled on iOS platforms.

## Root cause
The authenticate function checked the stored biometric status and failed immediately if unavailable, ignoring fallback logic when `allowDeviceCredential = true`.

## Proposed solution
Ensure fallback credential is also disabled before failing early due to biometric availability in the iOS module of the plugin.

## Additional Notes
I've tested the fix on a physical iPhone 15 Pro Max running iOS 18.3 and confirmed that a fallback prompt now appears if biometry is disabled and fallback is allowed.

### DIFF
--- a/.changes/fix-ios-biometry-fallback-auth.md
+++ b/.changes/fix-ios-biometry-fallback-auth.md
@@ -3,4 +3,4 @@
 "biometric-js": patch:bug
 ---
 
-Fix biometric plugin not prompting for fallback credentials when biometry status is unavailable or not enrolled on iOS.
+Fix biometric plugin ignoring fallback logic when biometry status is unavailable or not enrolled on iOS.
--- a/.changes/fix-ios-biometry-fallback-auth.md
+++ b/.changes/fix-ios-biometry-fallback-auth.md
@@ -1,0 +1,6 @@
+---
+"biometric": patch:bug
+"biometric-js": patch:bug
+---
+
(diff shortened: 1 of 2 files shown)


pjf-dev:
I believe I covered mostly everything, but let me know if anything else is needed for this PR

lucasfernog (approved):
nice catch!

PR #2636, opened 2025-04-16
fix(http): handle status code 204 to use js null fix #2580

### LINKED ISSUE #2580: [bug] tauri-apps/plugin-http incorrectly handles responses for status code 204
### Describe the bug
HTTP plugin throws the following error, when receiving a response with status code `204`
```js
TypeError: Failed to construct 'Response': Response with null body status cannot have body
```

### Reproduction
```js
import { fetch } from '@tauri-apps/plugin-http';

const response = await fetch('https://www.google.com/generate_204');
```

@tauri-apps/plugin-http v2.4.2

asomethings:
## Description

`tauri-apps/plugin-http` handles response as stream which causes error `Response with null body status cannot have body` in Javascript.

## Fix

When status is `204` then it uses `null` instead of `readableStreamBody`.

fixes: #2580


lucasfernog (approved):
thank you!

FabianLars:
we probably should do the same for 205 and 304. They will throw the same error (at least in [firefox](https://searchfox.org/mozilla-central/source/dom/fetch/Response.cpp#258))

PR #2641, opened 2025-04-17
feat: support message dialogs with 3 buttons

### LINKED ISSUE #2640: Allow for a third button in the MessageDialogBuilder
Hello,
It would be great if you can allow for a third button in the MessageDialogBuilder, in order to cover some pretty common scenarios, such as "Yes/No/Cancel", or for example when exiting the application "Save/Don't save/Cancel".
For the record, this was also requested in the past in issues [#7855](https://github.com/tauri-apps/tauri/issues/7855) and [#7888](https://github.com/tauri-apps/tauri/pull/7888).
Thanks!

amrbashir:
closes #2640


amrbashir:
iOS is not tested ofc

Windows and Android are tested and works.

plugins/dialog/guest-js/index.ts
```diff
@@ -202,6 +297,32 @@ async function save(options: SaveDialogOptions = {}): Promise<string | null> {
...
+ * @since 2.3.0
+ */
+export type MessageDialogResult = 'Yes' | 'No' | 'Ok' | 'Cancel' | (string & {})
+
+type MessageDialogResultRust =
+  | 'Yes'
+  | 'No'
+  | 'Ok'
+  | 'Cancel'
+  | { Custom: string }
```
Legend-Master: If we're going to convert it in the js side, what about we just do a `#[serde(untagged)]` in the rust side?

plugins/dialog/src/models.rs
```diff
@@ -65,4 +67,44 @@ pub enum MessageDialogButtons {
...
+    YesNoCancelCustom(String, String, String),
+}
+
+/// Result of a message dialog
+#[derive(Debug, Default, Clone, PartialEq, Eq, Deserialize, Serialize)]
+pub enum MessageDialogResult {
+    Yes,
+    No,
+    Ok,
+    #[default]
```
Legend-Master, line 80: We won't need this if we impl `into` (see the comment below)

plugins/dialog/src/models.rs
```diff
@@ -65,4 +67,44 @@ pub enum MessageDialogButtons {
...
+            rfd::MessageDialogResult::Yes => Self::Yes,
+            rfd::MessageDialogResult::No => Self::No,
+            rfd::MessageDialogResult::Ok => Self::Ok,
+            rfd::MessageDialogResult::Cancel => Self::Cancel,
+            rfd::MessageDialogResult::Custom(s) => Self::Custom(s),
+        }
+    }
+}
+
+impl FromStr for MessageDialogResult {
```
Legend-Master: If this would never fail, maybe we just change it to `From<String>` and `From<str>`

plugins/dialog/src/models.rs
```diff
@@ -65,4 +67,44 @@ pub enum MessageDialogButtons {
     OkCustom(String),
     /// 2 buttons `Ok` and `Cancel` with custom texts
     OkCancelCustom(String, String),
+    /// 3 buttons `Yes`, `No` and `Cancel` with custom texts
+    YesNoCancelCustom(String, String, String),
```
Legend-Master, line 71: Maybe we could also add `YesNoCancel`?

plugins/dialog/src/models.rs
amrbashir: we still need it, for when the message from kotling/swift to rust fails

Legend-Master: Got it

plugins/dialog/guest-js/index.ts
```diff
@@ -77,6 +77,72 @@ interface SaveDialogOptions {
...
+ *
+ * @since 2.3.0
+ */
+export type MessageDialogButtonsYesNoCancel = {
+  /** The Yes button. */
+  yes?: string
+  /** The No button. */
+  no?: string
+  /** The Cancel button. */
+  cancel?: string
```
Legend-Master: I think these shouldn't be optional right? also the other ones

plugins/dialog/src/desktop.rs
```diff
@@ -115,6 +115,12 @@ impl From<MessageDialogButtons> for rfd::MessageButtons {
             MessageDialogButtons::YesNo => Self::YesNo,
             MessageDialogButtons::OkCustom(ok) => Self::OkCustom(ok),
             MessageDialogButtons::OkCancelCustom(ok, cancel) => Self::OkCancelCustom(ok, cancel),
+            MessageDialogButtons::YesNoCancel => {
+                Self::YesNoCancelCustom(YES.into(), NO.into(), CANCEL.into())
+            }
```
Legend-Master:
```suggestion
            MessageDialogButtons::YesNoCancel => Self::YesNoCancel,
```

amrbashir: rfd doesn't have this variant

amrbashir: Nvm, it actually has it, I must have been blind

plugins/dialog/guest-js/index.ts
```diff
@@ -202,6 +301,16 @@ async function save(options: SaveDialogOptions = {}): Promise<string | null> {
...
 
+/**
+ * The result of a message dialog.
+ *
+ * The result is a string if the dialog has custom buttons,
+ * otherwise it is one of the default buttons.
+ *
+ * @since 2.3.0
+ */
+export type MessageDialogResult = 'Yes' | 'No' | 'Ok' | 'Cancel' | (string & {})
```
Legend-Master, line 337: Should these be in lower case?

plugins/dialog/guest-js/index.ts
```diff
@@ -77,6 +77,72 @@ interface SaveDialogOptions {
   canCreateDirectories?: boolean
 }
 
+/**
+ * Default buttons for a message dialog.
+ *
+ * @since 2.3.0
+ */
+export type MessageDialogDefaultButtons =
```
Legend-Master, line 85: Should we use kebab case (`ok-cancel`) or camel case (`okCancel`) for these?

plugins/dialog/src/models.rs
```diff
@@ -61,8 +61,49 @@ pub enum MessageDialogButtons {
...
+#[derive(Debug, Default, Clone, PartialEq, Eq, Deserialize, Serialize)]
+#[serde(untagged)]
+pub enum MessageDialogResult {
+    Yes,
+    No,
+    Ok,
+    #[default]
+    Cancel,
+    Custom(String),
+}
```
Legend-Master: The untagged needs to be on the field or other fields will become `null`

```suggestion
#[derive(Debug, Default, Clone, PartialEq, Eq, Deserialize, Serialize)]
pub enum MessageDialogResult {
    Yes,
    No,
    Ok,
    #[default]
    Cancel,
    #[serde(untagged)]
    Custom(String),
}
```

plugins/dialog/guest-js/index.ts
amrbashir: Not really, I think PascalCase is better here as they are sort of constants

plugins/dialog/guest-js/index.ts
amrbashir: Same as above, PascalCase seems better here

plugins/dialog/guest-js/index.ts
Legend-Master: Asking because our `MessageDialogOptions.kind` is in lower case (`kind?: 'info' | 'warning' | 'error'`) and things like `checkPermissions` returns `type PermissionState = 'granted' | 'denied' | 'prompt' | 'prompt-with-rationale'`

amrbashir: I still don't like them tbh, so I changed `MessageDialogDefaultButtons` to an Enum

amrbashir: `MessageDialogResult` should stay the same tbh as it matches what `rfd` returns as the result for custom and sometimes built-in buttons.

Legend-Master: > `MessageDialogResult` should stay the same tbh as it matches what `rfd` returns as the result for custom and sometimes built-in buttons.

That makes sense

Legend-Master: > I still don't like them tbh, so I changed MessageDialogDefaultButtons to an Enum

To be honest that enum feels too lengthy than it could have been `await message('hi', { buttons: MessageDialogDefaultButtons.Ok })`, also typescript enums are kinda weird and have runtime costs that I would like to avoid most of the time, I would prefer we just use the union type with the changed casing or not

About the casing, I think this is what most of the web are using, and we should follow the convention

> https://w3ctag.github.io/design-principles/#casing-rules

amrbashir: We shouldn't strictly follow these design principles as tauri is a mix of Rust and JS, I would prefer to have APIs consistent between JS and Rust, as it makes it easier to find and less cognitive load imo.

amrbashir: > To be honest that enum feels too lengthy than it could have been `await message('hi', { buttons: MessageDialogDefaultButtons.Ok })`

I am fine to change it back to constant strings, but I would still keep them PascalCase

Legend-Master: Alright, let's just use the union type and keep the casing then, I don't mind it too much, just feel like that's more consistent with other APIs we currently have

amrbashir:
@FabianLars or @lucasfernog can you test the iOS stuff?

FabianLars:
Can do but I'm kinda away over the weekend so please remind me on Monday if I or Lucas didn't get to it yet by then.

amrbashir:
don't worry about it, enjoy your weekend

lucasfernog:
I started taking a look at it but got carried away to other things. I was trying to improve the return type but.. it's complicated

Legend-Master (dismissed):
Desktop code looks good, tested on Windows

plugins/dialog/guest-js/index.ts
```diff
@@ -77,6 +77,72 @@ interface SaveDialogOptions {
...
+
+/**
+ * The Ok button of a message dialog.
+ *
+ * @since 2.3.0
+ */
+export type MessageDialogButtonsOk = {
+  /** The Ok button. */
+  ok: string
+}
```
Legend-Master: Just a small suggestion, maybe we can change these to something like this, so typescript doesn't hint `ok` and `yes` at the same type

```suggestion
export type MessageDialogButtonsYesNoCancel = {
  /** The Yes button. */
  yes: string
  /** The No button. */
  no: string
  /** The Cancel button. */
  cancel: string
}

/**
 * The Ok and Cancel buttons of a message dialog.
 *
 * @since 2.3.0
 */
export type MessageDialogButtonsOkCancel = {
  /** The Ok button. */
  yes: string
  /** The Cancel button. */
  cancel: string
}

/**
 * The Ok button of a message dialog.
 *
 * @since 2.3.0
 */
export type MessageDialogButtonsOk = {
  /** The Ok button. */
  yes: string
}
```

FabianLars:
@lucasfernog 
> I was trying to improve the return type but.. it's complicated

Can you maybe expand on that so someone else can potentially take over?

FabianLars:
i love this :D
grafik

lucasfernog:
> i love this :D grafik

:joy: :joy: omg it took me a minute to realize wtf is this

lucasfernog (dismissed):
i think we can go with this now

Legend-Master:
To be honest, although it's correct at the type level and it does error out if extra buttons are passed, this type makes it harder to follow to me

image

instead of 

image

lucasfernog:
weird, i only ever got the second type of warning.. maybe we just go with the other way :|
i didn't really like using `ok` instead of `yes` because then it wouldn't match the rust side.. and `ok / no / cancel` is a weird trio IMO, but maybe that's just me

Legend-Master:
I feel like `ok / no / cancel` is ok because most of the time, 3 buttons will be 1 primary action 1 secondary action and cancel, it's not really `yes / no / cancel`

But this doesn't quite match the `rfd`'s default `YesNoCancel` so...

I feel like as long as we put some examples in `MessageDialogOptions.buttons` so people can copy paste them, I think the error message wouldn't be a problem in that case

Legend-Master (approved):

HumanWannabe:
Hello,

I tried using this, but it doesn't seem to work for me. I made sure to update my Tauri, so when I run this I get "2.9.4":

	import { getTauriVersion } from '@tauri-apps/api/app';

	// Show Tauri version on load
	getTauriVersion().then(version => {
		console.log('Tauri version:', version);
		alert(`Tauri version: ${version}`);
	});

Then I run the following for example:
			const userChoice = await message('Hello World!', { title:'title works', buttons: {ok:'OkCancel'}, kind: 'warning', okLabel: 'save', cancelLabel: 'cancelme' })

But I get only a "save" button. I tried all variations, with and without custom labels, but "buttons" seems to be completely ignored. 
Am I missing something?

Legend-Master:
Take a look at the examples https://github.com/tauri-apps/plugins-workspace/pull/2641/commits/355fc91f13a62faecb8139d5a8cdf7c7ff530a56, you want to use either `buttons: 'OkCancel'` or `buttons: { ok: 'save, cancel: 'cancel' }`

HumanWannabe:
> Take a look at the examples [355fc91](https://github.com/tauri-apps/plugins-workspace/commit/355fc91f13a62faecb8139d5a8cdf7c7ff530a56), you want to use either `buttons: 'OkCancel'` or `buttons: { ok: 'save, cancel: 'cancel' }`

I am :( 
I literally copy/pasted this from the examples:
   await message('Show the file content?', {
     buttons: { yes: 'Show content', no: 'Show in folder', cancel: 'Cancel' }
   })
and it still gives me only an "Ok" button. Is it possible that I require a special build, or that I do not have the proper Tauri version? how can I confirm this?

Legend-Master:
> await message('Show the file content?', {
> buttons: { yes: 'Show content', no: 'Show in folder', cancel: 'Cancel' }
> })

image
 
The exact code seems to work for me on Windows, could you share your `tauri info` output?

HumanWannabe:
image

FabianLars:
try updating your `@tauri-apps/plugin-dialog` npm package then. This feature was added in 2.4

HumanWannabe:
I tried creating a new project, and it worked there. I guess I need to dig a bit deeper to see why it still refuses to work in my existing project after all the updates. 
Thanks everyone for your help!! :)

PR #2648, opened 2025-04-19
fix(biometric) Incorrect error reported on Android with no device passcode set and fallback allowed

pjf-dev:
## Describe the problem
When performing biometric authentication on Android with no biometry or device credential set, but fallback credential allowed, the error `biometryNotEnrolled` is reported. While this error is *technically* correct, the expected error would be `passcodeNotSet`, closer matching the actual error and the behavior of other platforms.

## Steps to reproduce
In my case, I tested this using the `examples/api` app, but general reproduction steps are as follows:
- Run a Tauri app that uses the biometric plugin on an Android device or simulator; ensure that biometry is not enrolled and no device passcode is set.
- Call the authenticate command with `allowDeviceCredential` set to true
- Observe the reported error is `biometryNotEnrolled` as opposed to `passcodeNotSet`

## Cause
In my testing, I found that the following seem to be the cause of the issue:
- Early disabling of fallback credential if `manager.isDeviceSecure` returns false that prevents setting `Authenticators.DEVICE_CREDENTIAL` as a valid authenticator -- This is actually intended behavior and isn't the root cause as seen next
- The Android API itself reports the error as `ERROR_NO_BIOMETRICS` even after allowing device credential on an insecure device (`manager.isDeviceSecure == false`). It seems this is the intended behavior of the Android API, but nevertheless should be corrected in the biometric plugin as `passcodeNotSet` more accurately describes the error.

## Proposed solution
Override the error code and message to `ERROR_NO_DEVICE_CREDENTIAL` in the event that a fallback credential is allowed, but error code `ERROR_NO_BIOMETRICS` was received.

## Additional notes
Also corrected the Android module mapping `ERROR_NO_DEVICE_CREDENTIAL` to `noDeviceCredential` when it should be `passcodeNotSet`.

### DIFF
--- a/.changes/fix-bio-fallback-error-android.md
+++ b/.changes/fix-bio-fallback-error-android.md
@@ -1,0 +1,6 @@
+---
+"biometric": patch:bug
+"biometric-js": patch:bug
+---
+
+Fix biometric plugin reporting an inconsistent and incorrect error on Android when no device passcode was set.
--- a/plugins/biometric/android/src/main/java/BiometricActivity.kt
+++ b/plugins/biometric/android/src/main/java/BiometricActivity.kt
@@ -38,41 +38,38 @@ class BiometricActivity : AppCompatActivity() {
         var title = intent.getStringExtra(BiometricPlugin.TITLE)
         val subtitle = intent.getStringExtra(BiometricPlugin.SUBTITLE)
         val description = intent.getStringExtra(BiometricPlugin.REASON)
-        allowDeviceCredential = false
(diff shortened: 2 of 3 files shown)


pjf-dev:
Made a slight oversight in my first solution where I put ALL the code dealing with setting the `BiometricManager.Authenticators` on API >= 30 into a single if checked by `allowDeviceCredential`. This caused `builder.setAllowedAuthenticators(authenticators)` to only be called when fallback credentials were allowed even though we still needed to supply `BiometricManager.Authenticators.BIOMETRIC_WEAK` as an authenticator to the builder.

Fixed the issue by reverting most of the code and using a separate variable to hold the supplied `allowDeviceCredential` argument needed for determining the proper error type.

PR #2650, opened 2025-04-21
feat: add the ability to check if the app is autostart

### LINKED ISSUE #2568: [feat][autostart] added `is_autostart` method
Hopefully a similar method can be added to determine if the application starts automatically.

ayangweb:
Resolved #2568

### DIFF
--- a/plugins/autostart/src/lib.rs
+++ b/plugins/autostart/src/lib.rs
@@ -98,6 +98,17 @@ async fn is_enabled(manager: State<'_, AutoLaunchManager>) -> Result<bool> {
     manager.is_enabled()
 }
 
+#[command]
+async fn is_autostart(manager: State<'_, AutoLaunchManager>) -> Result<bool> {
+    let process_args: Vec<String> = std::env::args().collect();
+
+    let passed_args = manager.0.get_args();
+
+    let is_autostart = passed_args.iter().all(|arg| process_args.contains(arg));
+
+    Ok(is_autostart)
+}
+
 #[derive(Default)]
 pub struct Builder {
     #[cfg(target_os = "macos")]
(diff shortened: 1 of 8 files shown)


plugins/autostart/src/lib.rs
```diff
@@ -98,6 +98,17 @@ async fn is_enabled(manager: State<'_, AutoLaunchManager>) -> Result<bool> {
...
+#[command]
+async fn is_autostart(manager: State<'_, AutoLaunchManager>) -> Result<bool> {
+    let process_args: Vec<String> = std::env::args().collect();
+
+    let passed_args = manager.0.get_args();
+
+    let is_autostart = passed_args.iter().all(|arg| process_args.contains(arg));
+
+    Ok(is_autostart)
+}
```
Legend-Master, line 110: This is just checking if all the args used to launch the program match the ones we passed into the plugin builder, this wouldn't work if we didn't pass in anything or if we changed the args

ayangweb: This can be used to determine whether the app was launched automatically or manually during its first startup. In the case of a manual launch, these parameters are usually not present.

Legend-Master: Well, you can pass in nothing in the builder, and this wouldn't work in that case, also when you relaunch/restart your app with `app.restart`, it will include all the args you previously set and that's going to be another false positive

ayangweb: Sorry, I really didn't think about that.

ayangweb: Are there any other approaches to determine whether an application is set to launch automatically at startup?

Legend-Master: I think it's fine to use command line args for this purpose, maybe a function in the builder that adds in a special cli arg to the args and we can detect if that cli arg is present or not, about the restart, that one would probably be a bit more involved since we don't have a way to change the restart cli args right now

ayangweb: https://github.com/tauri-apps/plugins-workspace/issues/2568#issuecomment-2750801350

Legend-Master: If that's an opt-in, I'm fine with that 🙃, or we could provide some docs for people to do it in their own apps for now

PR #2655, opened 2025-04-22
feat(http): add status code 101, 103 for null body

asomethings:
followup for: #2647 and #2636

## Description

For maximum compatibility I think it should follow [fetch spec](https://fetch.spec.whatwg.org/#null-body-status)

Though [Chromium](https://github.com/chromium/chromium/blob/main/third_party/blink/renderer/core/fetch/response.cc#L107) only handles 101, 204, 205, 304

And [Firefox](https://searchfox.org/mozilla-central/source/dom/fetch/Response.cpp#258) only handles 204, 205, 304


FabianLars:
Ah didn't check chromium and just Firefox but iirc in Firefox you can't even construct Responses with 10X statuses so I didn't bother mentioning them in that pr (which was dumb)

asomethings:
@FabianLars To improve performance, should we consider using an if statement instead? [Benchmark](https://measurethat.net/Benchmarks/Show/34301/0/if-vs-includes)

FabianLars:
Hmm I don't think this difference matters much.

amrbashir (approved):

PR #2657, opened 2025-04-23
fix(single-instance): Check if pointer is null for V1

mrjackwills:
This just adds the [fix](https://github.com/tauri-apps/plugins-workspace/pull/2452) that was merged into the V2 version of the plugin into the V1 version.

Tested working on Windows 11 with `rust 1.86.0`


FabianLars (approved):
thanks :)

PR #2695, opened 2025-05-14
fix(opener-js): allow URL type

ShaunSHamilton:
This updates the JS type to allow `URL` for the `openUrl` IPC.

As seen here, using a URL already serializes correctly:

https://github.com/tauri-apps/plugins-workspace/blob/a6b854032d0b10f0f17c4ffa6bdf4a05429e05fb/plugins/opener/guest-js/init.ts#L45

I thought this could either be `patch` or `minor`. Decided on `patch`, because it does not feel much like a feature.

### DIFF
--- a/.changes/opener-url-type.md
+++ b/.changes/opener-url-type.md
@@ -1,0 +1,5 @@
+---
+"opener-js": patch
+---
+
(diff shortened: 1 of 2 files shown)


Legend-Master (approved):
Thanks

PR #2707, opened 2025-05-19
feat(autostart): add prefered name configuration for auto start entry

Tunglies:
### Add Prefered Name Configuration for AutoStart Entry  

This PR introduces the ability to customize the application name used in auto start entries by adding a `PreferedName` enum to support different naming strategies.  

#### Key Changes:  
- Added `PreferedName` enum with variants:  
  - `PackageInfoName` (default) – Uses name from Tauri's package info  
  - `PackageInfoCrateName` – Uses the crate name from `Cargo.toml`  
  - `ConfigIdentifier` – Uses the identifier from Tauri's runtime config  
  - `ConfigProductName` Uses the product_name from Tauri's runtime config
  - `Custom(String)` – Allows specifying a custom name  
- Implemented `Default` trait for `PreferedName` (defaults to `PackageInfoName`)  
- Added `prefered_name` builder method with documentation and usage examples  
- Integrated prefered name selection into auto start configuration  

#### Usage Example:  
```rust  
Builder::new()
     .prefered_name(PreferedName::PackageInfoName)
     .build();

Builder::new()
    .prefered_name(PreferedName::Custom("My Custom Name".into()))
     .build();
```  

This enhancement provides flexibility for apps that need to control their display name in system auto start entries, supporting use cases like branding adjustments or multi-instance differentiation.

### DIFF
--- a/.changes/autostart-feature.md
+++ b/.changes/autostart-feature.md
@@ -1,5 +1,6 @@
 ---
 autostart: minor
+autostart-js: minor
 ---
 
 Added a new builder method app_name() to allow customizing the application name in the autostart entry.
--- a/.changes/autostart-feature.md
+++ b/.changes/autostart-feature.md
@@ -1,0 +1,5 @@
+---
+autostart: minor
+---
+
+Added a new builder method app_name() to allow customizing the application name in the autostart entry.
--- a/plugins/autostart/src/lib.rs
+++ b/plugins/autostart/src/lib.rs
@@ -13,7 +13,9 @@
 use auto_launch::{AutoLaunch, AutoLaunchBuilder};
 use serde::{ser::Serializer, Serialize};
 use tauri::{
-    command, plugin::{Builder as PluginBuilder, TauriPlugin}, AppHandle, Manager, Runtime, State
+    command,
+    plugin::{Builder as PluginBuilder, TauriPlugin},
+    AppHandle, Manager, Runtime, State,
 };
 
 use std::env::current_exe;
@@ -27,15 +29,15 @@ pub enum MacosLauncher {
     AppleScript,
 }
 
-#[derive(Debug, Clone)]
+#[derive(Default)]
 pub enum PreferedName {
-    PackageInfoName,
-    PackageInfoCrateName,
-    ConfigIdentifier,
-    ConfigProductName,
-    Custom(String)
+    #[default]
+    AppName,
+    CrateName,
+    Identifier,
+    ProductName,
+    Custom(String),
 }
-
 
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
@@ -54,29 +56,25 @@ impl Serialize for Error {
     }
 }
 
-impl Default for PreferedName {
-    fn default() -> Self {
-        PreferedName::PackageInfoName
-    }
-}
-
 impl PreferedName {
     pub fn get_value<R: Runtime>(&self, app: &AppHandle<R>) -> String {
         match self {
-            PreferedName::PackageInfoName => {
-                app.package_info().name.to_string()
-            },
-            PreferedName::PackageInfoCrateName => {
-                app.package_info().crate_name.to_string()
-            },
-            PreferedName::ConfigIdentifier => {
-                app.config().identifier.to_string()
-            },
-            PreferedName::ConfigProductName => {
-                app.config().product_name.clone().unwrap_or("TauriAPP".to_string())
-            },
+            PreferedName::AppName => app.package_info().name.to_string(),
+            PreferedName::CrateName => app.package_info().crate_name.to_string(),
+            PreferedName::Identifier => app.config().identifier.to_string(),
+            PreferedName::ProductName => app
+                .config()
+                .product_name
+                .clone()
+                .unwrap_or("TauriAPP".to_string()),
             PreferedName::Custom(s) => s.clone(),
         }
+    }
+}
+
+impl<S: Into<String>> From<S> for PreferedName {
+    fn from(name: S) -> Self {
+        PreferedName::Custom(name.into())
     }
 }
 
@@ -137,7 +135,7 @@ pub struct Builder {
     #[cfg(target_os = "macos")]
     macos_launcher: MacosLauncher,
     args: Vec<String>,
-    prefered_name: PreferedName,
+    app_name: PreferedName,
 }
 
 impl Builder {
@@ -189,23 +187,23 @@ impl Builder {
         self
     }
 
-    /// Sets the preferred name to be used for the auto start entry.
+    /// Sets the app name to be used for the auto start entry.
     ///
     /// ## Examples
     ///
     /// ```no_run
     /// Builder::new()
-    ///     .prefered_name(PreferedName::PackageInfoName)
+    ///     .app_name(PreferedName::AppName)
     ///     .build();
     /// ```
     ///
     /// ```no_run
     /// Builder::new()
-    ///     .prefered_name(PreferedName::Custom("My Custom Name".into()))
+    ///     .app_name("My Custom Name"))
     ///     .build();
     /// ```
-    pub fn prefered_name(mut self, prefered_name: PreferedName) -> Self {
-        self.prefered_name = prefered_name;
+    pub fn app_name(mut self, app_name: impl Into<PreferedName>) -> Self {
+        self.app_name = app_name.into();
         self
     }
 
@@ -215,7 +213,7 @@ impl Builder {
             .setup(move |app, _api| {
                 let mut builder = AutoLaunchBuilder::new();
 
-                let prefered_name = self.prefered_name.get_value(&app);
+                let prefered_name = self.app_name.get_value(app);
                 builder.set_app_name(&prefered_name);
 
                 builder.set_args(&self.args);
@@ -274,7 +272,6 @@ impl Builder {
 pub fn init<R: Runtime>(
     #[allow(unused)] macos_launcher: MacosLauncher,
     args: Option<Vec<&'static str>>,
-    prefered_name: Option<PreferedName>
 ) -> TauriPlugin<R> {
     let mut builder = Builder::new();
     if let Some(args) = args {
@@ -284,6 +281,5 @@ pub fn init<R: Runtime>(
     {
         builder = builder.macos_launcher(macos_launcher);
     }
-    builder = builder.prefered_name(prefered_name.unwrap_or_default());
     builder.build()
 }
--- a/plugins/autostart/src/lib.rs
+++ b/plugins/autostart/src/lib.rs
@@ -164,8 +164,8 @@ impl Builder {
     ///     .app_name("My Custom Name"))
     ///     .build();
     /// ```
-    pub fn app_name(mut self, app_name: &str) -> Self {
-        self.app_name = Some(app_name.to_string());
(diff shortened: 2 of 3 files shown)


plugins/autostart/src/lib.rs
```diff
@@ -216,6 +274,7 @@ impl Builder {
 pub fn init<R: Runtime>(
     #[allow(unused)] macos_launcher: MacosLauncher,
     args: Option<Vec<&'static str>>,
+    prefered_name: Option<PreferedName>
```
Legend-Master: We can not add new parameters to this function or it will be a breaking change, let's keep  it only in the builder

plugins/autostart/src/lib.rs
```diff
@@ -46,6 +54,32 @@ impl Serialize for Error {
...
+            PreferedName::ConfigIdentifier => {
+                app.config().identifier.to_string()
+            },
+            PreferedName::ConfigProductName => {
+                app.config().product_name.clone().unwrap_or("TauriAPP".to_string())
+            },
+            PreferedName::Custom(s) => s.clone(),
+        }
+    }
+}
```
Legend-Master: We can implement a `From` trait for the builder to be more ergonomic

```suggestion
impl PreferedName {
    pub fn get_value<R: Runtime>(&self, app: &AppHandle<R>) -> String {
        match self {
            PreferedName::PackageInfoName => {
                app.package_info().name.to_string()
            },
            PreferedName::PackageInfoCrateName => {
                app.package_info().crate_name.to_string()
            },
            PreferedName::ConfigIdentifier => {
                app.config().identifier.to_string()
            },
            PreferedName::ConfigProductName => {
                app.config().product_name.clone().unwrap_or("TauriAPP".to_string())
            },
            PreferedName::Custom(s) => s.clone(),
        }
    }
}

impl<S: Into<String>> From<S> for PreferedName {
    fn from(name: S) -> Self {
        PreferedName::Custom(name.into())
    }
}
```

plugins/autostart/src/lib.rs
```diff
@@ -154,12 +189,35 @@ impl Builder {
...
+    ///     .prefered_name(PreferedName::PackageInfoName)
+    ///     .build();
+    /// ```
+    ///
+    /// ```no_run
+    /// Builder::new()
+    ///     .prefered_name(PreferedName::Custom("My Custom Name".into()))
+    ///     .build();
+    /// ```
+    pub fn prefered_name(mut self, prefered_name: PreferedName) -> Self {
```
Legend-Master:
```suggestion
    pub fn prefered_name(mut self, prefered_name: impl Into<PreferedName>) -> Self {
```

plugins/autostart/src/lib.rs
```diff
@@ -154,12 +189,35 @@ impl Builder {
...
+    ///
+    /// ```no_run
+    /// Builder::new()
+    ///     .prefered_name(PreferedName::PackageInfoName)
+    ///     .build();
+    /// ```
+    ///
+    /// ```no_run
+    /// Builder::new()
+    ///     .prefered_name(PreferedName::Custom("My Custom Name".into()))
```
Legend-Master:
```suggestion
    ///     .prefered_name("My Custom Name")
```

plugins/autostart/src/lib.rs
```diff
@@ -103,6 +137,7 @@ pub struct Builder {
     #[cfg(target_os = "macos")]
     macos_launcher: MacosLauncher,
     args: Vec<String>,
+    prefered_name: PreferedName,
```
Legend-Master: Since this is for the auto start entry name corresponding to `set_app_name`, maybe we could just call it `app_name`

plugins/autostart/src/lib.rs
```diff
@@ -29,6 +27,16 @@ pub enum MacosLauncher {
...
 }
 
+#[derive(Debug, Clone)]
+pub enum PreferedName {
+    PackageInfoName,
+    PackageInfoCrateName,
+    ConfigIdentifier,
+    ConfigProductName,
+    Custom(String)
+}
```
Legend-Master: To be honest, I feel like we could just have 2 variants, a custom one, and a default one called `AppName`, and we can document that saying it's corresponding to the `productName` in the tauri config file and falls back to the cargo name

If a user wants to customize this further with thing like product name, they can use the builder in the `setup` hook, and they have the `App` available to them there

Legend-Master (changes requested):

Tunglies:
Thanks @Legend-Master  for review.

I kept the original init api which not brings a breaking change.
```rust
pub fn init<R: Runtime>(
    #[allow(unused)] macos_launcher: MacosLauncher,
    args: Option<Vec<&'static str>>,
)
```
To setup a app name with builder method
```rust
fn main() {
    tauri::Builder::default()
        .setup(|app| {
            app.app_handle().plugin(
                tauri_plugin_autostart::Builder::new()
                    .app_name(PreferedName::CrateName)
                    // .app_name("MyTauriAPP")
                    .build(),
            )?;

            Ok(())
        })
}
```
It supports 5 naming variants: default (`AppName`) (package.json), `CrateName` (Cargo.toml), `Identifier` and `ProductName` (Tauri config), and a customizable `String` override.  
 ```rust
#[derive(Default)]
pub enum PreferedName {
    #[default]
    AppName,
    CrateName,
    Identifier,
    ProductName,
    Custom(String),
}
```
The extra variants improve configuration file synchronization, which is a useful feature.

File: plugins/autostart/src/lib.rs (hunk 6 above)
Comment (FabianLars) [reply plugins/autostart/src/lib.rs]: > To be honest, I feel like we could just have 2 variants

I'd just use no enum. Have the current default be the default and let them specify a string if desired. Getting the desired value from tauri or cargo is imo easy enough to not have to deal with the logic ourselves.

**Comment by Legend-Master** [comment]
> The extra variants improve configuration file synchronization, which is a useful feature.

I can see it's more ergonomic to use when you need them, but I can't see when would you want to use them to be honest, for the crate name, you can just do `.app_name(env!("CARGO_CRATE_NAME"))`, for the id, normally we should use a display name instead of a code name, and for the product name, isn't it just the app name but fallback to `TauriAPP`?

**Comment by Tunglies** [comment]
> > The extra variants improve configuration file synchronization, which is a useful feature.
> 
> I can see it's more ergonomic to use when you need them, but I can't see when would you want to use them to be honest, for the crate name, you can just do `.app_name(env!("CARGO_CRATE_NAME"))`, for the id, normally we should use a display name instead of a code name, and for the product name, isn't it just the app name but fallback to `TauriAPP`?

The `PreferedName::CrateName` variant is redundant since it can be directly achieved via `.app_name(env!("CARGO_CRATE_NAME"))`, making it removable.

On MacOS, the plist file path is generated based on `app_name`, resulting in `/Users/user/Library/LaunchAgents/{app_name}.plist` .
https://github.com/zzzgydi/auto-launch/blob/fedaeba7a5ee6a5124b056907e03c399acd6df27/src/macos.rs#L179-L181

and for LaunchAgent, the `Label` field typically uses the app's identifier for unique identification. Applications control their LaunchAgent naming through the `CFBundleIdentifier` specified in `Info.plist` .
https://github.com/zzzgydi/auto-launch/blob/fedaeba7a5ee6a5124b056907e03c399acd6df27/src/macos.rs#L95-L111

The `Config` structure is defines `product_name: Option<String>` so I added fallback just in case.
https://github.com/tauri-apps/tauri/blob/dbcfaa18d7a0a4b41fbdde5c16616cce1b702ac6/crates/tauri-utils/src/config.rs#L2941-L2948

--- 

```rust
#[derive(Default)]
pub enum PreferedName {
   #[default]
   AppName,
   Identifier,
   Custom(String),
}
```

Keep only these three variants: default to `AppName` (package.json), a variant using `Identifier`(Tauri config, particularly useful for macOS), and a Custom option allowing users to define their own logic. 

Alternatively, remove the Identifier variant if you think it’s too cumbersome to handle.
So there would be just None or String, None is default to `AppName` or setting by String `.app_name(String)`

FabianLars:
The thing here is that devs themselves can use `app.config().identifier` as well. It's not like we don't want them to not have a way to use the identifier without hardcoding it, it's just that the code logic of the enum can be used in their code just as well.
Imo that's good enough to keep our api small and simple 🤷

Tunglies:
Thanks @Legend-Master and @FabianLars for reviews.

Let's keep the api small and simple.

The only one additional function for builder is
```rust
/// Sets the app name to be used for the auto start entry.
///
/// ## Examples
///
/// ```no_run
/// Builder::new()
///     .app_name("My Custom Name"))
///     .build();
/// ```
pub fn app_name(mut self, app_name: &str) -> Self {
    self.app_name = Some(app_name.to_string());
    self
}
```
no more, no less.

plugins/autostart/src/lib.rs
```diff
@@ -154,12 +155,29 @@ impl Builder {
...
+    ///
+    /// ```no_run
+    /// Builder::new()
+    ///     .app_name("My Custom Name"))
+    ///     .build();
+    /// ```
+    pub fn app_name(mut self, app_name: &str) -> Self {
+        self.app_name = Some(app_name.to_string());
+        self
+    }
```
Legend-Master:
```suggestion
    pub fn app_name<S: Into<String>(mut self, app_name: S) -> Self {
        self.app_name = Some(app_name.into());
        self
    }
```

FabianLars:
Thank you! And sorry for being kinda harsh about your ideas. In theory I like them but with the very limited resources we have for maintenance I try to keep things simple like that.

plugins/autostart/src/lib.rs
```diff
@@ -154,12 +155,29 @@ impl Builder {
...
+
     pub fn build<R: Runtime>(self) -> TauriPlugin<R> {
         PluginBuilder::new("autostart")
             .invoke_handler(tauri::generate_handler![enable, disable, is_enabled])
             .setup(move |app, _api| {
                 let mut builder = AutoLaunchBuilder::new();
-                builder.set_app_name(&app.package_info().name);
+
+                let app_name = self.app_name.unwrap_or(app.package_info().name.clone());
+                builder.set_app_name(&app_name);
```
Legend-Master:
```suggestion
                let app_name = self
                    .app_name
                    .as_ref()
                    .unwrap_or_else(|| &app.package_info().name);
                builder.set_app_name(app_name);
```

Legend-Master (changes requested):
Generally looks good, just a few nit picks, thanks for the contribution

Also could you add a change file? 

Tunglies:
> Generally looks good, just a few nit picks, thanks for the contribution
> 
> Also could you add a change file?
> 
> > https://github.com/tauri-apps/plugins-workspace/blob/v2/.changes/readme.md

I added the change file. Let's go

.changes/autostart-feature.md
```diff
@@ -0,0 +1,5 @@
+---
+autostart: minor
```
Legend-Master, line 2: We bump them at the same time, so 🙃

```suggestion
autostart: minor
autostart-js: minor
```

Tunglies: Sry, I didn't noticed it. :D

Legend-Master: No worries, it's kinda our fault since almost nobody did it right for the first time 😂

Legend-Master (approved):
Thanks!

Issue #2620, opened 2025-04-11
[window-state] Maybe the `restore_state` API should use `LogicalSize` (not `PhysicalSize`) when restoring the window size


1111mp:
### Describe the bug:

My screen information (Changing the system language settings requires restarting the computer, so please forgive the Chinese interface):
A main screen and an external display
![Image](https://github.com/user-attachments/assets/cd0c4078-53d3-4eb8-bc88-2b6db9fe5ece)

- First I run the app and drag it into the external screen, it works fine
![Image](https://github.com/user-attachments/assets/5521e83e-a429-4b33-8973-c8f76f1afd2a)

- Then I restart the app and the window size will be smaller
![Image](https://github.com/user-attachments/assets/2c2f6676-f33c-4426-8403-400aefaf9c4c)

- This is the information saved in `.window-state.json`:
  ```json
  {
    "main": {
      "width": 800,
      "height": 600,
      "x": 2142,
      "y": 196,
      "prev_x": 2143,
      "prev_y": 196,
      "maximized": false,
      "visible": true,
      "decorated": true,
      "fullscreen": false
    }
  }
  ```

The width and height it saves are the logical pixel dimensions of the declared window size `tauri.conf.json`:
```json
{
  "$schema": "https://schema.tauri.app/config/2",
  // ...
  "app": {
    "windows": [
      {
        "title": "tauri-app",
        "width": 800,
        "height": 600
      }
    ],
    "security": {
      "csp": null
    }
  },
  // ...
}
```

When I changed it to `LogicalSize`, the window size became correct.

https://github.com/tauri-apps/plugins-workspace/blob/c698e7259452d8556c3533ed53dbda4f6c2ca5f4/plugins/window-state/src/lib.rs#L210-L215

```rust
if flags.contains(StateFlags::SIZE) {
    self.set_size(LogicalSize {
        width: state.width,
        height: state.height,
    })?;
}
```
The window size is correct, but the window display position is not correct. I think this is because the position recorded in `.window-state.json` is the position of the application displayed on the external screen. Get The issue will be fixed when we support [Remember monitor](https://github.com/tauri-apps/plugins-workspace/issues/1282).

![Image](https://github.com/user-attachments/assets/7faf8fb6-f143-48c9-845e-97f82380072a)

### Full `tauri info` output:

```shell
[✔] Environment
    - OS: Mac OS 15.4.0 arm64 (X64)
    ✔ Xcode Command Line Tools: installed
[...]

PR #2727, opened 2025-05-30
feat(window-state): added `reset_window_state` method

1111mp:
Currently, in scenarios with multiple screens and different resolutions, the `window-state` plug-in may cause the window position to be incorrect and not as expected after restarting the application. For details, please refer to:

- #2620
- https://github.com/tauri-apps/tao/issues/816
- https://github.com/rust-windowing/winit/issues/2645

However, for now, this problem cannot be perfectly solved in the short term (this problem is more difficult on the `MacOS` platform because it allows users to customize almost any screen arrangement combination). So we can add a new method to allow users to quickly reset the window state when a window position error occurs, instead of continuing to restore the wrong window position with the previously saved state.


Legend-Master:
Honestly, I think a better way to do this is to have an option in your app that allows the user to disable the restoration

1111mp:
Disabling restoration can temporarily solve the problem of application window disorder, but once it is enabled, the problem will reappear. In this case, there is no reason to continue using the `window-state` plugin. Because if you want to solve this problem fundamentally, you need to manually delete the `window-state.json` file.

Legend-Master:
I mean, the reason you would delete that file is to prevent the restoration on next launch then what's the difference? Even if you deleted that file, if the bug is still there, it will still be recording the wrong values next time

1111mp:
Yes, but it is equivalent to giving users a chance to start over. In this way, users will not drag the window to another screen again based on their last experience. Although it can be done by disabling the button, the whole process requires users to manually operate several times:

- First click to disable the restore button
- Restart the application
- Then activate the restore button
- Finally, the next time you exit, the error information in the previous `window-state.json` will be overwritten

Of course this is just my opinion, and perhaps most people would not think so.

Legend-Master:
I still don't quite understand the steps you would like to have, I would imagine you want something like this:

1. The window gets restored to a weird position
2. User clicks on a button in your app that: deletes `window-state.json` and then restarts the app

But why don't they just drag the window to the position they want instead of doing this 'reset'? I mean it's just as easy isn't it? And at next launch, it wouldn't have mattered if you delete it or not, the new states will be saved anyways

1111mp:
Yes, the app's position can be dragged, but in some cases the app will disable the ability to resize the window.

Legend-Master:
I see, but wouldn't it be nicer to make that button change the window's position and size to default instead of doing this?

1111mp:
Yes, this is also a way. I just think that this method provided by `window-state` can reduce the confuse for developers.

I have forked a copy of the `window-state` code and maintained it in my own project, so it doesn't matter if this PR is not passed. Thank you for your timely reply. 🌹

PR #2757, opened 2025-06-10
fix(updater): headers option in `Updater.download`

### LINKED ISSUE #2754: Updater download and downloadAndInstall typescript definition are wrong
Hello, I think I found an issue in updater plugin.
There's a mismatch between the typescript description of a command and the actually expected parameters from the `download` command

The plugin declare the following command
https://github.com/tauri-apps/plugins-workspace/blob/f501bc2f1ce87d6a3a0fb286018d08a0a2d2c122/plugins/updater/src/commands.rs#L98-L104

But Typescript is:
https://github.com/tauri-apps/plugins-workspace/blob/f501bc2f1ce87d6a3a0fb286018d08a0a2d2c122/plugins/updater/guest-js/index.ts#L32-L41
https://github.com/tauri-apps/plugins-workspace/blob/f501bc2f1ce87d6a3a0fb286018d08a0a2d2c122/plugins/updater/guest-js/index.ts#L80-L83

Indeed, when supplying directly headers, and not an DownloadOptions object, it works.
If I submit a DownloadOptions object, I get an error.

Legend-Master:
Fix headers option in `Update.download` and `Update.downloadAndInstall` doesn't work with `Record<string, string> | Headers` types

Fix #2754


Legend-Master:
Note: didn't test

FabianLars (approved):

PR #2761, opened 2025-06-11
fix(updater): Escape current_exe args for nsis installer

### LINKED ISSUE #2752: [bug] Updating an app launched with arguments that have a space in them does not relaunch the app after update
On windows, if a tauri app is launched with arguments that have a space in them and an update is available, the updater will download and install the new version, but will fail to relaunch the app. Running `app.exe --with "Some Argument"` while an update is available will download, install and close the current instance of the app, but will not relaunch it.

FabianLars:
fixes #2752

I took a more conservative approach than i first wanted after reading Tony's comment. Really didn't want to touch the bundler or nsis utils at the moment and this whole cli arg stuff gives me a headache.
Maybe we can revise this whole logic when we refactor the code a bit (if that happens).

also i think we'll need to double check the msi escape method. looks like i missed args with spaces not starting with `-` ? Can't test it right now.

### DIFF
--- a/plugins/updater/src/updater.rs
+++ b/plugins/updater/src/updater.rs
@@ -1448,7 +1448,7 @@ mod tests {
 
     #[test]
     #[cfg(windows)]
-    fn it_escapes_correctly() {
+    fn it_escapes_correctly_for_msi() {
         use crate::updater::escape_msi_property_arg;
 
         // Explanation for quotes:
@@ -1491,6 +1491,49 @@ mod tests {
 
         for (orig, escaped) in cases.iter().zip(cases_escaped) {
             assert_eq!(escape_msi_property_arg(orig), escaped);
+        }
+    }
+
+    #[test]
+    #[cfg(windows)]
+    fn it_escapes_correctly_for_nsis() {
+        use crate::updater::escape_nsis_current_exe_arg;
+        use std::ffi::OsStr;
+
+        let cases = [
+            "something",
+            "--flag",
+            "--empty=",
+            "--arg=value",
+            "some space",                     // This simulates `./my-app "some string"`.
+            "--arg value", // -> This simulates `./my-app "--arg value"`. Same as above but it triggers the startsWith(`-`) logic.
+            "--arg=unwrapped space", // `./my-app --arg="unwrapped space"`
+            "--arg=\"wrapped\"", // `./my-app --args=""wrapped""`
+            "--arg=\"wrapped space\"", // `./my-app --args=""wrapped space""`
+            "--arg=midword\"wrapped space\"", // `./my-app --args=midword""wrapped""`
+            "",            // `./my-app '""'`
+        ];
+        // Note: These may not be the results we actually want (monitor this!).
+        // We only make sure the implementation doesn't unintentionally change.
+        let cases_escaped = [
+            "something",
+            "--flag",
+            "--empty=",
+            "--arg=value",
+            "\"some space\"",
+            "\"--arg value\"",
+            "\"--arg=unwrapped space\"",
+            "--arg=\\\"wrapped\\\"",
+            "\"--arg=\\\"wrapped space\\\"\"",
+            "\"--arg=midword\\\"wrapped space\\\"\"",
+            "\"\"",
+        ];
+
+        // Just to be sure we didn't mess that up
+        assert_eq!(cases.len(), cases_escaped.len());
+
+        for (orig, escaped) in cases.iter().zip(cases_escaped) {
+            assert_eq!(escape_nsis_current_exe_arg(&OsStr::new(orig)), escaped);
         }
     }
 }
--- a/plugins/updater/src/updater.rs
+++ b/plugins/updater/src/updater.rs
@@ -1371,19 +1371,38 @@ impl PathExt for PathBuf {
     }
 }
 
+// adapted from https://github.com/rust-lang/rust/blob/1c047506f94cd2d05228eb992b0a6bbed1942349/library/std/src/sys/args/windows.rs#L174
 #[cfg(windows)]
 fn escape_nsis_current_exe_arg(arg: &&OsStr) -> String {
-    let mut arg = arg.to_string_lossy().to_string();
+    let arg = arg.to_string_lossy();
+    let mut cmd: Vec<char> = Vec::new();
 
-    /* if arg.contains('"') {
-        arg = arg.replace('"', r#""""#);
-    } */
-
-    if arg.contains(' ') || arg.contains('\t') || arg.contains('/') {
-        arg = format!("\"{arg}\"");
+    let quote = arg.chars().any(|c| c == ' ' || c == '\t' || c == '/') || arg.is_empty();
+    let escape = true;
+    if quote {
+        cmd.push('"');
     }
-
-    arg
+    let mut backslashes: usize = 0;
+    for x in arg.chars() {
+        if escape {
+            if x == '\\' {
+                backslashes += 1;
+            } else {
+                if x == '"' {
+                    // Add n+1 backslashes to total 2n+1 before internal '"'.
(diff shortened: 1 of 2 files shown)


plugins/updater/src/updater.rs
```diff
@@ -1363,6 +1371,21 @@ impl PathExt for PathBuf {
     }
 }
 
+#[cfg(windows)]
+fn escape_nsis_current_exe_arg(arg: &&OsStr) -> String {
+    let mut arg = dbg!(arg.to_string_lossy().to_string());
```
```suggestion
    let mut arg = arg.to_string_lossy().to_string();
```

plugins/updater/src/updater.rs
```diff
@@ -1363,6 +1371,21 @@ impl PathExt for PathBuf {
...
 
+#[cfg(windows)]
+fn escape_nsis_current_exe_arg(arg: &&OsStr) -> String {
+    let mut arg = arg.to_string_lossy().to_string();
+
+    /* if arg.contains('"') {
+        arg = arg.replace('"', r#""""#);
+    } */
+
+    if arg.contains(' ') || arg.contains('\t') || arg.contains('/') {
```
Legend-Master: Do we need to escape `/`? (by the way, just from testing, even if you add a surrounding quote at `/SOME_RANDOM_ARGS`, NSIS's `GetOptions` function seems to still treat that as the end of the previous option)

plugins/updater/src/updater.rs
```diff
@@ -1363,6 +1371,21 @@ impl PathExt for PathBuf {
     }
 }
 
+#[cfg(windows)]
+fn escape_nsis_current_exe_arg(arg: &&OsStr) -> String {
```
amrbashir, line 1376: we should add unit tests for this function as well

plugins/updater/src/updater.rs
FabianLars: Yes, I added this because without it `app.exe path/to --whatever` breaks on the /

Didn't think about args starting with a /

Legend-Master: Looks like it, don't know how we escape `/` to be honest, but quotes don't seem to work for me

image

FabianLars: Yeah me neither, no symbol worked so I went back to quotes. That definitely worked for me in the updater though, will double check when I'm back home, maybe I reverted too much.

plugins/updater/src/updater.rs
FabianLars: Problem is I don't know what everything's actually supposed to look like (compared to MSI where i had a rough idea). This was all trial and error. So we can't really use them to confirm this PR but only to make sure it keeps its behavior over time.

FabianLars: But will do when I'm back

plugins/updater/src/updater.rs
Legend-Master: Hmm never mind, I think it works in `ShellExecuteW` 😂

image

When I execute it from the powershell, `DetailPrint "$CMDLINE"` gives me this which doesn't have the quotes in them

image

plugins/updater/src/updater.rs
```diff
@@ -1363,6 +1371,21 @@ impl PathExt for PathBuf {
...
+
+    /* if arg.contains('"') {
+        arg = arg.replace('"', r#""""#);
+    } */
+
+    if arg.contains(' ') || arg.contains('\t') || arg.contains('/') {
+        arg = format!("\"{arg}\"");
+    }
+
+    arg
```
Legend-Master:
```suggestion
    let arg = arg.to_string_lossy().to_string();
    if arg.contains(' ') || arg.contains('\t') || arg.contains('/') {
        format!("\"{arg}\"")
    } else {
        arg
    }
```

FabianLars: That's a reminder. Couldn't test " in args yet (well I tried but rusts args parsing is weird (It's probably me))

Legend-Master: I see, after some testing, I think only the one from the std really worked, basically, you need to escape `"` as `\"` (it also handles the `\"` case to `\\\"` I believe)

```suggestion
    let mut cmd: Vec<char> = Vec::new();

    let quote = arg.chars().any(|c| c == ' ' || c == '\t' || c == '/') || arg.is_empty();
    let escape = true;
    if quote {
        cmd.push('"');
    }

    let mut backslashes: usize = 0;
    for x in arg.chars() {
        if escape {
            if x == '\\' {
                backslashes += 1;
            } else {
                if x == '"' {
                    // Add n+1 backslashes to total 2n+1 before internal '"'.
                    cmd.extend((0..=backslashes).map(|_| '\\'));
                }
                backslashes = 0;
            }
        }
        cmd.push(x);
    }

    if quote {
        // Add n backslashes to total 2n before ending '"'.
        cmd.extend((0..backslashes).map(|_| '\\'));
        cmd.push('"');
    }
    cmd.into_iter().collect()
```

My mind is completely gone now after looking at them 😂, need some rests

FabianLars: Have a good rest haha

My problem in testing was that I couldn't even get the " into the initial app instance's env::args() call lol
I think I messed up my Powershell profile _again_

Anyway, will get back to this tomorrow

Legend-Master: try with `app.exe '"some args"'`, `'` escapes things inside in powershell

FabianLars: Yeah that's what I tried first though can't remember if I typed something inbetween.

FabianLars: okay either my powershell instance is broken beyond repair or that std args parser is missing a few edge cases (i may be testing stuff that aren't ever used idk). will just use this and call it a day.

plugins/updater/src/updater.rs
```diff
@@ -1363,6 +1371,40 @@ impl PathExt for PathBuf {
     }
 }
 
+// adapted from https://github.com/rust-lang/rust/blob/1c047506f94cd2d05228eb992b0a6bbed1942349/library/std/src/sys/args/windows.rs#L174
+#[cfg(windows)]
+fn escape_nsis_current_exe_arg(arg: &&OsStr) -> String {
+    let arg = arg.to_string_lossy();
+    let mut cmd: Vec<char> = Vec::new();
+
+    let quote = arg.chars().any(|c| c == ' ' || c == '\t' || c == '/') || arg.is_empty();
```
Legend-Master, line 1381: Let's add a comment here since `/` is not checked by rust standard library

plugins/updater/src/updater.rs
```diff
@@ -1493,4 +1493,47 @@ mod tests {
...
+            "--empty=",
+            "--arg=value",
+            "\"some space\"",
+            "\"--arg value\"",
+            "\"--arg=unwrapped space\"",
+            "--arg=\\\"wrapped\\\"",
+            "\"--arg=\\\"wrapped space\\\"\"",
+            "\"--arg=midword\\\"wrapped space\\\"\"",
+            "\"\"",
+        ];
```
Legend-Master, line 1531: Can we use `r#` here? I'm having a hard time reading through this 😂

FabianLars: Can't you just let me be lazy for once 😭 I copied the failed test results and they were in this format and I didn't want to read them either haha

Will do when I'm back home

Legend-Master: Actually, rust analyzer shows me the resolved the literal, cool, guess it's all good, the `r#` thing isn't that easy to read either when used with `"` anyways 😂 (e.g. `r#""""#`)

image

FabianLars: oh what, that's awesome. i never noticed that...

Legend-Master (approved):

amrbashir (approved):

PR #2773, opened 2025-06-16
chore(deps): fix brace expansion pnpm audit

Legend-Master:
Don't know why `pnpm update brace-expansion` would update a whole bunch of other things, so ended up editing it manually


FabianLars:
yeah, either `pnpm update` is stupid or we are x) couldn't make it work for usecases like this either.

FabianLars (approved):

PR #2784, opened 2025-06-20
feat: sqlite pool options

rpander93:
I use SQLite as storage for my Tauri application. By default, SQLx (the underlying driver) uses max. 10 parallel pool connections for SQLite.

When inserting a bunch of rows, I frequently get foreign key failure issues because the INSERT statements are spread across all connections and thus not in the right order. This plugin does not have proper transaction support (executing all statements on the same connection) and thus doesnt work either.

This PR implements the ability to limit the maximum number of connections in the pool. For me, by setting this to 1 connection maximum, I can ensure that all INSERTs are applied in the same order as I send them. Also, I can now start a transaction manually, because there's only ever 1 connection active.

Rust implementation based on https://github.com/tauri-apps/plugins-workspace/pull/2553


rpander93:
@FabianLars is there any interest of merging this?

FabianLars:
interest yes but i have 0 time for the sql plugin at the moment. also, some ci checks failed :)

rpander93:
okay. i've updated the PR to address the failing checks.

rpander93:
hi @FabianLars, any interest still?

FabianLars:
yesyes, i swear i'll get to it at some point 🙃 apologies for the delay.

PR #2787, opened 2025-06-22
feat: [cli] Pass optional args to get matches

### LINKED ISSUE #2785: Use tauri-plugin-cli with tauri-plugin-single-instance
Is there any way to use the `args` provided by `tauri-plugin-single-instance` with the CLI arg parsing of `tauri-plugin-cli`? I'd like to have a single instance, but my app has some CLI commands, and I don't want to lose them by having a single instance.

mikew:
This makes it possible for people to use the CLI plugin in tandem with the single-instance plugin:

```rust
.plugin(tauri_plugin_single_instance::init(|app, args, _cwd| {
  let matches = app.cli().matches(Some(args));
  match matches {
    Ok(matches) => {
      // Do something with matches
    }
    Err(e) => {
      eprintln!("Error parsing CLI arguments: {}", e);
    }
  }
}))
```

Note it's still a little awkward to actually use. If people want to send the matches to the client, they need to encode / decode them themselves as `emit_to` expects the data to be clone-able.

eg Here's what I have to do on the Rust side:

```rust
let serialized = serde_json::to_string(&matches).unwrap();
let _ = app.emit_to("main", "cli", serialized);
```

and on the client side:

```ts
listen<string>('cli', async (event) => {
  const matches: CliMatches = JSON.parse(event.payload)

  // Do something with matches
})
```

Closes #2785

### DIFF
--- a/.changes/add-optional-args-cli.md
+++ b/.changes/add-optional-args-cli.md
@@ -1,5 +1,0 @@
----
-"cli": minor
----
-
-Added `tauri_plugin_cli::matches_from`. This can be combined with the `args` passed to the callback of `tauri_plugin_single_instance` to parse  the command line arguments passed to subsequent instances of the application.
--- a/.changes/add-optional-args-cli.md
+++ b/.changes/add-optional-args-cli.md
@@ -1,0 +1,5 @@
+---
+"cli": minor
+---
+
+Added `tauri_plugin_cli::matches_from`. This can be combined with the `args` passed to the callback of `tauri_plugin_single_instance` to parse  the command line arguments passed to subsequent instances of the application.
--- a/plugins/cli/src/lib.rs
+++ b/plugins/cli/src/lib.rs
@@ -28,8 +28,8 @@ pub use parser::{ArgData, Matches, SubcommandMatches};
 pub struct Cli<R: Runtime>(PluginApi<R, Config>);
 
 impl<R: Runtime> Cli<R> {
-    pub fn matches(&self) -> Result<parser::Matches> {
-        parser::get_matches(self.0.config(), self.0.app().package_info())
+    pub fn matches(&self, args: Option<Vec<String>>) -> Result<parser::Matches> {
+        parser::get_matches(self.0.config(), self.0.app().package_info(), args)
     }
 }
 
@@ -45,7 +45,7 @@ impl<R: Runtime, T: Manager<R>> CliExt<R> for T {
 
 #[tauri::command]
 fn cli_matches<R: Runtime>(_app: AppHandle<R>, cli: State<'_, Cli<R>>) -> Result<parser::Matches> {
-    cli.matches()
+    cli.matches(None::<Vec<String>>)
 }
 
 pub fn init<R: Runtime>() -> TauriPlugin<R, Config> {
--- a/plugins/cli/src/lib.rs
+++ b/plugins/cli/src/lib.rs
@@ -28,8 +28,12 @@ pub use parser::{ArgData, Matches, SubcommandMatches};
 pub struct Cli<R: Runtime>(PluginApi<R, Config>);
 
 impl<R: Runtime> Cli<R> {
-    pub fn matches(&self, args: Option<Vec<String>>) -> Result<parser::Matches> {
-        parser::get_matches(self.0.config(), self.0.app().package_info(), args)
+    pub fn matches(&self) -> Result<parser::Matches> {
+        parser::get_matches(self.0.config(), self.0.app().package_info(), None)
+    }
+
+    pub fn matches_from(&self, args: Vec<String>) -> Result<parser::Matches> {
+        parser::get_matches(self.0.config(), self.0.app().package_info(), Some(args))
     }
 }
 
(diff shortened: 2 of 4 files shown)


plugins/cli/src/lib.rs
```diff
@@ -28,8 +28,8 @@ pub use parser::{ArgData, Matches, SubcommandMatches};
 pub struct Cli<R: Runtime>(PluginApi<R, Config>);
 
 impl<R: Runtime> Cli<R> {
-    pub fn matches(&self) -> Result<parser::Matches> {
-        parser::get_matches(self.0.config(), self.0.app().package_info())
+    pub fn matches(&self, args: Option<Vec<String>>) -> Result<parser::Matches> {
+        parser::get_matches(self.0.config(), self.0.app().package_info(), args)
     }
```
Legend-Master: Since this is a breaking change, let's add another function `matches_from` here

```suggestion
    pub fn matches(&self) -> Result<parser::Matches> {
        parser::get_matches(self.0.config(), self.0.app().package_info(), None)
    }

     pub fn matches_from(&self, args: Vec<String>) -> Result<parser::Matches> {
        parser::get_matches(self.0.config(), self.0.app().package_info(), Some(args))
    }
```

plugins/cli/src/lib.rs
```diff
@@ -45,7 +45,7 @@ impl<R: Runtime, T: Manager<R>> CliExt<R> for T {
 
 #[tauri::command]
 fn cli_matches<R: Runtime>(_app: AppHandle<R>, cli: State<'_, Cli<R>>) -> Result<parser::Matches> {
-    cli.matches()
+    cli.matches(None::<Vec<String>>)
```
Legend-Master:
```suggestion
    cli.matches()
```

Legend-Master (changes requested):
Also do you mind also add a change file like this one?

https://github.com/tauri-apps/plugins-workspace/blob/5642283dba7d3dbc256c41111d555be6bc5f7c41/.changes/add-global-flag-cli.md?plain=1#L6

mikew:
> Also do you mind also add a change file like this one?
> 
> https://github.com/tauri-apps/plugins-workspace/blob/5642283dba7d3dbc256c41111d555be6bc5f7c41/.changes/add-global-flag-cli.md?plain=1#L6

Can maybe scrounge together a docs update too if you think it would help!

Legend-Master:
About the clone trait thing, I don't mind we add derive `Clone` for `Matches` and its fields, just add it to `#[derive(Default, Debug, Serialize)]` like this `#[derive(Default, Debug, Serialize, Clone)]`

plugins/cli/src/parser.rs
```diff
@@ -79,7 +79,11 @@ impl Matches {
 ///     Ok(())
 ///   });
 /// ```
-pub fn get_matches(cli: &Config, package_info: &PackageInfo) -> crate::Result<Matches> {
+pub fn get_matches(
+    cli: &Config,
+    package_info: &PackageInfo,
+    args: Option<Vec<String>>,
+) -> crate::Result<Matches> {
```
mikew, line 86: @Legend-Master Curious and slightly bored here, hope you don't mind me asking. You called out the changes to `Cli.matches` as a breaking change, how is this also not considered a breaking change? Both `Cli.matches` and `parser::get_matches` are public, and neither are referenced anywhere in the usage guide.

mikew:
> About the clone trait thing, I don't mind we add derive `Clone` for `Matches` and its fields, just add it to `#[derive(Default, Debug, Serialize)]` like this `#[derive(Default, Debug, Serialize, Clone)]`

I tried that in my testing but Rust then complained about `Matches.args` not being clone-able:


```
Diagnostics:
1. the trait bound `ArgData: Clone` is not satisfied
   required for `HashMap<std::string::String, ArgData>` to implement `Clone` [E0277]
```

```
Diagnostics:
1. the trait bound `SubcommandMatches: Clone` is not satisfied
   required for `Box<SubcommandMatches>` to implement `Clone`
   1 redundant requirement hidden
   required for `std::option::Option<Box<SubcommandMatches>>` to implement `Clone` [E0277]
```

plugins/cli/src/parser.rs
mikew: Apologies, did not notice that the docs _do_ mention `Cli.matches`.

Legend-Master: `parser` is a private module, so `parser::get_matches` is not actually public, yeah I know it's kinda confusing 😂

Legend-Master:
> I tried that in my testing but Rust then complained about Matches.args not being clone-able:

You'll need to also add it to `SubcommandMatches` and `ArgData`

.changes/add-optional-args-cli.md
```diff
@@ -0,0 +1,5 @@
+---
+"cli": minor
```
Legend-Master:
```suggestion
"cli": minor
"cli-js": minor
```

Legend-Master (approved):
Thanks!

PR #2791, opened 2025-06-23
(store): Improved type-safety of JavaScript functions

theinfinit:
Proposition to overhaul the typescript definitions of `store` plugin. 

## Why?
Current approach is error prone and cumbersome to work with. Setters are basically untyped and getters require manual type provisioning for each call.

## Note
As far as TypeScript types go, this is a breaking change. 

## Disclaimer
Everything written by hand. No form of "AI" was involved. Feel free to ask questions about the code.

## Example

Example below is outlining the workflow and areas of improvement.

```ts
import { load } from '@tauri-apps/plugin-store';

// Each store has type-safety provided by a single source of truth.

// 1. Define types
// In the example we have: 4 keys and 3 possible types
type StateStore = {
  darkMode: boolean
  alwaysOnTop: boolean
  position: number
  'some-key': {
    value: number
  }
}

// 2. Pass type definition on load
const store = await load<StateStore>('store.json');

// 3. Setter 
// - Autocompletion of key's name,
// - Type of value automatically enforced by provided key
await store.set('alwaysOnTop', true); // (key: "alwaysOnTop", value: boolean)
await store.set('some-key', { value: 5 }); // (key: "some-key", value: { value: number; })

// 4. Getter 
// - Key name autocompletion,
// - Return type based on provided key.
const val = await store.get('some-key');
console.log(`[tauri-store] some-key:`, val); // { value: number } | undefined

await store.onKeyChange('darkMode', (value) => {
  console.log(`🚀 ~ store onKeyChange:`, value) // value: boolean | undefined
})

/* All possible types derived from the provided StateStore type:
  (number | boolean | { value: number })[]
 */
const storeValues = await store.values() 
console.log(`🚀 ~ storeValues:`, storeValues)

const storeEntries = await store.entries()
console.log(`🚀 ~ storeEntries:`, JSON.stringify(storeEntries))
```

### DIFF
--- a/plugins/store/guest-js/index.ts
+++ b/plugins/store/guest-js/index.ts
@@ -36,6 +36,10 @@ export type StoreOptions = {
   createNew?: boolean
 }
 
+interface StoreEntries {
+  [key: string]: unknown
+}
+
 /**
  * Create a new Store or load the existing store with the path.
  *
@@ -48,10 +52,10 @@ export type StoreOptions = {
  * @param path Path to save the store in `app_data_dir`
  * @param options Store configuration options
  */
-export async function load(
+export async function load<S extends StoreEntries>(
   path: string,
   options?: StoreOptions
-): Promise<Store> {
+): Promise<Store<S>> {
   return await Store.load(path, options)
 }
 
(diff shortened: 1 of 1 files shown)


Legend-Master:
Cool! I'm fan of this, but like you said, this is a breaking change, so we would probably need to wait for v3

mafiefa02:
I was just complaining about how I can't pass generics to the store. Great to know this is coming!

Issue #2593, opened 2025-04-03
[bug] Invoking download() locks main thread on Android


velocitysystems:
### Describe the problem

Invoking the [download()](https://v2.tauri.app/reference/javascript/upload/#download) method in [tauri-plugin-upload](https://github.com/tauri-apps/plugins-workspace/tree/v2/plugins/upload) locks the main thread when downloading on Android.
This issue does not occur on iOS, macOS or Windows.

### Describe the solution you'd like

Invoking the method should not lock the main thread.

### Alternatives considered

Use [tauri-plugin-download](https://github.com/silvermine/tauri-plugin-download) which is fully asynchronous and non-blocking.

### Additional context

N/A

PR #2838, opened 2025-07-11
fix(upload): fix download() locks main thread on Android

velocitysystems:
Fixed:
#2593 

**Key Changes**
- Add **Upload** demo to example project
- Add tests for `upload` method
- Use Tokio to spawn task when invoking `download` or `upload` command
- Fix bug where upload progress not properly reported from `file_to_body` function

**Screenshots**
image

Screenshot_1752595073

### DIFF
--- a/.changes/upload-locks-main-thread.md
+++ b/.changes/upload-locks-main-thread.md
@@ -1,6 +1,6 @@
 ---
-upload: minor
-upload-js: minor
+upload: patch
+upload-js: patch
 ---
 
 Fix `download` and `upload` locks main thread on Android.
--- a/.changes/upload-locks-main-thread.md
+++ b/.changes/upload-locks-main-thread.md
@@ -1,0 +1,7 @@
+---
+upload: minor
+upload-js: minor
+---
+
+Fix `download` and `upload` locks main thread on Android.
+Use Tokio to spawn task when invoking commands.
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ This repo and all plugins require a Rust version of at least **1.77.2**
 | [store](plugins/store)                         | Persistent key value storage.                                                                                                                                  | ✅  | ✅  | ✅  | ✅  | ✅  |
 | [stronghold](plugins/stronghold)               | Encrypted, secure database.                                                                                                                                    | ✅  | ✅  | ✅  | ?   | ?   |
 | [updater](plugins/updater)                     | In-app updates for Tauri applications.                                                                                                                         | ✅  | ✅  | ✅  | ❌  | ❌  |
-| [upload](plugins/upload)                       | Tauri plugin for file uploads through HTTP.                                                                                                                    | ✅  | ✅  | ✅  | ?   | ?   |
+| [upload](plugins/upload)                       | Tauri plugin for file uploads through HTTP.                                                                                                                    | ✅  | ✅  | ✅  | ✅  | ✅  |
 | [websocket](plugins/websocket)                 | Open a WebSocket connection using a Rust client in JS.                                                                                                         | ✅  | ✅  | ✅  | ?   | ?   |
 | [window-state](plugins/window-state)           | Persist window sizes and positions.                                                                                                                            | ✅  | ✅  | ✅  | ❌  | ❌  |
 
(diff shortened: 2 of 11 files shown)


velocitysystems:
Tested upload demo on the following platforms:
- macOS
- Windows^
- Android (emulator)
- iOS (simulator)

Verified no locking of the main (UI) thread occurs when downloading or uploading.

^ When testing on Windows, I ran into the issues described in #2852. Switching to the [Brownfield](https://v2.tauri.app/concept/inter-process-communication/brownfield/) pattern in a local branch allowed me to test and validate the behavior was correct.

velocitysystems:
@lucasfernog Fixed this issue.

FYI - I noticed mobile support isn't added in the example project; I had to manually run `tauri ios init` and `tauri android init`. Should this be included in another PR? I wasn't sure if there was a compelling reason why this isn't included.

velocitysystems:
Added tests for `upload` method.

README.md
```diff
@@ -32,8 +32,8 @@ This repo and all plugins require a Rust version of at least **1.77.2**
...
 | [stronghold](plugins/stronghold)               | Encrypted, secure database.                                                                                                                                    | ✅  | ✅  | ✅  | ?   | ?   |
-| [updater](plugins/updater)                     | In-app updates for Tauri applications.                                                                                                                         | ✅  | ✅  | ✅  | ❌  | ❌  |
-| [upload](plugins/upload)                       | Tauri plugin for file uploads through HTTP.                                                                                                                    | ✅  | ✅  | ✅  | ?   | ?   |
+| [updater](plugins/updater)                     | In-app updates for Tauri applications.                                                                                                                         | ✅  | ✅  | ✅  | ✅  | ✅  |
+| [upload](plugins/upload)                       | Tauri plugin for file downloads and uploads via HTTP.                                                                                                          | ✅  | ✅  | ✅  | ?   | ?   |
```
FabianLars: looks like those 2 lines are mixed up

plugins/upload/src/lib.rs
```diff
@@ -72,44 +72,50 @@ async fn download(
...
-    } else {
-        client.get(url)
-    };
-    // Loop trought the headers keys and values
-    // and add them to the request object.
-    for (key, value) in headers {
-        request = request.header(&key, value);
-    }
+    let url = url.to_string();
+    let file_path = file_path.to_string();
```
FabianLars: Might as well change the function arguments then. PathBuf implement Deserialize as well so if we need that later we could use that instead of a string.

plugins/upload/src/lib.rs
```diff
@@ -119,46 +125,53 @@ async fn upload(
...
-        .header(reqwest::header::CONTENT_LENGTH, file_len)
-        .body(file_to_body(on_progress, file));
-
-    // Loop through the headers keys and values
-    // and add them to the request object.
-    for (key, value) in headers {
-        request = request.header(&key, value);
-    }
+    let url = url.to_string();
+    let file_path = file_path.to_string();
```
FabianLars: same

FabianLars (changes requested):
thanks for the pr just a nitpick annotated and we'll also need a changefile :) (check .changes dir for examples)

README.md
velocitysystems: Good catch. Fixed.

plugins/upload/src/lib.rs
velocitysystems: Good idea. Done.

plugins/upload/src/lib.rs
velocitysystems: Done.

velocitysystems:
> thanks for the pr just a nitpick annotated and we'll also need a changefile :) (check .changes dir for examples)

Thanks for the review @FabianLars! Just pushed a commit to resolve your comments.

.changes/upload-locks-main-thread.md
```diff
@@ -0,0 +1,7 @@
+---
+upload: minor
+upload-js: minor
```
FabianLars:
```suggestion
upload: patch
upload-js: patch
```

FabianLars (approved):
thanks :)

lucasfernog:
amazing PR!

Issue #2829, opened 2025-07-04
(store) store|reload command has unexpected side-effect (self.cache.extend)


theinfinit:
https://github.com/tauri-apps/plugins-workspace/blob/ca3c3aa28afd302178270df48dc3f6df45fa373b/plugins/store/guest-js/index.ts#L395-L397

As the description of `reload` function suggests, it should load the on-disk state into memory. But what is actually happening is a merge operation (or, using Rust's vocabulary, extending HashMap that is already in memory).

### Example

After changing JSON externally as follows:
```diff
 {
   "one": {
     "someKey": "someValue"
   },
-  "two": {
+  "three": {
     "someKey": "someValue"
   }
 }
```

Instead of getting cache like this...
```json
{
  "one": {
    "someKey": "someValue"
  },
  "three": {
    "someKey": "someValue"
  }
}
```

...calling `reload` function will result in the following:
```json
{
  "one": {
    "someKey": "someValue"
  },
  "two": {
    "someKey": "someValue"
  },
  "three": {
    "someKey": "someValue"
  }
}
```

## Current implementation

When it comes to `store|load` function, using `cache.extend` is probably fine, but for `store|reload`, given the current description, it makes the current behavior quite confusing. 

```ts
await invoke('plugin:store|reload', { rid: this.rid })
```

https://github.com/tauri-apps/plugins-workspace/blob/ca3c3aa28afd302178270df48dc3f6df45fa373b/plugins/store/src/lib.rs#L212-L215

https://github.com/tauri-apps/plugins-workspace/blob/ca3c3aa28afd302178270df48dc3f6df45fa373b/plugins/store/src/store.rs#L502-L504

https://github.com/tauri-apps/plugins-workspace/blob/ca3c3aa28afd302178270df48dc3f6df45fa373b/plugins/store/src/store.rs#L287-L294

## Solutions

Depending on why the current implementation is using `cache.extend`, and assuming that my understanding is correct, I would like to ask, to update the description of `store|reload` and warn about the side-effect (as changing the current implementation [doesn't seem wise](https://xkcd.com/1172/)).

## Question

In the meantime, I have a question. Is there some obvious way I'm missing, to actually update the cache to reflect the on-disk state?

Legend-Master:
Honestly, I have no idea why it even works like that, it makes no sense to me for it to be additive 

Let's document this for now and maybe change the behavior in v3

PR #2843, opened 2025-07-12
docs(store): reload merges the value

Legend-Master:
Reference: #2829


Legend-Master:
Genuine question, does anyone know why it was designed to work like that or was that a mistake?

Legend-Master:
I see, it's from #295

FabianLars:
not my proudest moment though it made sense to me at that time. i mean to some degree it still does. their requirements back then were valid imo so maybe we need 2 different functions.

FabianLars:
but most of the api sucks a bit so maybe we should rethink the plugin from the ground up in v3

Legend-Master:
I'm thinking the more intuitive way of how it should work, is we reset the store to `defaults` first and then call `extend` to merge the values in `load`

@FabianLars @lucasfernog Any thoughts? Also should we change the behavior in v2 or wait for v3 for this? Asking because if the original intension was to making `load` work with `defaults`, this change will not break that usage, and unless someone actually uses the `load` to merge the on-disk values, we should be good

FabianLars:
since the pr you linked was from before v2 imo we shouldn't change the behavior now. but idk both behaviors are flawed tbh

Legend-Master:
> not my proudest moment though it made sense to me at that time. i mean to some degree it still does. their requirements back then were valid imo so maybe we need 2 different functions.

Yeah, that made sense to me as well, I didn't really thought about the defaults

> since the pr you linked was from before v2 imo we shouldn't change the behavior now. but idk both behaviors are flawed tbh

Same thoughts now, let's wait for v3, also making 2 functions sounds good

I'm thinking 2 potential use cases

1. merging with default: when you add in for example a new config setting, the new setting's default will not be covered up by the old config file on user's computer
2. on the other hand, say you have a form or something, with some pre-filled default values, and in this case you want to load the exact file on disk, the defaults don't matter here

Issue #2855, opened 2025-07-18
[store] When reading a store from an existing file, a logical error occurs if store defaults are set (they will be merged instead of replaced)


MorningBiteFan:
### Example

`    let store = StoreBuilder::new(app_handle, "config.json")
        .defaults(defaults)
        .build()?;`

Theoretically, if the config.json file exists, the store should be read from the file; otherwise, the default store should be used

If the config.json content is
```
{
  "1": "a"
}
```

defaults content is
```
{
  "2": "b"
}
```

After the program runs, the expected cache content is
```
{
  "1": "a"
}
```

but the actual content is
```
{
  "1": "a"，
  "2": "b"
}
```

### Reason:
I examined the source code and found the following statement in the `build_inner` method of `StoreBuilder`.

https://github.com/tauri-apps/plugins-workspace/blob/80d4d8e128922b022e76049189a657b71ce95a97/plugins/store/src/store.rs#L181-L209

Normally, `self.create_new` is false and the stores content is empty, so neither line 187 nor line 191 will execute.

Then, line 202 will load the defaults, and the `load` method will be called on line 208.

The current `load` method executes `self.cache.extend`, which merges the config.json file's content with the defaults.
https://github.com/tauri-apps/plugins-workspace/blob/80d4d8e128922b022e76049189a657b71ce95a97/plugins/store/src/store.rs#L287-L294

The issue regarding the `load` method's merging behavior was also mentioned in another post.
https://github.com/tauri-apps/plugins-workspace/issues/2829

The commit updating `self.cache.extend` in the `load` method has also been located; it was introduced to resolve https://github.com/tauri-apps/plugins-workspace/issues/296, Commit 06f3958 (https://github.com/x0f5c3/plugins-workspace/commit/06f39586afde4d6d45d02c6e9b3341fb8782b469).

### Question
Is there a way to fix this right now, or am I misunderstanding the correct usage of the store plugin?

Legend-Master:
Thanks for the digging! I was wondering about why it works like this, there's some documentation work going on at #2843, and we should probably rethink about this behavior

PR #2857, opened 2025-07-18
feat(store): load override defaults

Legend-Master:
Reference: #2829
Reference: #2855
Reference: https://github.com/tauri-apps/plugins-workspace/pull/2843#issuecomment-3088738636

Closes #2843

Add an new option `overrideDefaults` for creating/loading and reloading the store that overrides the store with the on-disk state, ignoring defaults

Also added support for setting `defaults` from the JavaScript API

For the `reload` function, we should change its behavior to resetting the store to `defaults` first and then calling `extend` to merge the values in `load` in v3, and document this in v2 for now


Legend-Master:
~~Didn't test yet, will mark as ready after that~~

Done

FabianLars:
should it be ignoreDefaults and load_without_defaults maybe? override to me sounds like we're still loading the defaults first and then overriding those with the ondisk state.

Legend-Master:
`override_defaults` makes more sense in `StoreBuilder` since we will still preserve the defaults if the file doesn't exist

We can change the name on the other ones though

FabianLars:
the store plugin is giving me such a headache and i'm not even working on it 🫠 
Just merge it when you're happy with it ig 😂

FabianLars (approved):

PR #2860, opened 2025-07-20
fix(single-instance): explicitly D-Bus name replacement

MorpheusXAUT:
Fix D-Bus name replacement logic on Linux to prevent multiple instances from acquiring the same well-known name.

This patch updates the `zbus` dependency to the latest compatible version (`^5.9`) and explicitly sets `RequestNameFlags` to ensure a second instance fails to acquire the D-Bus name when another one is already running.

This resolves an issue with previous versions of `zbus` resulting in deadlocks in connection name request tasks (see https://github.com/dbus2/zbus/pull/1431). The new default behavior of the library allows name replacement (by setting `ReplaceExisting` and ` DoNotQueue` flags, so we're explicitly disabling this behavior to ensure only a single instance can be spawned.

### DIFF
--- a/.changes/single-instance-dbus-name.md
+++ b/.changes/single-instance-dbus-name.md
@@ -1,0 +1,6 @@
+---
+"single-instance": patch
+---
+
+Fix D-Bus name replacement logic on Linux to prevent multiple instances from acquiring the same well-known name.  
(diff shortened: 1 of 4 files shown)


FabianLars (approved):
thx

Issue #2826, opened 2025-07-02
[Bug] `open` from `Dialog` plugin has issues with filters on android


ppmpreetham:
While using `open` from the Dialog plugin, only windows allows the usage of filters, but not Android.

 The following code snippet works on windows but not on android. 
The issue being the `filters`.
```js
const filePath = await open({
      filters: [
        {
          name: "HTML Files",
          extensions: ["html", "htm"],
        },
      ],
      multiple: false,
      directory: false,
});
```
while the below code snippet works fine on android 
```js
const filePath = await open({
      multiple: false,
      directory: false,
});
```

System info:
```bash
[✔] Environment
    - OS: Windows 10.0.22631 x86_64 (X64)
    ✔ WebView2: 137.0.3296.93
    ✔ MSVC: Visual Studio Build Tools 2019
    ✔ rustc: 1.87.0 (17067e9ac 2025-05-09)
    ✔ cargo: 1.87.0 (99624be96 2025-05-06)
    ✔ rustup: 1.28.2 (e4f3ad6f8 2025-04-28)
    ✔ Rust toolchain: stable-x86_64-pc-windows-msvc (default)
    - node: 20.10.0
    - pnpm: 9.10.0
    - yarn: 1.22.22
    - npm: 10.2.3

[-] Packages
    - tauri 🦀: 2.6.0, (outdated, latest: 2.6.2)
    - tauri-build 🦀: 2.3.0
    - wry 🦀: 0.52.1
    - tao 🦀: 0.34.0
    - @tauri-apps/api : 2.5.0 (outdated, latest: 2.6.0)
    - @tauri-apps/cli : 2.5.0 (outdated, latest: 2.6.2)

[-] Plugins
    - tauri-plugin-fs 🦀: 2.4.0
    - @tauri-apps/plugin-fs : 2.3.0 (outdated, latest: 2.4.0)
    - tauri-plugin-single-instance 🦀: 2.3.0
    - @tauri-apps/plugin-single-instance : not installed!
    - tauri-plugin-deep-link 🦀: 2.4.0
    - @tauri-apps/plugin-deep-link : 2.3.0 (outdated, latest: 2.4.0)
    - tauri-plugin-dialog 🦀: 2.3.0
    - @tauri-apps/plugin-dialog : 2.2.2 (outdated, latest: 2.3.0)
    - tauri-plugin-barcode-scanner 🦀: 2.3.0
[...]

PR #2866, opened 2025-07-22
fix: Custom extensions issue with `open` from `Dialog` plugin with filters (#2826)

ppmpreetham:
> [!WARNING]  
> THIS PR REQUIRES TESTING ON ANDROID

Now the following code snipped works on Android:

```js
const filePath = await open({
      filters: [
        {
          name: "HTML Files",
          extensions: ["html", "htm"],
        },
      ],
      multiple: false,
      directory: false,
});
```

### DIFF
--- a/.changes/fix-android-filter-extensions.md
+++ b/.changes/fix-android-filter-extensions.md
@@ -1,0 +1,6 @@
+---
+"dialog": patch
+"dialog-js": patch
+---
+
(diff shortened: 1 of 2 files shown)


FabianLars:
I didn't try this PR yet but it looks like now you filter the results _after_ the user selected the files, right?

FabianLars:
i did a bit of testing and ended up with this https://github.com/tauri-apps/plugins-workspace/pull/2871

PR #2870, opened 2025-07-23
fix(mobile): deeplinks

### LINKED ISSUE #123: [log] Separation between rust and JavaScript logs
Log messages from JavaScript are written like the following: 
`[2023-02-06][20:17:17][ERROR][INFO] message from JS!`
Instead of `ERROR` I would like the plugin to know that the log was written from JS, or even better in which component for exmaple `JS::App::MainPage`.
I also would like to get the ability to save rust logs to one file and js logs to other file, with the ability to naming these files as an option.

sosweetham:
## Issues faced 

The deep link mobile config had the following problems: 
- custom url schemes required a host, even when it might not be preferred 
- app link entitlements would be force generated even when only custom url schemes were being used, problematic for developers that only need custom url schemes support and not app link support 

### TL;DR

* **Custom URL schemes** now **no longer require a host**, unless you're using `"http"` or `"https"` schemes (which do).
* A new `app_link` flag lets you **explicitly opt-in** to AppLink/UniversalLink behavior.
* **Entitlements files are only generated** when `app_link: true`, avoiding unnecessary configuration.
* **iOS/macOS plist files are still manual** – developers must copy & register them themselves in `src-tauri`.
* **Build system is append-only**: previously generated properties aren’t removed automatically—clean them up manually if needed.

### Associated Domain Validation

The "host" config value is now optional to provide, the validation at build time will fail and panic if, the config declares the deep link with the default scheme, `["http", "https"]` (web link) which requires a host as declared at  
https://developer.android.com/training/app-links 
link-types-capabilities

even so, this is not enough information to determine whether the developer intends to support app links/universal links, which require additional logic like entitlements file generation on ios

thus a new `app_link` property has been introduced to determine whether the particular "AssociatedDomain" is an AppLink/UniversalLink or not. 

Apple does not support plain old non secure http transport, so the plugin will throw a warn at build time but will NOT panic. 

### Conditional Entitlements Generation

This leads to the next fix, entitlements file is now conditionally generated, not all types of deeplinks on ios/macos need those.

## Known Issues

Ios/Macos plist files are not generated automatically like android,~~and also no automatic registration is performed post build, developer must add the relevant `Info.plist` (mac) or `Info.ios.plist` (ios) to the `src-tauri` folder manually containing the associated schemes declared in the `tauri.conf.json` as can be observed from the example provided/updated~~ (they are now generated if the patch is applied in the cargo.toml via the commit used in this pr)

I was not able to bring ios to parity with android wrt intents, i tried AppDelegate but I just kept running into weird issues, so i just scratched that otherwise i would have loved a native plugin for all sorts of events data lol


lucasfernog:
can't we automatically infer `app_link: true` from a host being set?

sosweetham:
> can't we automatically infer `app_link: true` from a host being set?

we can infer that it's a web link, but we still cannot say that its most definitely an app link, as suggested by the linked resource

sosweetham:
missed that mb 😛

FabianLars:
> can't we automatically infer app_link: true from a host being set?

if we wouldn't want to support web links (app links without the verifification) we could derive it from the scheme.

The current appLink config is also a breaking change. Your change makes autoVerify opt-in while before it was always enabled so if people were to simply bump the plugin version their app links would break.

FabianLars:
> Ios/Macos plist files are not generated automatically like android, and also no automatic registration is performed post build, developer must add the relevant Info.plist (mac) or Info.ios.plist (ios) to the src-tauri folder manually containing the associated schemes declared in the tauri.conf.json as can be observed from the example provided/updated

Only true for iOS. macOS is handled by tauri's cli.

FabianLars:
> A new app_link flag lets you explicitly opt-in to AppLink/UniversalLink behavior.

Not for universal links since ios doesn't support android's "inbetween" web links. You either have a custom scheme or use verified universal links. I'm not 100% sure if this is actually true, that's just based on what apple says in their docs, the actual behavior may be different 🤷

sosweetham:
> > can't we automatically infer app_link: true from a host being set?
> 
> if we wouldn't want to support web links (app links without the verifification) we could derive it from the scheme.
> 
> The current appLink config is also a breaking change. Your change makes autoVerify opt-in while before it was always enabled so if people were to simply bump the plugin version their app links would break.

true, i thought of that when i defined it but i thought i'll just make the pr as is and ask if this could perhaps be in the next patch or minor version release for the plugin, change it accordingly 

so rn i can just set the default to true, or would you like something different?

FabianLars:
hmm yeah i think default true is fine

sosweetham:
> > A new app_link flag lets you explicitly opt-in to AppLink/UniversalLink behavior.
> 
> Not for universal links since ios doesn't support android's "inbetween" web links. You either have a custom scheme or use verified universal links. I'm not 100% sure if this is actually true, that's just based on what apple says in their docs, the actual behavior may be different 🤷

i am not sure, i just read the docs yesterday lol 

but rn within the scope of this pr i am just checking if any of associated domains have app_link true and then deciding whether to generate the entitlements or not, none of the other particulars are handled in this one, i was mainly concerned with custom url schemes

FabianLars:
yeah i get that. the problem with this is cross-platform compatiblity. What if a dev wants to use web links on android but universal links on ios? probably not realistic but still a possible situation.

sosweetham:
> yeah i get that. the problem with this is cross-platform compatiblity. What if a dev wants to use web links on android but universal links on ios? probably not realistic but still a possible situation.

i suppose they'd have to setup separate associated domains, although it really does seem like an edge case scenario

FabianLars:
> i suppose they'd have to setup separate associated domains, although it really does seem like an edge case scenario

that's not what i wanted to hint at. right now if a dev wants to do that and uses tauri.conf.json instead of separate tauri.ios.conf.json and tauri.android.conf.json they will end up with broken links on ios.
If iOS doesn't support "web links" then i think it'd be more reasonable to ignore the appLink config (making it android-only) and derive deep link vs universal link via the scheme only.

On a similar note i think we can also make the android implementation ignore the appLink config if the scheme is != http(s) to prevent misconfigurations but i guess that's not as important and raises the question of how much logic we should handle and how much we should just report back as errors.

sosweetham:
> If iOS doesn't support "web links" then i think it'd be more reasonable to ignore the appLink config (making it android-only) and derive deep link vs universal link via the scheme only.

Hmm i understand what you mean, but as a developer, this implicit behaviour that, I am building for ios so the framework assumes that I support universal links and have already setup the surrounding infra with all the association files and assetlinks and what not is quite a bit of a stretch no? But ofc this is an opinion not a statement lol

That's why I went with an opt in approach than out but the breaking change argument made sense so I backtracked it

lucasfernog:
> this implicit behaviour that, I am building for ios so the framework assumes that I support universal links and have already setup the surrounding infra with all the association files and assetlinks and what not is quite a bit of a stretch

well it would be if there was an alternative approach for iOS (doesn't seem like it, since we're saying it's either universal links or custom schemes)

sosweetham:
You are right, but it only makes sense if the developer performing the build already knows all the infra they need to setup

We could do it with docs 

But on the plugin level I think a warn on build citing what they need to if they don't do an explicit declaration "you have not explicitly declared --- build platform only supports ----, further reading; --some link to tauri/apple docs here maybe?-" would help out a ton 

This is still coming from someone with mostly a js/ts background so maybe system level engineers do the rtfm stuff first but I am sure this would save a lot of headache in this approach 🙂

FabianLars:
> Hmm i understand what you mean, but as a developer, this implicit behaviour that, I am building for ios so the framework assumes that I support universal links and have already setup the surrounding infra with all the association files and assetlinks and what not is quite a bit of a stretch no? But ofc this is an opinion not a statement lol

Just went through the changes a bit more and right now, setting appLink to false simply prevents the entitlements from being added. If this triggers no other logic (which it currently doesn't - edit: this also means we have to remember that once we generate the plist if we end up doing that) then appLink: false basically also means androidOnly: true (the configured link will be ignored on ios).
I guess that's a more likely wanted situation than what i said so i'll take that part back. (needs to be documented though)

(forgot to press send)

lucasfernog:
does a platform-specific toggle really makes sense? can't we just document that it should be added separetely in tauri.android.conf.json?

sosweetham:
> does a platform-specific toggle really makes sense? can't we just document that it should be added separetely in tauri.android.conf.json?

imo docs would be better all round, the linking stuff has a lot of components outside of tauri too so it would be better to aggregate resources/docs for the devs too than to leave implied knowledge via the plugin behavior mutating depending on just the configs unnecessarily

lucasfernog:
wait so a deep link can also have a host? that's crazy :joy: 

EDIT: if that's true we need to tweak the validation a bit

sosweetham:
it's more of a url spec thing from what i made of it, it's supposed to be called authority, but cuz http/ftp/sftp urls are common and in those the authority is generally a host... i suppose the term got coined as host, most url parsers dont call the thing authority either it's generally called host

but rn it's also possible to say, implement a ftp/sftp or whatever compatible mobile app with the plugin, so i suppose in a use case like that, calling it host makes more sense, so it can be a deeplink and also have hosts and stuff at the same time... really depends on the usecase 

i am not really sure i would change it though cuz someone could have deeplinks with a common "host" (authority) and build upon it with paths, identifiers and whatever, currently i dont think mentioning the "host" would cause any harm, provided the scheme is explicitly not, http or https

FabianLars:
i guess deep links are more or less free form so if you want it to have a host nothing's stopping you. But at least android explicitly allows a host config for deep links in the intent filter - no idea if it actually makes a difference though. But that's kinda why you were able to use deep links before this pr on android as well (seems like autoVerify is ignored for non http(s) schemes - or i was lied to) - "kinda" because you could set host to an empty string as well lol.

iOS doesn't care, it only wants to know the scheme (for deep links).

p.s. your cleanup removed android web link support, just in case.

sosweetham:
the technically correct thing might be to have either host or authority for correct semantics, but i dont really think it would make a difference

sosweetham:
> p.s. your cleanup removed android web link support, just in case.

ah i found what messed it up, the property in the config got removed for some reason? 
https://github.com/tauri-apps/plugins-workspace/pull/2870/commits/340ed4920e44b946e3181a0998f77c9f68991067

sosweetham:
oh ok it shifted to the is app link function, i'll add the property back to fix the regression

sosweetham:
Pacha Perfect

hmm does this mean that a lot of the manual Info.ios.plist stuff on the other plugins can finally be automated 👀

lucasfernog:
> Pacha Perfect
> hmm does this mean that a lot of the manual Info.ios.plist stuff on the other plugins can finally be automated 👀

yesssssss

lucasfernog (approved):

PR #2871, opened 2025-07-24
fix(dialog): remove use of ACTION_PICK

### LINKED ISSUE #2423: Tauri dialog box not opening in Android 15
I was trying to test this out on a phyisical Android 15 device.  Logs read by running `adb logcat`. I can test on other virtual devices if desired.  

Main error is here:
`No Activity found to handle Intent { act=android.intent.action.PICK typ=application/*...`

Which seems to indicate that the plugin needs to optionally use a different intent such as `new Intent(Intent.ACTION_OPEN_DOCUMENT);`

I'm not that familiar with the tauri plugin codebase otherwise I'd make the change myself.

Longer logs to give more context.
```
02-13 17:39:13.966  5761  5761 I GoogleInputMethodService: GoogleInputMethodService.onStartInput():1568 onStartInput(EditorInfo{EditorInfo{packageName=com.read_anywhere.app, inputType=0, inputTypeString=NULL, enableLearning=false, autoCorrection=false, autoComplete=false, imeOptions=12000000, privateImeOptions=null, actionName=UNSPECIFIED, actionLabel=null, initialSelStart=-1, initialSelEnd=-1, initialCapsMode=0, label=null, fieldId=0, fieldName=null, extras=null, hintText=null, hintLocales=[]}}, false)
02-13 17:39:13.966  5761  5761 I Module  : DeviceLockedStatusModuleProvider$Module.updateDeviceLockedStatus():100 repeatCheckTimes = 1, locked = false
02-13 17:39:13.967  1722  1759 W PackageConfigPersister: App-specific configuration not found for packageName: com.read_anywhere.app and userId: 0
02-13 17:39:13.967  5761  5761 I AndroidIME: InputBundleManager.loadActiveInputBundleId():407 loadActiveInputBundleId: und-Latn-x-password, password
02-13 17:39:13.969 15059 15059 V Tauri/Plugin: Tauri plugin: pluginId: dialog, command: showFilePicker
02-13 17:39:13.972 15059 15071 I ad_anywhere.app: Background concurrent mark compact GC freed 10019KB AllocSpace bytes, 26(2248KB) LOS objects, 84% free, 4567KB/28MB, paused 189us,1.650ms total 113.834ms
02-13 17:39:13.981 15059 15073 W ad_anywhere.app: ApkAssets: Deleting an ApkAssets object '<empty> and /data/app/~~02vgaGieu_dJGP8ldaa57A==/com.google.android.webview--2CRQaFaUX48UJZyESyd0w==/base.apk' with 2 weak references
02-13 17:39:14.024  1722  1759 I ActivityTaskManager: START u0 {act=android.intent.action.PICK typ=application/* (has extras)} with LAUNCH_MULTIPLE from uid 10050 result code=-91
02-13 17:39:14.024 15059 15059 E Tauri   : No Activity found to handle Intent { act=android.intent.action.PICK typ=application/* (has extras) }
```

### LINKED ISSUE #2826: [Bug] `open` from `Dialog` plugin has issues with filters on android

FabianLars:
fixes #2423
fixes #2826
closes #2866

from a bit of googling around it seems like there's no reason to use ACTION_PICK like ever. Replacing it with what we use when no filters are set was enough to fix the issue. From quite a bit of testing there also seems no reason for that `setIntentMimeTypes` function to exist. I did not see any difference in behavior so i'd prefer to remove it.

And yes, setting */* even if you set a single filter is totally fine.

### DIFF
--- a/plugins/dialog/android/src/main/java/DialogPlugin.kt
+++ b/plugins/dialog/android/src/main/java/DialogPlugin.kt
@@ -130,34 +130,6 @@ class DialogPlugin(private val activity: Activity): Plugin(activity) {
     return mimeTypes.toTypedArray()
   }
 
-  private fun setIntentMimeTypes(intent: Intent, mimeTypes: Array<String>) {
-    if (mimeTypes.isNotEmpty()) {
-      var uniqueMimeKind = true
-      var mimeKind: String? = null
-      for (mime in mimeTypes) {
-        val kind = mime.split("/")[0]
-        if (mimeKind == null) {
-          mimeKind = kind
-        } else if (mimeKind != kind) {
-          uniqueMimeKind = false
-        }
-      }
-
-      if (uniqueMimeKind) {
-        if (mimeTypes.size > 1) {
-          intent.putExtra(Intent.EXTRA_MIME_TYPES, mimeTypes)
-          intent.type = Intent.normalizeMimeType("$mimeKind/*")
-        } else {
-          intent.type = mimeTypes[0]
-        }
-      } else {
-        intent.type = "*/*"
-      }
-    } else {
-      intent.type = "*/*"
-    }
-  }
-
   @Command
   fun showMessageDialog(invoke: Invoke) {
     val args = invoke.parseArgs(MessageOptions::class.java)
@@ -211,10 +183,14 @@ class DialogPlugin(private val activity: Activity): Plugin(activity) {
       val parsedTypes = parseFiltersOption(args.filters)
 
       val intent = Intent(Intent.ACTION_CREATE_DOCUMENT)
-      setIntentMimeTypes(intent, parsedTypes)
-
       intent.addCategory(Intent.CATEGORY_OPENABLE)
       intent.putExtra(Intent.EXTRA_TITLE, args.fileName ?: "")
+      intent.type = "*/*"
+
+      if (parsedTypes.isNotEmpty()) {
+        intent.putExtra(Intent.EXTRA_MIME_TYPES, parsedTypes)
+      }
+
       startActivityForResult(invoke, intent, "saveFileDialogResult")
     } catch (ex: Exception) {
       val message = ex.message ?: "Failed to pick save file"
--- a/plugins/dialog/android/src/main/java/DialogPlugin.kt
+++ b/plugins/dialog/android/src/main/java/DialogPlugin.kt
@@ -56,20 +56,18 @@ class DialogPlugin(private val activity: Activity): Plugin(activity) {
     try {
       val args = invoke.parseArgs(FilePickerOptions::class.java)
       val parsedTypes = parseFiltersOption(args.filters)
-      
-      val intent = if (parsedTypes.isNotEmpty()) {
-        val intent = Intent(Intent.ACTION_PICK)
-        setIntentMimeTypes(intent, parsedTypes)
-        intent
-      } else {
-        val intent = Intent(Intent.ACTION_GET_CONTENT)
-        intent.addCategory(Intent.CATEGORY_OPENABLE)
-        intent.type = "*/*"
-        intent
+
+      // TODO: ACTION_OPEN_DOCUMENT ??
+      val intent = Intent(Intent.ACTION_GET_CONTENT)
+      intent.addCategory(Intent.CATEGORY_OPENABLE)
+      intent.type = "*/*"
+
+      if (parsedTypes.isNotEmpty()) {
+        intent.putExtra(Intent.EXTRA_MIME_TYPES, parsedTypes)
       }
 
       intent.putExtra(Intent.EXTRA_ALLOW_MULTIPLE, args.multiple ?: false)
-      
+
       startActivityForResult(invoke, intent, "filePickerResult")
     } catch (ex: Exception) {
       val message = ex.message ?: "Failed to pick file"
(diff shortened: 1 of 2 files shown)


FabianLars:
successfully tested by community https://github.com/tauri-apps/plugins-workspace/issues/2423#issuecomment-3117655548

plugins/dialog/android/src/main/java/DialogPlugin.kt
```diff
@@ -56,20 +56,18 @@ class DialogPlugin(private val activity: Activity): Plugin(activity) {
...
-        setIntentMimeTypes(intent, parsedTypes)
-        intent
-      } else {
-        val intent = Intent(Intent.ACTION_GET_CONTENT)
-        intent.addCategory(Intent.CATEGORY_OPENABLE)
-        intent.type = "*/*"
-        intent
+
+      // TODO: ACTION_OPEN_DOCUMENT ??
+      val intent = Intent(Intent.ACTION_GET_CONTENT)
```
Legend-Master, line 61: Most of the docs I read about were using [`ACTION_OPEN_DOCUMENT`](https://developer.android.com/reference/android/content/Intent#ACTION_OPEN_DOCUMENT), and https://developer.android.com/guide/components/intents-common#GetFile seems to suggest that `EXTRA_MIME_TYPES` is not available for [`ACTION_GET_CONTENT`](https://developer.android.com/reference/android/content/Intent#ACTION_GET_CONTENT), but from testing it does work

And seems like if I use `ACTION_OPEN_DOCUMENT`, I'll be limited to just the files, but with `ACTION_GET_CONTENT` I can pick files from other apps (so I would probably prefer `ACTION_GET_CONTENT` as the one you're currently using)

image

Legend-Master (approved):
Seems like Android got tons of different APIs for doing this, from [`ACTION_GET_CONTENT`](https://developer.android.com/reference/android/content/Intent#ACTION_GET_CONTENT)'s docs, it seems like we should almost always prefer it over `ACTION_PICK`

Note: didn't test

plugins/dialog/android/src/main/java/DialogPlugin.kt
FabianLars: iirc (already forget 4 days later) other similar frameworks use open_document instead as well. One thing that seems like a major advantage of open_document is that is opens them in-place without copying them into your app cache or whatever. so i guess offering it as an option (after more investigation) sounds valuable to me.

> And seems like if I use ACTION_OPEN_DOCUMENT, I'll be limited to just the files

not sure what you mean

> it seems like we should almost always prefer it over ACTION_PICK

yep, action_pick only makes sense if we already have a list of data/files which i can't really think of a usecase for here

Legend-Master: > not sure what you mean

This is with `ACTION_OPEN_DOCUMENT` (there's no side bar available and you can't choose files from Photos for example)

image

FabianLars: ah got it. i guess that makes sense though with how it opens files without copying them. hard to do if the file may be in the cloud.

Issue #2884, opened 2025-07-31
[bug] [plugin-nfc] NFC Write on iOS can't be used after an NFC prompt is cancelled


zaphim12:
### Describe the bug
On iOS, when an NFC write is initiated, there is a native Apple prompt that appears saying "Ready to Scan". Writing works fine when this prompt is open. However, if the prompt is closed or cancelled before a card is scanned, then future NFC writes will instantaneously fail with the error: "connected tag not found, please wait for it to be available and then call write()" - which comes from [this line in the NFC plugin](https://github.com/tauri-apps/plugins-workspace/blob/v2/plugins/nfc/ios/Sources/NfcPlugin.swift#L435).

This error triggers when a session is open but either session.nfcSession, session.tagStatus, or session.tag evaluates to 'nil'. As far as I can tell, the only way to remedy this currently is to reload the app or to complete a successful NFC read operation to close the session.

### Reproduction
I have tested this by creating a brand new tauri project and just implementing a button which triggers NFC writes. Here's the relevant code from my App.tsx
```
async function doNfcWrite() {
  try {
    const payload = [textRecord("12345")];

    const options = {
      kind: {
        type: "ndef" as const,
      },
      message: 'Scan a NFC tag',
      successfulReadMessage: 'NFC tag successfully scanned',
      successMessage: 'NFC tag successfully written',
    };
    await write(payload, options);
  } catch (e) {
    console.error(e);
  }
}
```

### Expected Behavior
The NFC write should be usable even if a prompt was cancelled one time

### Additional Context
I can think of two easy fixes for this. One is to modify the if/else mentioned above to close the session in the case that a session is open with no tag, nfcSession, or tagStatus available. The other is to enable the user to manually close a session in cases like these by exposing another plugin command which just closes the current session. I'm happy to implement the fix for this if that is acceptable.
EDIT --- 
Actually on further inspection I see a better fix. When the scan is cancelled, an error is already caught and printed [here on line 201](https://github.com/tauri-apps/plugins-workspace/blob/v2/plugins/nfc/ios/Sources/NfcPlugin.swift#L201). I believe that in this case it is best to add a call to 'closeSession()'. This will ensure future writes aren't blocked by a invalid session remaining.

FabianLars:
> EDIT ---
[...]

PR #2885, opened 2025-07-31
(fix: #2884) iOS NFC: Ensure that Session is closed when an error causes it to become invalid

zaphim12:
### Issue being fixed
Original issue: https://github.com/tauri-apps/plugins-workspace/issues/2884
Essentially, there are some cases where an NFC session becomes invalid, but the plugin doesn't properly close the session. This can occur if the user cancels the session before writing to a card, for example. In these cases, any subsequent writes fail immediately because a session is open, but it is invalid. 

### Changes made
In the case that a session is invalidated, instead of just logging the error and rejecting the invoke call, the session is also invalidated now.

### DIFF
--- a/plugins/nfc/ios/Sources/NfcPlugin.swift
+++ b/plugins/nfc/ios/Sources/NfcPlugin.swift
@@ -148,8 +148,8 @@ class NfcPlugin: Plugin, NFCTagReaderSessionDelegate, NFCNDEFReaderSessionDelega
 
   func tagReaderSession(_ session: NFCTagReaderSession, didInvalidateWithError error: Error) {
     Logger.error("Tag reader session error \(error)")
-    self.closeSession(session, error: "Tag reader session error: \(error)")
     self.session?.invoke.reject("session invalidated with error: \(error)")
+    self.session = nil
   }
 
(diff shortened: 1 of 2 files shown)


plugins/nfc/ios/Sources/NfcPlugin.swift
```diff
@@ -148,6 +148,7 @@ class NfcPlugin: Plugin, NFCTagReaderSessionDelegate, NFCNDEFReaderSessionDelega
 
   func tagReaderSession(_ session: NFCTagReaderSession, didInvalidateWithError error: Error) {
     Logger.error("Tag reader session error \(error)")
+    self.closeSession(session, error: "Tag reader session error: \(error)")
```
FabianLars: closeSession will remove self.session (of course it does, that's what we want) which will render the next line useless. Could you test if this also works?
```swift
self.session?.invoke.reject("session invalidated with error: \(error)")
self.session = nil
```
Since the error should only trigger when the session got invalidated already i don't think we need to call session.invalidate here but if it is required i think `self.closeSession(session)` would be better here (with the line order still swapped)

zaphim12: Makes sense. Setting the session to nil does work in these cases

FabianLars (approved):
thanks :)

PR #2896, opened 2025-08-04
feat: add tauri-plugin-dragout (macOS drag-out)

alexqqqqqq777:
Adds native macOS drag-out (file promise) support plugin.

Crate: https://crates.io/crates/tauri-plugin-dragout

### DIFF
--- a/plugins/dragout/Cargo.toml
+++ b/plugins/dragout/Cargo.toml
@@ -1,0 +1,43 @@
+[package]
+name = "tauri-plugin-dragout"
+version = "0.1.0"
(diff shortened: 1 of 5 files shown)


FabianLars:
Thanks for the PR. Unfortunately we currently cannot take on any more plugins in this repo. It also doesn't follow the repo standards as well.

I would recommend to keep hosting it yourself and add it to https://github.com/tauri-apps/awesome-tauri/ - we're also still thinking about an org for community plugins like how flathub works for example but we're a bit short staffed atm.

PR #2897, opened 2025-08-05
feature: reveal multiple items in dir

petersamokhin:
Hello and thanks a lot for your work!

This PR adds an option to reveal multiple files, instead of just one.
Given the contracts of the methods used under the hood on all platforms, the single path was always wrapped into an array anyway. 

Tested manually on macOS (and it works) — but I can also try Windows and Linux.

Please let me know where to add a change log entry, as currently it is a bit inconsistent between the plugins, and seems to be semi-automatic — I decided it's better to confirm it first.

Thanks!


plugins/opener/src/reveal_item_in_dir.rs
```diff
@@ -54,7 +92,7 @@ mod imp {
         },
     };
 
-    pub fn reveal_item_in_dir(path: &Path) -> crate::Result<()> {
+    pub fn reveal_item_in_dir<P: AsRef<Path>>(path: P) -> crate::Result<()> {
```
Legend-Master: This is used internally, please don't change it to `<P: AsRef<Path>>` here as that will increase generic bloat, same for a few other places

And for this specific case, I think maybe we could change it to take `paths: &[PathBuf]` directly since it's an internal function and they share most of the code

plugins/opener/guest-js/index.ts
```diff
@@ -95,3 +95,24 @@ export async function openPath(path: string, openWith?: string): Promise<void> {
...
+ * import { revealItemsInDir } from '@tauri-apps/plugin-opener';
+ * await revealItemsInDir(['/path/to/file']);
+ * ```
+ *
+ * @param paths The paths to reveal.
+ *
+ * @since 2.0.0
+ */
+export async function revealItemsInDir(paths: string[]) {
+  return invoke('plugin:opener|reveal_items_in_dir', { paths })
```
Legend-Master: Since it's JavaScript, I think we reuse the same function, and changing the param to `path: string | string[]`, and we can have a single permission to control both

Legend-Master (commented):
No a proper review yet, just some early feedbacks

plugins/opener/guest-js/index.ts
petersamokhin: @Legend-Master as I see permissions are auto-generated per commands declaration from Rust side, and my idea was to keep it backward-compatible without touching the existing commands. 
I could update the JS function only and call different commands there, but IMO it would only make code worse.

Do you think it makes sense to replace the original functions & commands to support multiple values only?
I wanted to see these changes in `2.5.0` and not `3.0.0` 😄 
Or what is the preferred way here?

plugins/opener/src/reveal_item_in_dir.rs
petersamokhin: @Legend-Master I pushed this change — but with `&[PathBuf]` it is now moved and `or_else` fallback requires cloning. Do you see a better way here?

plugins/opener/guest-js/index.ts
Legend-Master: > Do you think it makes sense to replace the original functions & commands to support multiple values only?

Yeah, that's how we usually do it, since the command's rust function is only used by the js function, it's fine to change, also we let the user to sync the js and rust package versions, it shouldn't cause any breakages

plugins/opener/src/reveal_item_in_dir.rs
Legend-Master: Ah, I see, maybe `&[&Path]` here?

plugins/opener/src/reveal_item_in_dir.rs
```diff
@@ -193,4 +325,25 @@ mod imp {
 
         Ok(())
     }
+
+    pub fn reveal_items_in_dir(paths: &[PathBuf]) -> crate::Result<()> {
```
Legend-Master, line 277: From the implementations, it seems to me that we could change the original function `reveal_item_in_dir` to `reveal_items_in_dir` directly so we don't need much duplicated code like we currently have, since these are all internal, it's fine to change

plugins/opener/src/reveal_item_in_dir.rs
```diff
@@ -35,10 +35,47 @@ pub fn reveal_item_in_dir<P: AsRef<Path>>(path: P) -> crate::Result<()> {
...
+/// ## Platform-specific:
+///
+/// - **Android / iOS:** Unsupported.
+pub fn reveal_items_in_dir<P: AsRef<Path>>(paths: &[P]) -> crate::Result<()> {
+    let mut path_bufs = vec![];
+
+    for path in paths.iter() {
+        let path = path.as_ref().canonicalize()?;
+        path_bufs.push(path);
+    }
```
Legend-Master: This way we could take all combination of `&[&Path]` `[PathBuf]`s

```suggestion
pub fn reveal_items_in_dir<I, P>(paths: I) -> crate::Result<()>
where
    I: IntoIterator<Item = P>,
    P: AsRef<Path>,
{
    let mut canonicalized = vec![];

    for path in paths {
        let path = path.as_ref().canonicalize()?;
        canonicalized.push(path);
    }
```

plugins/opener/src/lib.rs
```diff
@@ -148,6 +148,10 @@ impl<R: Runtime> Opener<R> {
     pub fn reveal_item_in_dir<P: AsRef<Path>>(&self, p: P) -> Result<()> {
         crate::reveal_item_in_dir::reveal_item_in_dir(p)
     }
+
+    pub fn reveal_items_in_dir<P: AsRef<Path>>(&self, p: &[P]) -> Result<()> {
+        crate::reveal_item_in_dir::reveal_items_in_dir(p)
+    }
```
Legend-Master:
```suggestion
    pub fn reveal_items_in_dir<I, P>(&self, paths: I) -> Result<()>
    where
        I: IntoIterator<Item = P>,
        P: AsRef<Path>,
    {
        crate::reveal_item_in_dir::reveal_items_in_dir(paths)
    }
```

plugins/opener/src/reveal_item_in_dir.rs
petersamokhin: That was my first attempt but no, this kind of slice doesn't seem to be applicable: `&[&Path]`, as I don't see a way to create it apart from `vec.push(&path)` but the lifetime is not enough 😅

Legend-Master: Yeah, I think `&[PathBuf]` should be fine here, at least it seems like that from the current implementation?

Legend-Master:
Also just a question, should we support supplying paths with different parent directories here? That would open multiple windows I think but could be fine

petersamokhin:
@Legend-Master my testing shows:
- macOS: works out of the box
- Linux (Ubuntu 22.04 desktop): works out of the box
- Windows: doesn't work (yet?). 
  - The method that is being used forces us to provide the parent directory, so we handle this on the code level and filter out the files that are not placed together with the 1st file in the list. 
  - Do you think it makes sense to complicate the logic and somehow handle it manually? TBH I don't see a smooth way to do this. I didn't test it on Windows, but on macOS if you call that method concurrently, it crashes without an option to catch the error. I assume on Windows it won't be reliable neither, if we just wrap it into a loop the same way. If the method is sync and we await its completion, it might work, but again doesn't look reliable + might take time on some devices. So, if there's no native method that supports this behaviour, I'd rather keep it as is, so it won't work only for Windows.

Legend-Master:
Makes sense, we could document this behavior though

plugins/opener/guest-js/index.ts
petersamokhin: @Legend-Master TL;DR I updated the code to leave the single method under the hood (with `&[path]` approach) but kept all the public exports, `Opener` methods and JS API backward-compatible. Please let me know if you still have concerns, happy to update 🫡 

My 2 cents "why":
People might use `invoke` directly, and this logic would be broken. I assume it's not a recommended approach (it helped me testing the forked plugin without publishing the npm package though).

Also, opener [README](https://github.com/tauri-apps/plugins-workspace/blob/v2/plugins/opener/README.md) suggests using the below approach:

```rust
use tauri_plugin_opener::OpenerExt;

fn main() {
    tauri::Builder::default()
        .plugin(tauri_plugin_opener::init())
        .setup(|app| {
            let opener = app.opener();

            // Reveal a path with the system's default explorer
            opener.reveal_item_in_dir("/path/to/file")?;
            Ok(())
        })
}
```

So, it'd be a breaking change to just change the commands and exports too.

plugins/opener/src/reveal_item_in_dir.rs
petersamokhin: I updated it, as per your recommendations below with `IntoIterable` and without any cloning, should be pretty neat now

plugins/opener/src/reveal_item_in_dir.rs
petersamokhin: Updated, now there's just one fn with this logic under the hood

plugins/opener/src/reveal_item_in_dir.rs
petersamokhin: Applied 🫡

plugins/opener/src/lib.rs
petersamokhin: Applied 🫡

petersamokhin:
@Legend-Master I added notes about this Windows-specific behaviour to the fn docs and to README

plugins/opener/guest-js/index.ts
Legend-Master: So `app.opener().reveal_item_in_dir("path")` points to https://github.com/tauri-apps/plugins-workspace/blob/61ffc83a27f67ca178645e05f76ee3c2ff6409ed/plugins/opener/src/lib.rs#L148 it doesn't point to the js command (the one with `#[tauri::command]`)

> People might use invoke directly, and this logic would be broken. I assume it's not a recommended approach (it helped me testing the forked plugin without publishing the npm package though).

I mean, we can't guarantee we don't break this type of usage, it's considered an implementation detail, no?

Legend-Master:
> @Legend-Master I added notes about this Windows-specific behaviour to the fn docs and to README

~~Would be nice to also include it in the rust functions 🙃~~ never mind, I see cbd918e

plugins/opener/src/reveal_item_in_dir.rs
```diff
@@ -115,24 +180,36 @@ mod imp {
...
+    use std::path::PathBuf;
 
-    pub fn reveal_item_in_dir(path: &Path) -> crate::Result<()> {
+    pub fn reveal_items_in_dir(paths: &[PathBuf]) -> crate::Result<()> {
         let connection = zbus::blocking::Connection::session()?;
 
-        reveal_with_filemanager1(path, &connection)
-            .or_else(|_| reveal_with_open_uri_portal(path, &connection))
+        reveal_with_filemanager1(paths, &connection).or_else(|e| {
+            // Fallback to opening the directory of the first item if revealing multiple items fails.
```
Legend-Master, line 207: I'm not sure if this fallback makes sense, is opening multiple prone to failures?

plugins/opener/src/reveal_item_in_dir.rs
```diff
@@ -54,56 +96,79 @@ mod imp {
...
                     };
-
-                    ShellExecuteExW(&mut info).inspect_err(|_| {
-                        ILFree(Some(dir_item));
-                        ILFree(Some(file_item));
-                    })?;
+                    ShellExecuteExW(&mut info).map(|_| ()).map_err(Into::into)
+                } else {
+                    Err(e.into())
                 }
```
Legend-Master, line 155: Any reason to change the original fallback function here? (the comments and `is_dir` handling)

plugins/opener/src/reveal_item_in_dir.rs
```diff
@@ -54,56 +96,79 @@ mod imp {
...
 
-        let dir = HSTRING::from(dir);
-        let dir_item = unsafe { ILCreateFromPathW(&dir) };
+        // Ensure dir_item is freed even if subsequent operations fail.
+        let mut created_file_items = Vec::new();
 
-        let file_h = HSTRING::from(file);
-        let file_item = unsafe { ILCreateFromPathW(&file_h) };
+        for path in &files_in_same_dir {
+            let file_hstring = HSTRING::from(path.as_os_str());
```
Legend-Master: I think we need the `file_hstring` to be alive even after the loop here, else the memory is freed and `file_item` will be pointing to that freed memory (I'm not entirely sure if it's the case that `ILCreateFromPathW` stores the input string, but better be safe here)

plugins/opener/src/reveal_item_in_dir.rs
```diff
@@ -54,56 +96,79 @@ mod imp {
...
+            let file_item = unsafe { ILCreateFromPathW(&file_hstring) };
+            if !file_item.is_null() {
+                created_file_items.push(file_item);
```
Legend-Master: I think we could implement something like https://github.com/microsoft/windows-rs/blob/05879ec957df7ad92d0653fb84a83f75a0a9b4e9/crates/libs/core/src/handles.rs#L18 to make it lesser error prone to free

plugins/opener/src/reveal_item_in_dir.rs
```diff
@@ -54,56 +96,79 @@ mod imp {
...
     };
 
-    pub fn reveal_item_in_dir(path: &Path) -> crate::Result<()> {
-        let file = dunce::simplified(path);
+    pub fn reveal_items_in_dir(paths: &[PathBuf]) -> crate::Result<()> {
+        if paths.is_empty() {
+            return Ok(());
+        }
+
+        let first_path = dunce::simplified(&paths[0]);
```
Legend-Master: I think we could remove all the `dunce::simplified` calls and use `dunce::canonicalize` instead of `PathBuf` on the ones above (https://github.com/tauri-apps/plugins-workspace/pull/2897#discussion_r2255656789)

plugins/opener/src/reveal_item_in_dir.rs
```diff
@@ -21,7 +21,48 @@ pub fn reveal_item_in_dir<P: AsRef<Path>>(path: P) -> crate::Result<()> {
...
+/// - **Windows:** Only supports revealing items in the same directory.
+pub fn reveal_items_in_dir<I, P>(paths: I) -> crate::Result<()>
+where
+    I: IntoIterator<Item = P>,
+    P: AsRef<Path>,
+{
+    let mut canonicalized = vec![];
+
+    for path in paths {
+        let path = path.as_ref().canonicalize()?;
```
Legend-Master:
```suggestion
        let path = dunce::canonicalize(path.as_ref())?;
```

Legend-Master:
Since I can't push to your branch, I opened https://github.com/getcompress/plugins-workspace/pull/1 against this branch to support multiple roots on Windows

plugins/opener/guest-js/index.ts
petersamokhin: In Rust, `reveal_item_in_dir` seems to be exported via `pub use` in the `lib.rs` file, so is usable directly without `.opener()` as well (please correct me if I'm wrong), so I kept both fns untouched (`pub use reveal_item_in_dir::reveal_item_in_dir` and `Opener::reveal_item_in_dir`. Let me know if this should be simplified as well (remove `reveal_item_in_dir` and keep only the new `reveal_items_in_dir`).

Regarding the commands, I don't think it's a blocker and is likely fine to change them, so let's simplify them as well, as you say 🫡

plugins/opener/src/reveal_item_in_dir.rs
petersamokhin: Fallback to `reveal_with_open_uri_portal` was already used here

plugins/opener/src/reveal_item_in_dir.rs
petersamokhin: I've merged your PR with this change applied

plugins/opener/src/reveal_item_in_dir.rs
petersamokhin: I've merged your PR with this change applied

plugins/opener/src/reveal_item_in_dir.rs
petersamokhin: I guess the implementation from your PR already covers this, right? Merged it

plugins/opener/src/reveal_item_in_dir.rs
petersamokhin: No, let's consider it an oversight

petersamokhin:
@Legend-Master I've approved your PR and seems like it fixes the rest of your comments, right?
Only question is that you removed some explanatory comments from code, shall we revert them?

UPD: actually the build was failing outside `windows` config, so I fixed it by adding `dunce` to all configs on the `opener` crate level

UPD2: tested your changes on Windows, it works 🚀

plugins/opener/src/reveal_item_in_dir.rs
petersamokhin: it looks like it's fixed by your PR as well, right? Merged it

plugins/opener/guest-js/index.ts
Legend-Master: > In Rust, `reveal_item_in_dir` seems to be exported via `pub use` in the `lib.rs` file, so is usable directly without `.opener()` as well (please correct me if I'm wrong), so I kept both fns untouched (`pub use reveal_item_in_dir::reveal_item_in_dir` and `Opener::reveal_item_in_dir`. Let me know if this should be simplified as well (remove `reveal_item_in_dir` and keep only the new `reveal_items_in_dir`).

Yeah they're public so we can't change these, I was only talking about changing the command here

> Regarding the commands, I don't think it's a blocker and is likely fine to change them, so let's simplify them as well, as you say 🫡

Let's do it then 🙃

Legend-Master:
> I've approved your PR and seems like it fixes the rest of your comments, right?

Yep

> Only question is that you removed some explanatory comments from code, shall we revert them?

The limitation is now gone, so I removed those comments

> UPD: actually the build was failing outside windows config, so I fixed it by adding dunce to all configs on the opener crate level
> 
> UPD2: tested your changes on Windows, it works 🚀

👍

plugins/opener/src/reveal_item_in_dir.rs
Legend-Master: Ah I see

plugins/opener/guest-js/index.ts
petersamokhin: @Legend-Master I did the change and then realised that replacing the command will be a breaking change for `capabilities/*.json`. Permission names are auto-generated based on the command name, thus if we rename the command, the permission will be changed from `opener:allow-reveal-item-in-dir` to `opener:allow-reveal-items-in-dir`. 

It can be avoided by using _only_  the `opener:default` — but I personally try to limit the permissions to only those I explicitly use.

**Our current options are:**
1. Do you think above is not a concern, we can ignore it and release a minor version like that?
    - It will affect only those users who explicitly used `opener:allow-reveal-item-in-dir`.
2. Or shall we better keep them as separate commands?
    - I agree it's not ideal as these functions may confuse people, when changing permissions as well.
    - Still looks like a compromise to me ¯\\\_(ツ)_/¯
    - This is the current option. If you think this is the way to go, we can proceed with merging the PR and publishing the minor version.
3. We can keep the Rust command name as `reveal_item_in_dir` but change the parameter type.
    - We then keep the contracts the same in JS, and command name in Rust, but the naming will be inconsistent and confusing (on the level of permissions as well, as people might rely on the "single" wording, vs actual "plural" behaviour).
5. Otherwise, can we consider a major version bump then?
    - Might feel weird as I don't see any other crates related to Tauri that have the `3.+` version yet.

Legend-Master: > We can keep the Rust command name as reveal_item_in_dir but change the parameter type.
We then keep the contracts the same in JS, and command name in Rust, but the naming will be inconsistent and confusing (on the level of permissions as well, as people might rely on the "single" wording, vs actual "plural" behaviour).

Let's go for option 3, I think we just keep it as `reveal_item_in_dir` and `revealItemInDir` for now, and leave a todo comment for changing it to `reveal_items_in_dir` in v3, the `item` and `items` difference doesn't matter too much in my opinion

petersamokhin: @Legend-Master I've pushed the commit:
- unify the JS public function via `string | string[]`
- keep only the "plural" implementation on the Rust side, except public functions
- added a TODO for the command only (implies renaming other things too)

Legend-Master (dismissed):
Goods good, we just need a change file and format the `plugins/opener/guest-js/index.ts`, and we are good to go

petersamokhin:
@Legend-Master done 🫡

Legend-Master (approved):
Thanks!

PR #2900, opened 2025-08-07
feat: add secure-storage / keychain plugin

FabianLars:
Yet another keychain tauri plugin? 😮‍💨 Yep yep, mainly because it serves as the replacement for stronghold and i thought it may be reasonable to hold our hands over this (more so than the non secure store plugin imo) and the other plugins all seem kindaaa dead (or maybe not since they just as slim as this one)

I probably won't merge this until keyring v4 is released or getting closer to a release at least.

About the implementation there are only 2 things i think are worth to discuss:
1) The plugin name. keychain and keyring or whatever were already taken and i thought that maybe people don't know what that mean so i went with a mix of expo's secureStore and electron's safeStorage but _very_ open for suggestions.
2) api naming: keyring uses secret and password and i used binary and string. i don't like either tbh but i didn't look into all of keyring's backends to see why they chose that naming scheme.
3) linux backend: for this poc i went with keyutils without persistence (== reset on reboot). the last (yanked) keyring v4 rc version used sync-secret-service as the default but i don't like that it links to c libs (dbus). Since this is linux i couldn't be bothered to look into it that much at this point.

Also i think in the future it would be cool to have something built-in that encrypts the secrets before giving them to keyring so that on Windows they aren't stored as plain text for everyone to read. Of course that wouldn't really be secure but better than nothing.


FabianLars:
oh man that edition shit is annoying in a workspace for ci

Ludea:
awesome idea to replace stronghold ! 

Some answer:
1 - SecureStorage is a good name. Simple and easy to understand what is the purpuoe of the plugin
2 - binary name can be confusing. WDYT aabout string and data ?
3 - if persistence is enable for Macos and Windows, it seems logical to enable it for linux too. If you don't like c api, maybe there is a pure rust lib to do the same thing ?

xty:
> 1. The plugin name. keychain and keyring or whatever were already taken and i thought that maybe people don't know what that mean so i went with a mix of expo's secureStore and electron's safeStorage but _very_ open for suggestions.

Calling it Secure Storage is about as good as secureStore, but we can do better – security has many facets (e.g. confidentiality, integrity, availability), so Secure Storage can be interpreted in different ways while this plugin focuses on confidentiality. It doesn't, for example, guard against data corruption. Therefore, I would propose going with **Secret Storage** or **Secret Store** instead in order to be more precise.

> 2. api naming: keyring uses secret and password and i used binary and string. i don't like either tbh but i didn't look into all of keyring's backends to see why they chose that naming scheme.

I know where you're coming from. "**bytes**" over "binary", perhaps? This should feel straightforward to Rust and JavaScript programmers alike.

> 3. linux backend: for this poc i went with keyutils without persistence (== reset on reboot). the last (yanked) keyring v4 rc version used sync-secret-service as the default but i don't like that it links to c libs (dbus). Since this is linux i couldn't be bothered to look into it that much at this point.

I agree with the previous commenter in stressing the need for consistency, but as a POC I'll let you get away without persistence. 😉 Despite the many challenges, please don't make GNU/Linux a second-tier platform.

FabianLars:
> Therefore, I would propose going with Secret Storage or Secret Store instead in order to be more precise.

That makes so much sense!! I'll still include the other name(s) as an alias somewhere for SEO reasons though.

> I know where you're coming from. "bytes" over "binary", perhaps? This should feel straightforward to Rust and JavaScript programmers alike.

In hindsight so obvious...

> 3 - if persistence is enable for Macos and Windows, it seems logical to enable it for linux too. If you don't like c api, maybe there is a pure rust lib to do the same thing ?

> I agree with the previous commenter in stressing the need for consistency, but as a POC I'll let you get away without persistence. 😉 Despite the many challenges, please don't make GNU/Linux a second-tier platform.

That goes against my nature of a professional Linux hater though :(
Anyway, seems like the keyring crates are now ready-ish so i'll take a proper look at the backends to see which one makes the most sense for us.

----

Thank you both for the valuable feedback!

FabianLars:
what about just `tauri-plugin-secrets` ? I can't decide between Store and Storage so i thought maybe neither? 😂
If we indeed end up adding something like electron's safeStorage apis (which are about encrypting/decrypting and not about storing data directly) that may also make more sense.

Why is naming stuff so hard 😮‍💨

xty:
> what about just `tauri-plugin-secrets` ? I can't decide between Store and Storage so i thought maybe neither? 😂

If you're not entirely happy with either, despite the alliteration, consider a third option: *Secret Vault*. 😂
1. "secret" goes well with "vault" – I even checked the *Oxford Collocations Dictionary*.
2. The term "vault" is used by password managers like Bitwarden and 1Password, as well as by [HashiCorp](https://github.com/hashicorp/vault). It should definitely evoke the right imagery.

> If we indeed end up adding something like electron's safeStorage apis (which are about encrypting/decrypting and not about storing data directly) that may also make more sense.

`tauri-plugin-secrets` works too, but only if you want de-emphasize the storage aspect. If I'm not mistaken, the user would be mostly interested in storing their sensitive data *somewhere* – whether that capability is provided by the plugin per se, from their perspective, is only an implementation detail.

FabianLars:
> tauri-plugin-secrets works too, but only if you want de-emphasize the storage aspect.

I kindaaa wanted to do the opposite. secret-store etc to me (as a non native speaker) sounds more like the store itself is secret and not that it's a store for secrets. maybe `secrets-store` with an extra s improves that better than dropping the second word idk

xty:
> I kindaaa wanted to do the opposite. secret-store etc to me (as a non native speaker) sounds more like the store itself is secret and not that it's a store for secrets. maybe `secrets-store` with an extra s improves that better than dropping the second word idk

Why not ask a large language model? They may not excel at programming but are best at answering language questions. 😂

Screenshot_2025-10-15-02-27-17-43

Honestly, **Secret Vault**, **Secret Store**, **Secrets**, **Secrets Store** are all probably fine. If you really can't decide, create a poll in GitHub Discussions.

brotskydotcom:
Hi folks, @brotskydotcom here, one of the maintainers of the [Rust Keyring ecosystem](https://github.com/open-source-cooperative/keyring-rs/wiki/Keyring). Just wanted to give you a heads-up that we expect a stable v4 release of the ecosystem sometime in January, and that all of the constituents (keyring-core and the various storage modules) are stable and ready for use now (although Android is getting features this month you might want to wait for).

WRT Linux, I would strongly suggest you use the dbus-secret-service-keyring-store module rather than the linux-keyutils-keyring-store module, because that way you get persistence, user control of storage back end (e.g., gnome vs. KDE vs. KeePass), and consistency with non-Linux *nixen. WRT the "non-pure-rust" complaint, leaving aside the fact that Tauri relies on a bunch of non-Rust code on every platform, you can use the "vendored" feature of the dbus-secret-service-keyring-store module to statically bind the C library into the module so no one needs lib-dbus at runtime.

You can track the status of the overall ecosystem at open-source-cooperative/keyring-rs#259.

FabianLars:
thanks for the headsup and your input on the linux topic. i was leaning towards the zbus variant instead, any opinion on that? I think all of our stack at some point switched from dbus to zbus which for us iirc meant way easier cross-compilation.
At the plugin level here the async runtime requirement isn't an issue cause we have tokio around anyway.

brotskydotcom:
The zbus variant is equally solid and since the Tauri infrastructure already uses tokio there's no downside. But be careful: because keyring uses the blocking APIs of the zbus secret service, and because those always spin up a separate runtime loop to wait for the async calls, tokio apps that call the keyring must always make those calls on a separate thread (not the main thread), because the second run loop will block the calling thread. So your plugin should be sure to have its own (reserved) thread on which it makes all its calls, rather than making the calls on the thread that calls the plugin.

FabianLars:
> The zbus variant is equally solid and since the Tauri infrastructure already uses tokio there's no downside.

okay, nice

> But be careful: ...

That shouldn't be much of a problem i hope. Thanks for the warning!

IT-ess:
Now that keyring-rs has hit the beta (thanks @brotskydotcom !), what are the remaining tasks for this plugin to be merged ? I'm willing to test it if needed, but I'm not sure it is already working right now ?

FabianLars:
1) naming (plugin itself and api)
2) figure out what to do with https://github.com/tauri-apps/plugins-workspace/pull/3071#issuecomment-3452245315
3) wait for stable releases - i'm currently using a fork of the android backend, i assume that's not needed anymore with the new betas
4) figure out a nice-ish way to have multiple MSRVs in the same repo without breaking CI
5) the todos in the code, eg docs, deletion apis, apple feature flags, ...
6) see comments above for the thread requirements, should also have a todo in the code somewhere

also since i opened this pr there were like 3 new keychain community plugins so maybe we should check them out again and see if it makes sense to reference them and/or ask them if they want to be upstreamed into our repo.

----

i have free time to do exactly none of the tasks so that'll be fun.

IT-ess:
Sounds like a lot of work :sweat_smile: 
I had forked a plugin (because it wasn't working on linux), and it did work pretty well, at least on Linux, iOS and Android (see the [repo](https://github.com/IT-ess/tauri-plugin-keyring)). I don't use this project anymore since I just moved these methods directly into my tauri app code. It didn't use `keyring-rs` but rather `keyring-core` and the related stores, but I guess it could help.

Ping me if you need testers :smile:

brotskydotcom:
@IT-ess is exactly right: use keyring-core and the android-native-keyring-store (no other crates). The keyring-rs crate is now just a demo crate with lots of sample code and examples of how to use keyring (including a Tauri app). All this plugin really needs to do is expose commands for the keyring-core entries. See [the Keyring Demo core](https://github.com/open-source-cooperative/keyring-rs/tree/next/tauri/src-tauri) for an example.

suasgn:
`tauri-plugin-safe`

IT-ess:
Hi @FabianLars, just saw you pushed code on this branch. 
When I tried to upgrade to tauri 2.11 this morning I figured out that the Android crate (`android-native-keyring-store`) used for keyring no longer worked because of this error `android context was not initialized`. 
I digged a bit, and I found out that it is because tao 0.35 no longer initializes the ndk-context, required by this crate. 
I just created this [issue](https://github.com/open-source-cooperative/android-native-keyring-store/issues/21) on the android-native-keyring-store repo to figure out a fix, but it may be easier if Tauri/TAO keeps initializing the ndk-context anyway ?

Looking forward to your response !

Issue #2720, opened 2025-05-26
Use SMAppService to register the app to LoginItems instead of an AppleScript


iparaskev:
At [Hopp](https://gethopp.app/), we want our app to start automatically on user login. Initially, we used the `autostart` plugin, which does the job but we thought that user experience can be improved.

In particular we found the two notifications the user receives slightly annoying.

![Image](https://github.com/user-attachments/assets/74d18898-60a4-4710-8e79-193cf8ff80ed)

![Image](https://github.com/user-attachments/assets/8d998d4b-4578-434a-92c4-3875a1bd320d)

We didn't like them because
* They are two instead of just one when using a native API.
* It isn’t entirely clear from the pop-ups which app is being added to the login items.

To address these, I decided to make [smappservice-rs](https://crates.io/crates/smappservice-rs), which is just a rust wrapper for macOS's `ServiceManagement` framework. When it is used instead of `auto-launch` it gives only one notification.

Would you consider  accepting an enhancement to the `autostart` plugin that introduces a third option leveraging `smappservice-rs`?

I can also try extending `auto-launch` If you think that this change is more appropriate there. The reason I am first trying here is that `auto-launch` can be used to launch any app on start up, while `smappservice-rs` is for the running app only. 

**Note:** I tested whether issue #2342 could be resolved using `smappservice-rs`, and it appears to work. Perhaps someone else could verify this as well.

FabianLars:
I wouldn't mind adding it to the plugin but isn't this something that should be upstreamed into auto-launch ?

iparaskev:
The problem I see with `auto-launch` is that it requires the `app-name` and `app-path` to [not be None](https://docs.rs/auto-launch/0.5.0/auto_launch/struct.AutoLaunchBuilder.html#method.build). When an app is registered via the [SMAppService](https://developer.apple.com/documentation/servicemanagement/smappservice?language=objc), setting the app-name and path doesn't make a lot of sense because the app is registering itself, for example see registering the [mainAppService](https://developer.apple.com/documentation/servicemanagement/smappservice/mainapp?language=objc). I guess we could modify `auto-launch` to ignore these parameters when `SMAppService` is used. 

I will create an MR for `auto-launch` (it won't take long) and let the maintainer decide. 

I will come back here to report any updates.

chad-fossa:
autolaunch seems to be sitting on this, would be good to get a patch at the very least

naoufalelh:
[...]

PR #2909, opened 2025-08-11
feat(autostart): remove `&'static` constraint for `init(args)` parameter

WSH032:
The `&'static str` makes `fn init` almost useless for [FFI scenarios](https://github.com/pytauri/pytauri/blob/87e0cd4d45e8ce5021428a8a7da16abc43e4b372/crates/pytauri-core/src/plugins/autostart/mod.rs#L74-L81). Therefore, I replaced it with `IntoIterator<Item = impl Into<String>>`, but this [could be a breaking change](https://doc.rust-lang.org/cargo/reference/semver.html#fn-generic-new).

What does the Tauri team think about this? If semantic compatibility prevents this improvement in v2, I am ok with temporarily implementing a new `init` downstream using a `Builder`.

### DIFF
--- a/plugins/autostart/src/lib.rs
+++ b/plugins/autostart/src/lib.rs
@@ -234,10 +234,14 @@ impl Builder {
 /// Initializes the plugin.
 ///
 /// `args` - are passed to your app on startup.
-pub fn init<R: Runtime>(
+pub fn init<R: Runtime, I, S>(
     #[allow(unused)] macos_launcher: MacosLauncher,
-    args: Option<Vec<&'static str>>,
-) -> TauriPlugin<R> {
+    args: Option<I>,
+) -> TauriPlugin<R>
+where
+    I: IntoIterator<Item = S>,
+    S: Into<String>,
+{
     let mut builder = Builder::new();
     if let Some(args) = args {
         builder = builder.args(args)
--- a/plugins/autostart/src/lib.rs
+++ b/plugins/autostart/src/lib.rs
@@ -234,6 +234,7 @@ impl Builder {
 /// Initializes the plugin.
 ///
 /// `args` - are passed to your app on startup.
+// TODO: Remove `init` function or `args` argument in v3
 pub fn init<R: Runtime, I, S>(
     #[allow(unused)] macos_launcher: MacosLauncher,
     args: Option<I>,
--- a/.changes/autostart-remove-static-constraint.md
+++ b/.changes/autostart-remove-static-constraint.md
@@ -1,0 +1,6 @@
+---
+autostart: minor:enhance
+autostart-js: minor:enhance
+---
+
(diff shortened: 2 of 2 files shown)


FabianLars:
i think this change is fine. that said, i'd remove the args argument in v3, doesn't make sense to me here and the Builder should be used if args need to be set.

plugins/autostart/src/lib.rs
```diff
@@ -234,10 +234,14 @@ impl Builder {
 /// Initializes the plugin.
 ///
 /// `args` - are passed to your app on startup.
-pub fn init<R: Runtime>(
+pub fn init<R: Runtime, I, S>(
     #[allow(unused)] macos_launcher: MacosLauncher,
-    args: Option<Vec<&'static str>>,
-) -> TauriPlugin<R> {
+    args: Option<I>,
```
FabianLars, line 240:
```suggestion
    // TODO: Remove this argument in v3
    args: Option<I>,
```

WSH032: I have no objection to this. Personally, I think `macos_launcher` could also be removed, since the usage in the documentation is `macos_launcher: MacosLauncher::default()`, which means most beginners probably won't set these two parameters. What do you think?

*But on the other hand, if we remove all the parameters, why not just use `Builder::new().build()`? 😂*

FabianLars: the difference here is that macos_launcher must be set. well, with the current api design/intention at least. maybe it becomes more relevant with https://github.com/tauri-apps/plugins-workspace/issues/2720 but i wouldn't mind removing it from init either.

> But on the other hand, if we remove all the parameters, why not just use Builder::new().build()? 😂

That would be more in line with most plugins. They typically have either a builder or init, not both (with exceptions). In this case the builder was added at a later point.

plugins/autostart/src/lib.rs
```diff
@@ -234,10 +234,14 @@ impl Builder {
 /// Initializes the plugin.
 ///
 /// `args` - are passed to your app on startup.
-pub fn init<R: Runtime>(
+pub fn init<R: Runtime, I, S>(
```
FabianLars, line 238:
```suggestion
// TODO: Remove `init` function or `args` argument in v3
pub fn init<R: Runtime, I, S>(
```
new suggestion ^^

Legend-Master:
I would probably prefer keeping this function as is so we don't breaking anything and remove it in v3, the builder was added for the exact reason because this function is not very ergonomic (forces you to pass in `args` and `macos_launcher`), I don't even mind we add `#[deprecated]` to it

If we don't remove it, we probably want to remove all the parameters and making it just an alias to `Builder::new().build()` (which is a little shorter and if most people don't need to config the builder, it could be used)

WSH032:
I agree with @Legend-Master's opinion (There should be one-- and preferably only one --obvious way to do it. --Zen of Python).

Having both `fn init()`/`fn init(macos_launcher)` and `Builder::new().build()` can be confusing, and seeing `&'static str` in the signature (as a library) is even more confusing 😂.

So, let's close this PR?

PR #2915, opened 2025-08-13
fix: Specify Generic for readFile ArrayBuffer (fix #2914)

### LINKED ISSUE #2914: fs' readFile incompatible with TypeScript 5.9.2
TypeScript 5.9.2 changed the default behavior of `BufferSource` in `lib.dom.d.ts`. Now, instead of accepting `type BufferSource = ArrayBufferView | ArrayBuffer;`, it accepts type `BufferSource = ArrayBufferView<ArrayBuffer> | ArrayBuffer;`.  This manifests when attempting to use the result of a `readFile` call to construct a `File`:

Example:
```
const filePath = "./myfile.txt"
const fileData = await readFile(filePath));
const file = new File([fileData], "myfile.txt");
```

Error:
```
Type 'Uint8Array<ArrayBufferLike>' is not assignable to type 'BlobPart'.
  Type 'Uint8Array<ArrayBufferLike>' is not assignable to type 'ArrayBufferView<ArrayBuffer>'.
    Types of property 'buffer' are incompatible.
      Type 'ArrayBufferLike' is not assignable to type 'ArrayBuffer'.
        Type 'SharedArrayBuffer' is not assignable to type 'ArrayBuffer'.
          Types of property '[Symbol.toStringTag]' are incompatible.
            Type '"SharedArrayBuffer"' is not assignable to type '"ArrayBuffer"'.
```

samhinshaw:
Resolves #2914.

Here we specify the generic argument to `Uint8Array`, `Uint8Array<ArrayBuffer>`.

If unspecified, the default is `Uint8Array<ArrayBufferLike>`. However, both `new Uint8Array()` and `Uint8Array.from()` return `Uint8Array<ArrayBuffer>`.

### DIFF
--- a/.changes/fs-write-uint8array-arraybuffer.md
+++ b/.changes/fs-write-uint8array-arraybuffer.md
@@ -1,0 +1,6 @@
+---
+fs: patch
+fs-js: patch
+---
+
+`readFile` now returns a more specific type `Promise<Uint8Array<ArrayBuffer>>` instead of the default `Promise<Uint8Array<ArrayBufferLike>`
--- a/plugins/fs/guest-js/index.ts
+++ b/plugins/fs/guest-js/index.ts
@@ -739,7 +739,7 @@ interface ReadFileOptions {
 async function readFile(
   path: string | URL,
   options?: ReadFileOptions
-): Promise<Uint8Array> {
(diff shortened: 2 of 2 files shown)


Legend-Master (approved):
Thanks

PR #2917, opened 2025-08-18
refactor: add type annotation on `as_ref` for upcoming tauri 2.8

Legend-Master:
In https://github.com/tauri-apps/tauri/pull/14012/ we added `as_ref` to support casting to a `Window`, and these plugins will no longer compile due to that

### DIFF
--- a/.changes/positioner-tauri-2.8.md
+++ b/.changes/positioner-tauri-2.8.md
@@ -1,0 +1,6 @@
+---
+positioner: patch
+positioner-js: patch
+---
+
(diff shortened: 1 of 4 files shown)


Legend-Master:
I think we should release this before we release tauri 2.8 so people will still get a good build when they run `cargo update` (or alternatively revert https://github.com/tauri-apps/tauri/pull/14012 and delay it to v3)

Legend-Master:
Went for the revert: https://github.com/tauri-apps/tauri/pull/14026

PR #2928, opened 2025-08-19
fix(deep-link): Add command name to linux errors

### LINKED ISSUE #2922: [deep-link] No such file or directory when calling register_all
Running on  OS: NixOS 25.11.20250814.fbcf476 (Xantusia) x86_64

```rust
        .setup(|app| {
            #[cfg(any(target_os = "linux", all(windows, debug_assertions)))]
            {
                 app.deep_link().register_all()?;
            }
```

Returns an io::error `no such file or directory (os error 2)`.

The `data_dir` is writable since its in my home directory, not in the nix store

FabianLars:
this is actually a breaking change but i don't think anyone cares about that on our error types. the only alternative (instead of waiting for v3) that comes to mind right now is to map the error to `std::io::Error::other("blabla + source.display()")`

closes #2922

### DIFF
--- a/.changes/deep-link-linux-errors.md
+++ b/.changes/deep-link-linux-errors.md
@@ -1,0 +1,6 @@
+---
+deep-link: patch
+deep-link-js: patch
+---
+
+On Linux, improved error messages when OS commands fail.
--- a/plugins/deep-link/src/error.rs
+++ b/plugins/deep-link/src/error.rs
@@ -23,6 +23,9 @@ pub enum Error {
     #[cfg(target_os = "linux")]
     #[error(transparent)]
     ParseIni(#[from] ini::ParseError),
+    #[cfg(target_os = "linux")]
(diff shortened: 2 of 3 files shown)


lucasfernog (approved):

WSH032:
@FabianLars This actually broke `pytauri` 😂:

- <https://github.com/pytauri/pytauri/actions/runs/17373493358/job/49314523641#step:8:2084>
- <https://github.com/pytauri/pytauri/blob/734d5a1c7834935abdd39c85b53bf990a563febe/crates/pytauri-core/src/plugins/deep_link/mod.rs#L25-L35>

Could you revert this change and yank this version?

FabianLars:
Yours is probably the only project that's sensitive to this and even though you're so present here I still keep forgetting about it.

I'm not sure but I may have access to crates.io/npm with the Board role now, I'll try to look into it this week.

PR #2933, opened 2025-08-19
Seanaye/feat/websocket cookies

seanaye:
This PR aims to expose a way that allows plugin consumers to integrate the cookie store from http-client plugin with the websocket plugin. 

This is useful in the scenario that some websocket api expects cookies to be present in the initial http request header.

This also allows rust developers to read and interact with the cookies store if they so choose.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,66 @@
+{
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+    flake-utils.url = "github:numtide/flake-utils";
+    fenix.url = "github:nix-community/fenix";
+    fenix.inputs.nixpkgs.follows = "nixpkgs";
+  };
+
+  outputs =
+    {
+      self,
+      nixpkgs,
+      flake-utils,
+      fenix,
+    }:
+    flake-utils.lib.eachDefaultSystem (
+      system:
+      let
+        pkgs = import nixpkgs {
+          system = system;
+        };
+
+        packages = with pkgs; [
+          curl
+          wget
+          pkg-config
+
+          cargo-tauri
+          cargo-info
+          cargo-udeps
+
+          (
+            with fenix.packages.${system};
+            combine [
+              complete.rustc
+              complete.rust-src
+              complete.cargo
+              complete.clippy
+              complete.rustfmt
+              complete.rust-analyzer
+            ]
+          )
+        ];
+
+        libraries = with pkgs; [
+          gtk3
+          libsoup_3
+          webkitgtk_4_1
+          cairo
+          gdk-pixbuf
+          glib
+          dbus
+          openssl
+          librsvg
+        ];
+      in
+      {
+        devShell = pkgs.mkShell {
+          buildInputs = packages ++ libraries;
+
+          LD_LIBRARY_PATH = "${pkgs.lib.makeLibraryPath libraries}:$LD_LIBRARY_PATH";
+          XDG_DATA_DIRS = "${pkgs.gsettings-desktop-schemas}/share/gsettings-schemas/${pkgs.gsettings-desktop-schemas.name}:${pkgs.gtk3}/share/gsettings-schemas/${pkgs.gtk3.name}:$XDG_DATA_DIRS";
+        };
+      }
+    );
+}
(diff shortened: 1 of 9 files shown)


plugins/http/Cargo.toml
```diff
@@ -43,33 +43,29 @@ tracing = { workspace = true, optional = true }
...
 brotli = ["reqwest/brotli"]
 deflate = ["reqwest/deflate"]
 zstd = ["reqwest/zstd"]
-trust-dns = ["reqwest/trust-dns"]
+hickory-dns = ["reqwest/hickory-dns"]
 socks = ["reqwest/socks"]
 http2 = ["reqwest/http2"]
 charset = ["reqwest/charset"]
-macos-system-configuration = ["reqwest/macos-system-configuration"]
 unsafe-headers = []
```
FabianLars, line 69: these are all breaking changes and which cannot be done before v3

plugins/http/Cargo.toml
```diff
@@ -34,7 +34,7 @@ tauri-plugin-fs = { path = "../fs", version = "2.5.1" }
 urlpattern = "0.3"
 regex = "1"
 http = "1"
-reqwest = { version = "0.12", default-features = false }
+reqwest = { version = "0.13", default-features = false }
```
FabianLars, line 37: reqwest is re-exported in lib.rs which means it's part of the public api therefore semver breaking reqwest updates are breaking changes in the plugin -> reqwest has to stay on 0.12 for v2

plugins/http/src/lib.rs
```diff
@@ -21,9 +21,9 @@ mod scope;
 #[cfg(feature = "cookies")]
 const COOKIES_FILENAME: &str = ".cookies";
 
-pub(crate) struct Http {
+pub struct Http {
     #[cfg(feature = "cookies")]
-    cookies_jar: std::sync::Arc<crate::reqwest_cookie_store::CookieStoreMutex>,
+    pub cookies_jar: std::sync::Arc<crate::reqwest_cookie_store::CookieStoreMutex>,
```
FabianLars, line 26: maybeee a getter makes more sense than direct struct access 🤔

flake.nix
```diff
@@ -0,0 +1,66 @@
+{
```
FabianLars, line 1: no nix files (yet?) in this repo please. same for the lockfile of course

plugins/websocket/Cargo.toml
```diff
@@ -32,7 +32,7 @@ futures-util = "0.3"
 tokio = { version = "1", features = ["net", "sync"] }
 tokio-tungstenite = { version = "0.29" }
 rustls = { version = "0.23", default-features = false, features = [
-  "ring",
+  "aws_lc_rs",
```
FabianLars, line 35: we intentionally left this as ring because it is easier to compile*. changing it needs proper testing (including cross-compiling) and a coordinated effort across all crates.

\* i think they fixed this at least for native builds, iirc it required cmake on windows before which isn't a normal requirement for tauri apps. according to their readme windows 32bit isn't supported though which we still technically do :/

plugins/websocket/src/lib.rs
```diff
@@ -266,27 +273,45 @@ async fn send(
...
     pub fn new() -> Self {
         Self {
             tls_connector: None,
+            merge_headers: None,
         }
     }
 
+    /// add a callback which is able to modify the initial headers of the http upgrade request.
+    /// This is useful for scenarios where the frontend may not know all the required headers that must be sent.
+    /// e.g. in the scenario of http-only cookies
```
FabianLars, line 303: i think this could use an example or something. I have to admit i'm failing to understand it at first glance / or what the use case is

plugins/http/src/lib.rs
seanaye: will change

plugins/http/Cargo.toml
seanaye: I can split this out into a different PR if you want

flake.nix
seanaye: will remove

plugins/websocket/Cargo.toml
seanaye: will remove

plugins/websocket/src/lib.rs
seanaye: I can add an example. The use case is that certain websocket servers expect some specific http header in the connection request before it gets upgraded to websocket. The browser does this automatically for cookies headers but the tauri plugin does not. This means its currently impossible to authenticate against websocket servers which expect cookie headers

plugins/http/Cargo.toml
seanaye: will remove

PR #2960, opened 2025-08-26
feat(updater): inject bundle_type into endpoint url

FabianLars:
follow up to https://github.com/tauri-apps/plugins-workspace/pull/2624


FabianLars:
btw those schema changes are because i'm on windows. seems like we updated some dependency in the last few weeks?

plugins/updater/src/updater.rs
```diff
@@ -414,8 +414,15 @@ impl Updater {
...
                 .replace("{{target}}", target)
-                .replace("{{arch}}", self.arch)
-                .parse()?;
+                .replace("{{arch}}", self.arch);
+
+            if let Some(installer) = installer_for_bundle_type(bundle_type()) {
+                url = url
+                    .replace("%7B%7Bbundle_type%7D%7D", installer.name())
+                    .replace("{{bundle_type}}", installer.name());
+            }
```
beanow-at-crabnebula: Wouldn't this produce URLs with the placeholder still in it when the installer is None?

Say there's an endpoint like:
```
https://foo.example/update/my-app/{{target}}-{{arch}}/{{current_version}}?bundle_type={{bundle_type}}
```

Wouldn't that end up as:
```
https://foo.example/update/my-app/linux-x86_64/1.2.3?bundle_type={{bundle_type}}
```

FabianLars: yep. any suggestions? Should we replace that with `unknown` ?

lucasfernog: using unknown is fine - though this should only happen in dev mode right? or if you run the target/release/app binary directly

lucasfernog: (and i raised this concern before, but we should avoid using unknowns ever - maybe use a default value in that case)

FabianLars: > using unknown is fine - though this should only happen in dev mode right? or if you run the target/release/app binary directly

Yes, if the dev uses the tauri cli to package the app this should never be unknown. If they're using their own packaging system it'll show unknown (or the last created installer)

FabianLars: > (and i raised this concern before, but we should avoid using unknowns ever - maybe use a default value in that case)

not sure i get what you mean. any kind of default value we set (assuming default means one of the existing types?) has a high chance to be incorrect.

beanow-at-crabnebula: > yep. any suggestions? Should we replace that with unknown ?

Having pulled the last of my hair out over Helm charts. The reasons this gets tricky is *string templating*.
To handle nullable values it pushes you down the route of conditional rendering or default values, and it's not fun :grimacing:

So my first suggestion would be, think about alternatives that support nulls easily.
Like a header with these values in JSON.

As for `{{bundle_type}}`. I do think `unknown` would make it slightly more useful than say _empty string_ because then you can use it in paths as well as query strings.

beanow-at-crabnebula: Another option could be we we don't support this as a `{{bundle_type}}` placeholder and instead only support this through the query string. Because we then can assume how query strings work, we can omit it when None.

So it's either:
- `https://foo.example/update/my-app/linux-x86_64/1.2.3`
- `https://foo.example/update/my-app/linux-x86_64/1.2.3?bundle_type=appimage`

lucasfernog (approved):
i think this is the best approach considering we want to align with the existing templating system (i.e. changing for custom query parameters etc would be a bigger change)

FabianLars:
agreed, though i do like the Headers idea but that's something for another PR.

PR #2969, opened 2025-09-01
docs(fs): `fs:scope` usage

Legend-Master:
It's not obvious that this permission should be used in this syntax:

```json
{
  "identifier": "fs:scope",
  "allow": [
    "$APPDATA/documents/**/*"
  ],
  "deny": [
    "$APPDATA/documents/secret.txt"
  ]
}
```

adding in an example for this

### DIFF
--- a/plugins/fs/permissions/autogenerated/reference.md
+++ b/plugins/fs/permissions/autogenerated/reference.md
@@ -3759,10 +3759,10 @@ An empty permission you can use to modify the global scope.
     {
       "identifier": "fs:scope",
       "allow": [
-        "$APPCONFIG/documents/**"
+        "$APPDATA/documents/**"
       ],
       "deny": [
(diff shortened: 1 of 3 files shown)


FabianLars:
I think we shouldn't use `**` to match files in our docs as people confuse that with `**/*` all the time (or was the latter what you meant to use?)

Legend-Master:
> I think we shouldn't use `**` to match files in our docs as people confuse that with `**/*` all the time (or was the latter what you meant to use?)

Nice catch, guess I'm one of them 😂, I meant to use `**/*`, to be honest though, `**` really feels like it could match everything recursively
Also the [glob documentation](https://man7.org/linux/man-pages/man7/glob.7.html) sucked, I wish I learned it through https://docs.rs/glob/latest/glob/struct.Pattern.html

FabianLars (approved):

PR #2970, opened 2025-09-02
fix(deep-link): revert the breaking change introduced by #2928

WSH032:
refer to <https://github.com/tauri-apps/plugins-workspace/pull/2928#issuecomment-3241824550>

@FabianLars Can we release a patch version as soon as possible? This issue caused the pytauri v0.8 release yesterday to be interrupted at the final stage.

---

Building on Windows results in differences in `**/schema.json` (different newline), any ideas? I'm concerned this might once again cause issue： https://github.com/tauri-apps/tauri/pull/13597

### DIFF
--- a/.changes/fix-deep-link-semver.md
+++ b/.changes/fix-deep-link-semver.md
@@ -1,0 +1,6 @@
+---
+deep-link: patch
+deep-link-js: patch
+---
+
+Revert the breaking change introduced by [#2928](https://github.com/tauri-apps/plugins-workspace/pull/2928).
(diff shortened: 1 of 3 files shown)


FabianLars:
Can you log::error the error instead and add a to-do saying that we change it into an error in v3? I can't make code changes at the moment so that would help me :)

---

I brought the Windows gen thing up as well but  we didn't spot what's causing it yet (I don't think anybody really looked into it). If possible I'd prefer if you could revert those files just to be sure the docs generator won't struggle.

WSH032:
Sure, I'll do it soon.

> Can you log::error the error instead

Do you mean to revert it back to `#[error(transparent)]`?

FabianLars:
Ah i didn't even notice that, but yes, that should be reverted as well since there are a few more places where io errors may be emitted (eg when we write files).

Just thought that as a bridge to v3 we could provide that context as a log instead.

WSH032:
> there are a few more places where io errors may be emitted (eg when we write files)

My bad, I missed those. Done.

FabianLars (approved):

Issue #2847, opened 2025-07-14
[Feature Request: upload] Allow PUT requests for upload


cohogan:
I’m using tauri-plugin-upload to send files to a Google Cloud Storage bucket, but GCS presigned URLs for non-resumable uploads require PUT requests--currently the plugin only supports POST.

It would be great if the plugin allowed specifying the HTTP method (e.g., PUT) for uploads. This would make it compatible with GCS and similar services. For now, I'm writing Rust, but I love the simplicity of this API and wish it worked for my use case.

(continuation of [this duscussion post](https://github.com/tauri-apps/tauri/discussions/13823))

velocitysystems:
@cohogan I've done a WIP of this work [here](https://github.com/velocitysystems/plugins-workspace/commit/6678004742eb4e32faaccbcacd78f2e020d99b53) which I'll publish once [PR 2838](https://github.com/tauri-apps/plugins-workspace/pull/2838) is merged.

velocitysystems:
@cohogan Aside from the HTTP method are there any other features which are missing for your application?

cohogan:
I think just the HTTP method would be a big improvement for my app. Thank you @velocitysystems I am really excited!

One other potentially nice to have feature would be support for resumable/multipart uploads, but that seems like overkill for most apps and what this plugin would be expected to support.

cohogan:
@velocitysystems just following up on this. Super excited about this feature. Appreciate any updates on when you think the [http method commit](https://github.com/velocitysystems/plugins-workspace/commit/6678004742eb4e32faaccbcacd78f2e020d99b53) you made might be published!

velocitysystems:
@cohogan Published #2991 to add HTTP method support.

velocitysystems:
@cohogan #2991 has been merged and will be available in the next plugins release.

PR #2991, opened 2025-09-09
feat(upload): Add HTTP method selection for upload

velocitysystems:
Resolves:
#2847 

Key Changes
- Support HTTP POST, PUT and PATCH methods
- Update example page
- Update tests
- Bump version to 2.4.0

image


FabianLars:
Thanks for the PR! The main maintainers are out of office (or will be in the next 1-2 days) for the next 2-3 weeks so this may get delayed quite a bit.

P.S. Please revert the changes in Cargo.toml and package.json. Versioning will be handled in CI.

velocitysystems:
Thanks @FabianLars - Changes made, good to know versioning is handled by CI.
No urgency on this; it's been on my radar for a few weeks and wanted to get it done while I remember.

FabianLars:
what do you think about using https://docs.rs/http/1.3.1/http/method/struct.Method.html instead of a custom enum? It's already accessible via `reqwest::Method` or `tauri::http::Method` i think. (really asking for your opinion, not telling you what to do)

velocitysystems:
> what do you think about using https://docs.rs/http/1.3.1/http/method/struct.Method.html instead of a custom enum? It's already accessible via `reqwest::Method` or `tauri::http::Method` i think. (really asking for your opinion, not telling you what to do)

@FabianLars Good question. Had a look and while we could use `reqwest::Method` it doesn't support the `Deserialize` trait. This means we'd need to deserialize the value as a string in Rust and then parse it using `Method::from_str`. While this would work we'd still need a corresponding enum in the Typescript API.

With this in mind it seems simpler/safer to just use a custom enum type in Rust. What do you think?

velocitysystems:
@FabianLars Rebased - All tests are passing locally.

jiangjiji:
Very helpful

FabianLars:
> With this in mind it seems simpler/safer to just use a custom enum type in Rust. What do you think?

I don't think there's much of a difference, _but_ when i wrote my comment i thought the plugin had a rust api but since that's not true, i'm fine with your approach.

p.s. re-run `pnpm build` to fix CI

velocitysystems:
> > With this in mind it seems simpler/safer to just use a custom enum type in Rust. What do you think?
> 
> I don't think there's much of a difference, _but_ when i wrote my comment i thought the plugin had a rust api but since that's not true, i'm fine with your approach.
> 
> p.s. re-run `pnpm build` to fix CI

Thanks @FabianLars. Just fixed CI build per your comments.

FabianLars (approved):
thanks :)

PR #2995, opened 2025-09-12
feat(single-instance): add logging when closing a 2nd instance

aecsocket:
While debugging a Tauri app which was crashing on startup, I got into a state where every time I opened the app, it would close unexpectedly with exit code 0. I had to open `gdb`, `break exit` and use `bt` to figure out that it was actually Tauri closing the app due to the `single-instance` plugin. I had a zombie process taking the DBus `SingleInstance` name for my app.

This PR adds a `tracing::info` message when `exit(0)`ing the app on Linux if the single instance DBus name is already taken.

### DIFF
--- a/plugins/single-instance/src/platform_impl/linux.rs
+++ b/plugins/single-instance/src/platform_impl/linux.rs
@@ -86,6 +86,9 @@ pub fn init<R: Runtime>(f: Box<SingleInstanceCallback<R>>) -> TauriPlugin<R> {
                             ),
                         );
                     }
+                    tracing::info!(
+                        "Closing this duplicate instance as DBus name {dbus_name} is already taken"
+                    );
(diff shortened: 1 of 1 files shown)


plugins/single-instance/src/platform_impl/linux.rs
```diff
@@ -86,6 +86,9 @@ pub fn init<R: Runtime>(f: Box<SingleInstanceCallback<R>>) -> TauriPlugin<R> {
                             ),
                         );
                     }
+                    tracing::info!(
```
Legend-Master, line 89: We probably want tochange this to `debug` level

Legend-Master (commented):
I think we should do the same for other platforms as well? (you can find them from `std::process::exit(0)` calls)

Also do you mind also adding a change file?

PR #3010, opened 2025-09-25
fix(geolocation): use timeout for min,max,interval on android

### LINKED ISSUE #3009: Geolocation update interval on Android slow
Hi,

in [Geolocation.kt](https://github.com/tauri-apps/plugins-workspace/blob/v2/plugins/geolocation/android/src/main/java/Geolocation.kt#L96) I see `setMinUpdateIntervalMillis(5000)`.
We experience slow updates on Android, so I assume this is the issue. Is there a reason it is set to 5000? Is there a way to change this number, without forking this repo?

``` kt
val locationRequest = LocationRequest.Builder(10000)
    .setMaxUpdateDelayMillis(timeout)
    .setMinUpdateIntervalMillis(5000)
    .setPriority(prio)
    .build()
```

FabianLars:
closes #3009 

this really needs some more thinking at some point but i'm not that much into android at the moment so this PR is a bit of a bandage until then.

### DIFF
--- a/.changes/geolocation-android-timeout.md
+++ b/.changes/geolocation-android-timeout.md
@@ -1,0 +1,6 @@
+---
+geolocation: patch
+geolocation-js: patch
+---
+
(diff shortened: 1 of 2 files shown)


FabianLars:
we really need some kotlin / swift CI at some point

PR #3018, opened 2025-09-29
fix(opener): doesn't open same origin links

### LINKED ISSUE #2990: [opener] Links to urls with explicit port number do not open if window content is served from that same port
Dear Tauri team,

I have created a minimal reproduction repository at https://github.com/simonrepp/tauri_opener_port_reproduce/tree/main that reproduces the below described issue, which I currently assume to be a bug (although I'm happy about other explanations just the same :)).

## Description

- Scenario: A tauri app that opens a single window:

```rs
        .setup(|app| {
            let webview_url = WebviewUrl::External(Url::parse("http://127.0.0.1:8080").unwrap());
            WebviewWindowBuilder::new(app, "main", webview_url)
                .build()
                .unwrap();

            Ok(())
        })
```

- In the context of the reproduction, the server that hosts `http://127.0.0.1:8080` is started from the tauri app itself as well (in a separate thread), it serves the following content (the links are all of the form `<a href="http://127.0.0.1:1234/" target="_blank">http://127.0.0.1:1234/</a>`):
Image

- The capabilities have been set up so that opening the technically "remote" links is permitted for the opener plugin:

```rs
{
  "$schema": "../gen/schemas/desktop-schema.json",
  "identifier": "default",
  "description": "Capability for the main window",
  "windows": ["main"],
  "permissions": [
    "core:default",
    "opener:default"
  ],
  "remote": {
    "urls": ["http://*:*/*"]        <---------
  }
}

```
- Now here's the problem: All the links open fine in the default system browser (producing 404s because the ports are mostly invented, that's fine), **except** those that include the port of the actually running server, from which the window's content is served (`:8080`). This is regardless of whether the link points to exactly the same page that is provided in the window (`http://127.0.0.1:8080`), or one that has an additional `?escape` or even `/escape/?escape#escape` suffix. Basically, I click them and nothing happens, and there is no error in the developer tools console in the window either.

I have spent well over 2 hours looking for possible explanations for this, as well as created a reproduction repository to possibly stumble upon the solution myself in the process, but as that did not succeed I'm now opening this as an issue. Happy about any ideas and help!

Thank you so much for Tauri and all your work, it's much appreciated! Looking forward to figuring this out, please let me know if I can provide any further information of course!

Legend-Master:
Fix #2990

So we're only opening links if they're supposed to be opened in new tabs/windows, no need to check for same origin here


FabianLars (approved):
(untested)

Legend-Master:
> (untested)

Have tested myself this time 🙃

Guess I'm not waiting for amr then, feel free to chime in though 🙃

PR #3019, opened 2025-09-30
fix(deep-link): update the Exec= in handler if the executable path changes

kevinschoonover:
I noticed that deep-linking recently broke on my test application in Linux and debugged it to our project folder structure was reorganized leading to the executable path being different. The deep-linking plugin today doesn't check if the Exec field is different and requires it to be deleted manually in order to properly update.

This PR updates the `register` function to consider if the `Exec` path has changed and will update it if needed.

### DIFF
--- a/.changes/deep-link-fix-different-exec.md
+++ b/.changes/deep-link-fix-different-exec.md
@@ -1,5 +1,6 @@
 ---
 "deep-link": patch
+"deep-link-js": patch
 ---
 
 Fix Exec= field in desktop handler if executable path changes
--- a/.changes/deep-link-fix-different-exec.md
+++ b/.changes/deep-link-fix-different-exec.md
@@ -1,0 +1,5 @@
+---
+"deep-link": patch
+---
+
+Fix Exec= field in desktop handler if executable path changes
--- a/plugins/deep-link/src/lib.rs
+++ b/plugins/deep-link/src/lib.rs
@@ -317,7 +317,7 @@ mod imp {
                         }
 
                         // if the exec command doesnt match, update to the new one
-                        let old_exec = section.remove("Exec").unwrap_or(qualified_exec.clone());
+                        let old_exec = section.remove("Exec").unwrap_or_default();
                         if old_exec != qualified_exec {
                             section.append("Exec", qualified_exec);
                             change = true;
--- a/plugins/deep-link/src/lib.rs
+++ b/plugins/deep-link/src/lib.rs
@@ -293,6 +293,7 @@ mod imp {
                     .unwrap_or_else(|| bin.into_os_string())
                     .to_string_lossy()
                     .to_string();
+                let qualified_exec = format!("{} %u", exec);
 
                 let target = self.app.path().data_dir()?.join("applications");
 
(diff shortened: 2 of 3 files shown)


plugins/deep-link/src/template.desktop
```diff
@@ -1,7 +1,7 @@
 [Desktop Entry]
 Type=Application
 Name={name}
-Exec={exec} %u
+Exec={qualified_exec}
 Terminal=false
 MimeType={mime_type}
-NoDisplay=true
\ No newline at end of file
+NoDisplay=true
```
kevinschoonover, line 7: i am not sure what this diff is with `NoDisplay`, but happy to remove it if needed

FabianLars: you just added an empty newline there which is actually a good thing as it acts as a terminator so we'll keep it :)

This symbol means that there was no newline at the end:
grafik
(screenshot of the left side of this diff)

FabianLars (commented):
Thanks for the PR! This is indeed how it's supposed to work and how it _used_ to work until we added the `ini` crate i think.

Can you also add a small changefile please? Similar to https://github.com/tauri-apps/plugins-workspace/pull/3018/files#diff-6e7dd8c86c89bc4897dc64adef7ec8eb12a051e59696f039fb9d1e63dbaa928f but with deep-link instead of opener.

kevinschoonover:
> Thanks for the PR! This is indeed how it's supposed to work and how it _used_ to work until we added the `ini` crate i think.
> 
> Can you also add a small changefile please? Similar to https://github.com/tauri-apps/plugins-workspace/pull/3018/files#diff-6e7dd8c86c89bc4897dc64adef7ec8eb12a051e59696f039fb9d1e63dbaa928f but with deep-link instead of opener.

Should now be done! Let me know if I should make any more changes

plugins/deep-link/src/lib.rs
```diff
@@ -304,12 +305,28 @@ mod imp {
...
+                        // if the mime type is not present, append it to the list
                         if !old_mimes.split(';').any(|mime| mime == mime_type) {
                             section.append("MimeType", format!("{mime_type};{old_mimes}"));
+                            change = true;
+                        } else {
+                            section.insert("MimeType".to_string(), old_mimes);
+                        }
+
+                        // if the exec command doesnt match, update to the new one
+                        let old_exec = section.remove("Exec").unwrap_or(qualified_exec.clone());
```
FabianLars:
```suggestion
                        let old_exec = section.remove("Exec").unwrap_or_default();
```
I think something like this makes more sense here just in case that Exec is empty or doesn't exist (pretty unrealistic)

.changes/deep-link-fix-different-exec.md
```diff
@@ -0,0 +1,5 @@
+---
+"deep-link": patch
```
FabianLars, line 2:
```suggestion
"deep-link": patch
"deep-link-js": patch
```

this part is a bit annoying at the moment

FabianLars:
> Let me know if I should make any more changes

Nope, you're done. Thanks!

For the next PR (🥹) we'd appreciate it if you could configure commit signing though. In theory we require this in our repos but since it's annoying to sign past commits i'll abuse my permissions and merge it anyway.

FabianLars (approved):

PR #3022, opened 2025-10-01
feat(window-state): add `with_dir()` method for setting custom directory

### LINKED ISSUE #3020: Specify where the Window State plugin's JSON file is stored
### Problem
Currently, the Window State plugin saves the `.window-state.json` file in a fixed location: `${dataDir}/${bundleIdentifier}/`. This creates issues for applications that need to maintain full portable installations where all data stays in the app/binary directory.

### Proposed Solution
Add a `with_dir()` method to the `Builder` that allows specifying a custom directory for the window state file, similar to how `with_filename()` allows custom filenames.

**Suggested API:**
```rust
pub fn with_dir<P: AsRef<std::path::Path>>(self, dir: P) -> Self
```

**Example Usage:**
```rust
// Relative path
    .with_state_flags(StateFlags::SIZE | StateFlags::POSITION)
    .with_dir("./config")
    .with_filename("my-window-state.json")
    .build()

// Absolute path
    .with_state_flags(StateFlags::SIZE | StateFlags::POSITION)
    .with_dir("C:\\Users\\Username\\AppData\\Local\\MyApp")
    .with_filename("my-window-state.json")
    .build()
```

zevnda:
closes #3020


FabianLars:
hmm, i feel like we should merge filename and dir in v3, not sure 🤔 

thanks for the pr :)

Issue #3030, opened 2025-10-07
[bug] iOS file picker filter, selection, and security-scoped resource issues


velocitysystems:
### Describe the problem

On iOS, when using [plugin-dialog](https://v2.tauri.app/plugin/dialog/) to select a file, the API does not behave as expected when filtering or selecting files.
Custom file extensions (e.g. `.foo`) appear in the picker but cannot be selected. Additionally, unless MIME types or known extensions (like "txt") are passed, the picker defaults to an image picker instead of a file picker — two distinct modes on iOS.

Further, proper handling of security-scoped resources is missing or incomplete, which is required for persistent file access on iOS.

Repro (a)
1. Open the file picker via the Tauri plugin on an iOS device.
2. Provide a custom file type filter (e.g. ["foo", "bar"]).
3. Observe that matching files appear in the picker but cannot be selected.

Repro (b)
1. Open the file picker via the Tauri plugin on an iOS device.
2. Try calling the picker without MIME types or standard extensions (e.g. "txt").
3. Note that an image picker appears instead of the expected file picker.

### Describe the solution you'd like

Files with custom extensions in the filter should be selectable.
- The picker should open the `UIDocumentPickerViewController` (file picker), not the `UIImagePickerController` (image picker), when no image MIME types are provided.
- Returned file handles should properly include security-scoped resource access, allowing safe reading and writing of files outside the app sandbox.

### Alternatives considered

N/A

### Related

#3029
#1578
#1596
[#3716](https://github.com/tauri-apps/tauri/issues/3716)
[PR 2061](https://github.com/tauri-apps/plugins-workspace/pull/2061)

velocitysystems:
@FabianLars In addition to the change [proposed](https://github.com/tauri-apps/plugins-workspace/issues/3029#issuecomment-3386839306) for Android, we'd like to propose this enhancement to the `open()` method:

### 1. Select picker mode

**Current:**
iOS attempts to infer which picker mode to use (document, media) based on the MIME type.

**Proposed:**
Rather than infer the picker "mode" allow the user to specify a _preferred_ mode:
```ts
enum PickerMode {
  Auto = 'auto',          // Automatically choose based on MIME types (default)
  Document = 'document',  // Use document picker
  Media = 'media'         // Use media picker
}

interface OpenDialogOptions {
  ...
  /**
   * Preferred picker mode for mobile platforms
   * Determines which native picker to use (e.g., document vs. media picker).
   */
[...]

Issue #1876, opened 2024-10-03
[bug] iOS app crashes on devices when `tauri_plugin_nfc` is used


IT-ess:
# Description
On a brand new Tauri v2 project, when the official `tauri_plugin_nfc` is added to the project with `pnpm tauri add nfc`, and both `Info.ios.plist` and entitlements files are modified with the correct lines from the [documentation](https://v2.tauri.app/plugin/nfc/), the app crashes when launched on a real device. On a simulator, the app launches correctly. 

I also tried to add `CoreNFC` to the iOS Frameworks to bundle into the app (`tauri.conf.json : bundle > iOS > frameworks`), after reinitiating the iOS project with `pnpm tauri ios init`, it still crashes on launch.

# Reproduction
- Create new Tauri v2 project with `pnpm create tauri-app testnfc --template svelte-ts --manager pnpm` (I also tried with npm and vanilla JS, with the same result)
- `cd` in the project folder, install dependencies and run `pnpm tauri ios init`
- Check that the app compiles on your device with `pnpm tauri ios dev`.
- Then, delete the `src-tauri/gen/apple` folder, run `pnpm tauri ios init` again, and run `pnpm tauri add nfc`
- Finally, add required lines from the [documentation](https://v2.tauri.app/plugin/nfc/) in both Info.ios.plist and entitlements files, and run `pnpm tauri ios dev` again. 

The app crashes on launch on your device. I tested with either an iPhone 12 or an iPad 2017.

# `pnpm tauri info` output
```
[✔] Environment
    - OS: Mac OS 14.7.0 arm64 (X64)
    ✔ Xcode Command Line Tools: installed
    ✔ rustc: 1.80.1 (3f5fd8dd4 2024-08-06)
    ✔ cargo: 1.80.1 (376290515 2024-07-16)
    ✔ rustup: 1.27.1 (2024-04-24)
    ✔ Rust toolchain: stable-aarch64-apple-darwin (default)
    - node: 20.14.0
    - pnpm: 9.7.0
    - yarn: 1.22.22
    - npm: 10.7.0

[-] Packages
    - tauri 🦀: 2.0.1
    - tauri-build 🦀: 2.0.1
    - wry 🦀: 0.44.1
    - tao 🦀: 0.30.2
    - tauri-cli 🦀: 2.0.1
    - @tauri-apps/api : 2.0.1
    - @tauri-apps/cli : 2.0.0

[-] Plugins
[...]

PR #3034, opened 2025-10-13
feat(dialog) - Support picker mode for open dialog (#3030)

onehumandev:
For iOS and Android, there are 2 distinct file picker dialogs, one for picking media and one for picking documents.

Previously, the filters provided by the user when calling the `open` method would determine which picker would be displayed.

However, this has led to undefined behavior when no filter was provided.

To resolve this, we now provide a PickerMode (meant for iOS and Android) to explicitly define which picker we want to display.

Eventually, we may need to provide more explicit ways of filtering by MIME type or having specific modes for ImagePicker or VideoPicker for ease of use on mobile platforms.

But for now, this is an initial implementation that allows specifying which UI would be preferable for a file picker on mobile platforms.

Related bug #3030

### DIFF
--- a/examples/api/src/views/Dialog.svelte
+++ b/examples/api/src/views/Dialog.svelte
@@ -8,6 +8,7 @@
   let filter = null;
   let multiple = false;
   let directory = false;
+  let pickerMode = "";
 
   function arrayBufferToBase64(buffer, callback) {
     var blob = new Blob([buffer], {
@@ -65,6 +66,7 @@
         : [],
       multiple,
       directory,
+      pickerMode: pickerMode === "" ? undefined : pickerMode,
     })
       .then(function (res) {
         if (Array.isArray(res)) {
@@ -94,7 +96,7 @@
                 onMessage(res);
               }
             })
-            .catch(onMessage(res));
+            .catch(onMessage);
         }
       })
       .catch(onMessage);
@@ -112,7 +114,7 @@
             },
           ]
         : [],
-    })
+      })
       .then(onMessage)
       .catch(onMessage);
   }
@@ -141,6 +143,16 @@
 <div>
   <input type="checkbox" id="dialog-directory" bind:checked={directory} />
   <label for="dialog-directory">Directory</label>
+</div>
+<div>
+  <label for="dialog-picker-mode">Picker Mode:</label>
+  <select id="dialog-picker-mode" bind:value={pickerMode}>
+    <option value="">None</option>
+    <option value="media">Media</option>
+    <option value="image">Image</option>
+    <option value="video">Video</option>
+    <option value="document">Document</option>
+  </select>
 </div>
 <br />
 
--- a/plugins/dialog/guest-js/index.ts
+++ b/plugins/dialog/guest-js/index.ts
@@ -14,6 +14,16 @@ interface DialogFilter {
   name: string
   /**
    * Extensions to filter, without a `.` prefix.
+   *
+   * **Note:** Mobile platforms have different APIs for filtering that may not support extensions.
+   * iOS: Extensions are supported in the document picker, but not in the media picker.
+   * Android: Extensions are not supported.
+   *
+   * For these platforms, MIME types are the primary way to filter files, as opposed to extensions.
+   * This means the string values here labeled as `extensions` may also be a MIME type.
+   * This property name of `extensions` is being kept for backwards compatibility, but this may be revisited to
+   * specify the difference between extension or MIME type filtering.
+   *
    * @example
    * ```typescript
    * extensions: ['svg', 'png']
@@ -30,7 +40,14 @@ interface DialogFilter {
 interface OpenDialogOptions {
   /** The title of the dialog window (desktop only). */
   title?: string
-  /** The filters of the dialog. */
+  /**
+   * The filters of the dialog.
+   * On mobile platforms, if either:
+   * A) the {@linkcode pickerMode} is set to `media`, `image`, or `video`
+   * -- or --
+   * B) the filters include **only** either image or video mime types, the media picker will be displayed.
+   * Otherwise, the document picker will be displayed.
+   */
   filters?: DialogFilter[]
   /**
    * Initial directory or file path.
@@ -52,6 +69,13 @@ interface OpenDialogOptions {
   recursive?: boolean
   /** Whether to allow creating directories in the dialog. Enabled by default. **macOS Only** */
   canCreateDirectories?: boolean
+  /**
+   * The preferred mode of the dialog.
+   * This is meant for mobile platforms (iOS and Android) which have distinct file and media pickers.
+   * If not provided, the dialog will automatically choose the best mode based on the MIME types or extensions of the {@linkcode filters}.
+   * On desktop, this option is ignored.
+   */
+  pickerMode?: PickerMode
 }
 
 /**
@@ -76,6 +100,16 @@ interface SaveDialogOptions {
   /** Whether to allow creating directories in the dialog. Enabled by default. **macOS Only** */
   canCreateDirectories?: boolean
 }
+
+/**
+ * The preferred mode of the dialog.
+ * This is meant for mobile platforms (iOS and Android) which have distinct file and media pickers.
+ * On desktop, this option is ignored.
+ * If not provided, the dialog will automatically choose the best mode based on the MIME types or extensions of the {@linkcode filters}.
+ *
+ * **Note:** This option is only supported on iOS 14 and above. This parameter is ignored on iOS 13 and below.
+ */
+export type PickerMode = 'document' | 'media' | 'image' | 'video'
 
 /**
  * Default buttons for a message dialog.
--- a/plugins/dialog/src/lib.rs
+++ b/plugins/dialog/src/lib.rs
@@ -9,7 +9,7 @@
     html_favicon_url = "https://github.com/tauri-apps/tauri/raw/dev/app-icon.png"
 )]
 
-use serde::Serialize;
+use serde::{Deserialize, Serialize};
 use tauri::{
     plugin::{Builder, TauriPlugin},
     Manager, Runtime,
@@ -43,6 +43,15 @@ use mobile::*;
 pub use desktop::Dialog;
 #[cfg(mobile)]
 pub use mobile::Dialog;
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[serde(rename_all = "lowercase")]
+pub enum PickerMode {
+    Document,
+    Media,
+    Image,
+    Video,
+}
 
 pub(crate) const OK: &str = "Ok";
 pub(crate) const CANCEL: &str = "Cancel";
@@ -369,6 +378,7 @@ pub struct FileDialogBuilder<R: Runtime> {
     pub(crate) file_name: Option<String>,
     pub(crate) title: Option<String>,
     pub(crate) can_create_directories: Option<bool>,
+    pub(crate) picker_mode: Option<PickerMode>,
     #[cfg(desktop)]
     pub(crate) parent: Option<crate::desktop::WindowHandle>,
 }
@@ -380,6 +390,7 @@ pub(crate) struct FileDialogPayload<'a> {
     file_name: &'a Option<String>,
     filters: &'a Vec<Filter>,
     multiple: bool,
+    picker_mode: &'a Option<PickerMode>,
 }
 
 // raw window handle :(
@@ -395,6 +406,7 @@ impl<R: Runtime> FileDialogBuilder<R> {
             file_name: None,
             title: None,
             can_create_directories: None,
+            picker_mode: None,
             #[cfg(desktop)]
             parent: None,
         }
@@ -406,6 +418,7 @@ impl<R: Runtime> FileDialogBuilder<R> {
             file_name: &self.file_name,
             filters: &self.filters,
             multiple,
+            picker_mode: &self.picker_mode,
         }
     }
 
@@ -463,6 +476,15 @@ impl<R: Runtime> FileDialogBuilder<R> {
     /// Set whether it should be possible to create new directories in the dialog. Enabled by default. **macOS only**.
     pub fn set_can_create_directories(mut self, can: bool) -> Self {
         self.can_create_directories.replace(can);
+        self
+    }
+
+    /// Set the picker mode of the dialog.
+    /// This is meant for mobile platforms (iOS and Android) which have distinct file and media pickers.
+    /// On desktop, this option is ignored.
+    /// If not provided, the dialog will automatically choose the best mode based on the MIME types of the filters.
+    pub fn set_picker_mode(mut self, mode: PickerMode) -> Self {
+        self.picker_mode.replace(mode);
         self
     }
 
--- a/.changes/dialog-file-picker-mode.md
+++ b/.changes/dialog-file-picker-mode.md
@@ -1,0 +1,6 @@
+---
+"dialog": minor
+"dialog-js": minor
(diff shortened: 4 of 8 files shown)


examples/api/src/views/Dialog.svelte
```diff
@@ -142,6 +144,15 @@
   <input type="checkbox" id="dialog-directory" bind:checked={directory} />
   <label for="dialog-directory">Directory</label>
 </div>
+<div>
+  <label for="dialog-preferred-mode">Preferred Mode:</label>
+  <select id="dialog-preferred-mode" bind:value={preferredMode}>
+    <option value="none">None</option>
```
velocitysystems: Nit: Maybe we can just use an empty string for `None` state.

plugins/dialog/src/lib.rs
```diff
@@ -44,6 +44,14 @@ pub use desktop::Dialog;
 #[cfg(mobile)]
 pub use mobile::Dialog;
 
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[serde(rename_all = "lowercase")]
```
velocitysystems, line 48: Use `#[serde(rename_all = "camelCase")]` even though this will be a no-op difference for these values.

plugins/dialog/src/lib.rs
```diff
@@ -44,6 +44,14 @@ pub use desktop::Dialog;
...
+pub enum PickerMode {
+    Auto,
```
velocitysystems: We should probably just remove the `Auto` option since this is the same as _not_ passing a preferred `PickerMode`.

velocitysystems (changes requested):

examples/api/src/views/Dialog.svelte
onehumandev: An empty string seems ambiguous or unclear; what about "No Preference" as the option string?

velocitysystems: My apologies; I meant like this: `<option value="">None</option>`.

Since we can't have a null value; an empty string is the closest to representing no value selected which is a typical pattern for HTML `<select>` elements.

examples/api/src/views/Dialog.svelte
```diff
@@ -65,6 +66,7 @@
         : [],
       multiple,
       directory,
+      preferredMode: preferredMode === "" ? undefined : preferredMode,
```
velocitysystems: Let's call this `pickerMode` to be more specific. Can we update the other usages too?

velocitysystems:
@onehumandev We have a couple of test failures due to Swift code which is targeting APIs > iOS 13.
While Tauri's minimum iOS version is iOS 9 and above [this](https://github.com/onehumandev/plugins-workspace/blob/8e13e98a0ffaea21296d0a7450ea9c381eb1de68/plugins/dialog/ios/Package.swift) plugin is targeting **iOS 13** and above.

Note: We may be able to raise the minimum version since iOS 14+ is the default for Tauri applications created with Tauri CLI v2.8 and above. This is also used by the [NFC plugin](https://v2.tauri.app/ja/plugin/nfc/#:~:text=Target%20IOS%20version,-Section%20titled%20“Target&text=The%20NFC%20plugin%20requires%20iOS,Xcode%20project%20to%20configure%20it.&text=...-,IPHONEOS_DEPLOYMENT_TARGET%20=%2014.0;,%3E%20Minimum%20Deployments%20%3E%20iOS%20configuration.).

onehumandev:
> NFC plugin

@velocitysystems It looks like the NFC docs say that the min version is iOS 14, but actual declared [min version is 13](https://github.com/tauri-apps/plugins-workspace/blob/v2/plugins/nfc/ios/Package.swift#L10-L13) .

For this MR, what is the preferred option: bumping the min version to 14, or changing the current implementation to work with iOS 13? I don't mind either way, but I'd rather not make a significant change without getting a consensus.

velocitysystems:
Thanks @onehumandev. To retain compatibility let's keep the minimum iOS version at 13; unless we have a very specific reason to bump the min SDK version.

onehumandev:
@lucasfernog @velocitysystems  I've started working through making some changes to fix the compilation errors with iOS 13.0, and I wanted to make sure I understand the expectation for these plugins (especially for future work).

Specifically: What defines the expected iOS deployment target for plugins?
Is it:
A) The deployment target defined when a plugin is created (currently iOS 14 as of August 16, from [this MR)](https://github.com/tauri-apps/tauri/pull/13997/files)
B) The deployment target of the `examples/api` project (currently iOS, 14 from the previously mentioned MR)
C) Variable (each plugin may have a different deployment target as needed)

I see that the NFC plugin [has some specific statements](https://github.com/tauri-apps/tauri-docs/pull/3455/files) stating that the min version is 14, which would indicate that this is not standard.
However, from what I can see, the deployment target when a plugin is created and when testing the plugin through `api/examples` is 14.0.

The reason I am asking about this is that I was unable to reproduce the compilation issue on this MR for iOS 13.0 until I locally changed the `examples/api/src-tauri/gen/apple/project.yml` to have a deployment target of 13.0.

If that is expected, that is fine as well, and I can work with that :) I would then like to publish a separate MR to the docs or guidelines explaining how the expected min versions for plugins are determined (for both iOS and Android).

I just want to make sure I understand if this is a defined expectation that all plugins support iOS 13.0 (and any higher deployment target is an exception), or if the deployment target is determined on a case-by-case basis.

Thanks!

Related discussion for NFC plugin: #1876

FabianLars:
>  C) Variable (each plugin may have a different deployment target as needed)

This. There's a baseline and individual plugins may increase that if needed.

I think our baseline is still considered to be 13.0 in the plugins even if new projects now target 14.0 by default. Not sure if we consider this a breaking change so changing the baselines may have to wait till v3.

onehumandev:
> > C) Variable (each plugin may have a different deployment target as needed)
> 
> This. There's a baseline and individual plugins may increase that if needed.
> 
> I think our baseline is still considered to be 13.0 in the plugins even if new projects now target 14.0 by default. Not sure if we consider this a breaking change so changing the baselines may have to wait till v3.

@lucasfernog  After looking into the iOS 13 support for this MR and discussing with @velocitysystems , we would like to request targeting this dialog plugin to iOS 14.

Rationale:
As of iOS 14, there is support for 1) [UTType](https://developer.apple.com/documentation/uniformtypeidentifiers/uttype-swift.struct) for better file and folder filtering, and 2) [PHPicker](https://developer.apple.com/documentation/PhotosUI/PHPickerViewController) for media selection.
Requiring iOS 13 would require branching logic in the [DialogPlugin.swift](https://github.com/tauri-apps/plugins-workspace/blob/v2/plugins/dialog/ios/Sources/DialogPlugin.swift) to allow for differing methodologies of setting the file filters (supporting both raw MIME type strings, as is currently supported, and UTType, which is the modern supported pattern).

With the above described complications, would it be acceptable to change the deployment target for the dialog plugin to 14? 

If so, would it be best for that to be done in this MR, or in a separate MR exclusively for updating the deployment target and removing any unneeded branching logic? 

Thanks for considering!

velocitysystems:
@FabianLars Per your comments above - are we happy to bump the deployment target to iOS 14 for this plugin?
This will allow us to use [UTType](https://developer.apple.com/documentation/uniformtypeidentifiers/uttype-swift.struct) which is an important part of this PR.

FabianLars:
Any estimate on how annoying the branching would be? As a softer approach, how about a kind of branching were what's not supported is simply no-op instead of trying to fallback to something? That should hopefully be not much work but would allow us to no raise the minimum version since we're maybe not sooo far off of v3  i honestly don't care about a breaking change like this in a minor release if this isn't feasible either

onehumandev:
@FabianLars The softer approach would be much easier. We can do that without much effort, since we'd be able to keep all the code that already exists. 
We can definitely document that the picker mode isn't supported below iOS 14, and passing in that value will have no effect if running on one of those deployment targets. 
It means nothing breaks for current users, because it means they'd get the exact same behavior as they're currently getting.

FabianLars:
That sounds like the least controversial approach to me then if you don't mind adding that :)

onehumandev:
@lucasfernog Done and done; ready for review :) Thanks!
c. @velocitysystems

velocitysystems (approved):

FabianLars:
sorry, meant to review today but didn't get to it. will do tomorrow since i'm stuck with my macbook then anyway. at first glance i don't expect to request any major changes, if at all. Thanks so far!

plugins/dialog/guest-js/index.ts
```diff
@@ -52,6 +59,27 @@ interface OpenDialogOptions {
...
+   * As a note, this is only fully implemented on iOS for the time being.
+   * Android does have a specific media picker, but it requires one of two of the below:
+   * 1) Defining exactly one mime type of media (for example, `image/*` or `video/*`) for use by the ACTION_GET_CONTENT intent.
+   * If both of these MIME types are provided to the filter, the Document picker will be displayed.
+   * 2) Using the registerForActivityResult API to register a result callback for the media picker as opposed to the ACTION_GET_CONTENT intent.
+   * This is the recommended way to implement the media picker on Android, as the ACTION_GET_CONTENT intent is currently marked as deprecated.
+   * However, using registerForActivityResult requires being called from within the actual Activity object (it is a protected method), for which Tauri currently
+   * does not have a hook.
+   * As a result, we currently only support the document picker on Android until either:
+   * 1) We allow for explicitly defining a VideoPicker or a ImagePicker on Android.
```
FabianLars: Would this require any more changes? If we just have to split up the `media` enum variant into `audio` or `video` this wouldn't sound to bad. Or we can simply bind this to the mime type so that users have to use media + _one_ mime type.

But i assume this will require more changes than just this, right?

plugins/dialog/guest-js/index.ts
```diff
@@ -30,7 +30,14 @@ interface DialogFilter {
...
   /** The title of the dialog window (desktop only). */
   title?: string
-  /** The filters of the dialog. */
+  /**
+   * The filters of the dialog.
+   * The behavior of filters on mobile platforms (iOS and Android) is different from desktop.
+   * On Android, we are not able to filter by extension, only by MIME type.
+   * On iOS, we are able to filter by extension in the document picker, but not in the media picker.
+   * In practice, if an extension for a video or image is included in the filters (such as `png` or `mp4`),
+   * the media picker will be displayed instead of the document picker (regardless of the {@linkcode pickerMode} value).
```
FabianLars: is this a ios limitation or our implementation (i didn't exactly study the code yet but i saw the filtersIncludeImage).

if this is ios' fault then alright, but if not, what i want to filter for jpg and pdf at the same time for example?

onehumandev: This is our limitation, but looking again I think it was a mistake from my end; I got myself turned around understanding the intent of logic that was is currently committed.

I will push up a commit shortly containing both a code fix and better documentation explaining how the filters and pickerMode work together.

plugins/dialog/guest-js/index.ts
onehumandev: Splitting up the enum is doable; my question would then be what the API should be when trying to allow selecting both images and videos.
Specifically, which of the below options would be preferred:

1)  pickerMode accepts an array
```ts
export type PickerMode = 'document' | 'image' | 'video'

// Usage to let the picker select both images and video
open({
      title: "My wonderful open dialog",
      defaultPath,
      filters,
      multiple,
      directory,
      pickerMode: ['audio', 'video'],
    })
```

2) The PickerMode enum is fine-grained
```ts
export type PickerMode = 'document' | 'media' | 'image' | 'video'

// Usage to let the picker select both images and video
open({
      title: "My wonderful open dialog",
      defaultPath,
      filters,
      multiple,
      directory,
      pickerMode: 'media',
    })
```

We'd still have the Android limitation where if they allow selecting *both* video and image, the document picker will be shown (instead of the media picker), but it will allow the media picker to be shown when selecting *either* video or images.

plugins/dialog/guest-js/index.ts
onehumandev: Changes have been pushed, which includes a fix for how we were building the extension and mime type filters on iOS.
I've updated the comments on the `DialogFilter.extensions` property for more explanation on how it is used and how mobile platforms handle filtering; please let me know if the comment should be changed or if the naming issue is something we should address in this MR.

plugins/dialog/guest-js/index.ts
FabianLars: The enum sounds better i think

I assume `audio` was a typo but that brings the question, are there audio pickers as well?

FabianLars: > 2) Using the registerForActivityResult API to register a result callback for the media picker as opposed to the ACTION_GET_CONTENT intent.
   * This is the recommended way to implement the media picker on Android, as the ACTION_GET_CONTENT intent is currently marked as deprecated.

I'm super tired right now so i may be wrong but i think we do have that: https://github.com/tauri-apps/plugins-workspace/blob/277a45f56c22d36ab397daf854aaafa475bb206a/plugins/dialog/android/src/main/java/DialogPlugin.kt#L72 -> https://github.com/tauri-apps/tauri/blob/ee3cc4a91bf1315ecaefe90f423ffd55ef6c40db/crates/tauri/mobile/android/src/main/java/app/tauri/plugin/PluginManager.kt#L126C7-L126C29 -> https://github.com/tauri-apps/tauri/blob/ee3cc4a91bf1315ecaefe90f423ffd55ef6c40db/crates/tauri/mobile/android/src/main/java/app/tauri/plugin/PluginManager.kt#L52

onehumandev: > The enum sounds better i think
> 
> I assume `audio` was a typo but that brings the question, are there audio pickers as well?

Whoops; yes, that was a typo :) There isn't a specific picker for audio files on either Android or iOS. The standard document picker is typically used to select audio files.

onehumandev: I've updated the enum to include specific options for `image` and `video`, and confirmed that everything works as expected on Android, with behavior parity  to iOS.
It also turns out I was wrong; Android *does* support a media picker that displays both images and videos. This is the behavior we have with the latest pushed changes, and the documentation has been simplified and clarified.

Re: `registerForActivityResult`
I remember @velocitysystems  and I looked at that implementation as well, but we were unsure if it would be an invasive change to add a new entry points into `PluginManager` to support the different file pickers, especially as we were trying to keep the changes scoped to this particular repo/plugin.
Would it make sense for this MR to go in as is with the support for the different picker modes, and then have another MR for the PluginManager to support those entry points and wire them into this plugin?

FabianLars: > Would it make sense for this MR to go in as is with the support for the different picker modes, and then have another MR for the PluginManager to support those entry points and wire them into this plugin?

Yeah, that's fine. I don't think the PluginManager will need changes at first glance though, seems like we can keep going with Intents here.

onehumandev: Right on; in that case, the current changes committed to this MR should be completely ready for review. 
Thanks very much for your communicativeness and patience with my learning the Tauri ecosystem :)

FabianLars: Thanks for your patience with my insanely long response times x)

It's already late in the evening here so i'll go over the PR again tomorrow and hopefully merge it straight away.

FabianLars (dismissed):

onehumandev:
@FabianLars Apologies, I had to fix the formatting error, and I realized I had not pushed a GPG signed commit. Both have been resolved. Thanks!

FabianLars (approved):

PR #3037, opened 2025-10-17
fix(single-instance): use a proper background thread to accept and receive unix socket streams

cynecx:
Previously this has blocked a whole tokio worker thread. Causing issues with task wakeups etc. We should never use blocking i/o operations inside an async task, that's why this PR moves that logic to a separate background thread. Alternatively we could still utilize an async task but we'd have to use the async aware UnixListener and co. which would require us to pull in a whole async runtime.

(There are still some things unknown to me because technically tokio should've been able to schedule the task on a different worker thread due to work stealing. But tokio-console shows that the hung task is still stuck in scheduling. EDIT: Ah I think this behavior I am seeing  might be the lifo optimization and it being non-stealable)

### DIFF
--- a/.changes/single-instance-macos-background-thread.md
+++ b/.changes/single-instance-macos-background-thread.md
@@ -1,0 +1,5 @@
+---
+"single-instance": patch
+---
+
+On macOS, accept and read incoming unix socket streams from a dedicated background thread instead of an async runtime task, so the blocking I/O calls don't stall the async executor.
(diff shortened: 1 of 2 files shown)


FabianLars:
Thanks for the PR! Would you mind adding a very small changefile like this as well please? https://github.com/tauri-apps/plugins-workspace/blob/85c83dd2a6ad6e52159ad0dbd1d33b8e6540d3bd/.changes/enhance-fs-error-message.md

cynecx:
@FabianLars Added the changelog entry and fixed the formatting issue.

cynecx:
I've rebased the branch to trigger another CI run. The last CI run failed due to an inrelated issue (lost connection to the runner or something).

cynecx:
Superseeded by https://github.com/tauri-apps/plugins-workspace/pull/3466.

PR #3038, opened 2025-10-20
feat(barcode-scanner): update `androidx.camera` from `1.1.0` to `1.5.1` to support 16 KB memory page sizes

### LINKED ISSUE #3023: [BUG] [barcode-scanner] support for 16KB page size
Hello,

main Info from here: https://developer.android.com/guide/practices/page-sizes

I encountered this while developing my app and recreated a new repo to reproduce this: https://github.com/TheLexoPlexx/tauri_repro_test

How to repro:

- `cargo tauri init`
- `cargo tauri android init`
- `cargo tauri add barcode-scanner`
- `cargo tauri android dev`
- error message

or clone the repo and `cargo tauri android dev`

The app still works and I honestly thought there would be an issue for this open already but apparently the overlap of people using the Android 16 Beta while developing a Tauri App that requires the barcode-scanner-plugin is not that large after all.

daniel-mader:
Starting November 1st, 2025, all Android builds [are required to support 16 KB memory page sizes](https://developer.android.com/guide/practices/page-sizes).

The plugin `barcode-scanner` currently uses `androidx.camera` dependency in version `1.1.0` ([June 2022](https://developer.android.com/jetpack/androidx/releases/camera#1.1.0)).

Google Play flags the following libraries to be incompatible with 16 KB page sizes:
- **base/lib/arm64-v8a/libimage_processing_util_jni.so**
- **base/lib/x86_64/libimage_processing_util_jni.so**

Upgrading to the latest version of the `androidx.camera` should resolve the issue.

### DIFF
--- a/.changes/bump-androidx-camera.md
+++ b/.changes/bump-androidx-camera.md
@@ -1,0 +1,6 @@
+---
+"barcode-scanner": patch
+"barcode-scanner-js": patch
+---
+
+Update `androidx.camera` from `1.1.0` to `1.5.1` to support 16 KB memory page sizes.
--- a/plugins/barcode-scanner/android/build.gradle.kts
+++ b/plugins/barcode-scanner/android/build.gradle.kts
@@ -33,16 +33,14 @@ android {
 }
 
 dependencies {
-
+    val camerax_version = "1.5.1"
     implementation("androidx.core:core-ktx:1.9.0")
     implementation("androidx.appcompat:appcompat:1.6.0")
     implementation("com.google.android.material:material:1.7.0")
-    implementation("androidx.camera:camera-core:1.1.0")
-    implementation("androidx.camera:camera-view:1.1.0")
-    implementation("androidx.camera:camera-lifecycle:1.1.0")
-    implementation("androidx.camera:camera-camera2:1.1.0")
-    implementation("androidx.camera:camera-lifecycle:1.1.0")
-    implementation("androidx.camera:camera-view:1.1.0")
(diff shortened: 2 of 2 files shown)


FabianLars:
Thanks for the PR! Would you mind adding a small changefile (check the .changes dir for examples) as well? :)

FabianLars (approved):

daniel-mader:
There seems to be an issue with an incompatible Gradle plugin version:

```
Dependency 'androidx.camera:camera-core:1.5.1' requires Android Gradle plugin 8.6.0 or higher.
This build currently uses Android Gradle plugin 8.5.1.
```

I am not sure if this is only happening with my local (cached) dependencies, or it will also generally break in this workspace.
When updating the plugin manually, I get:

```
Error resolving plugin [id: 'com.android.library', version: '8.13.0']
> The request for this plugin could not be satisfied because the plugin is already on the classpath with an unknown version, so compatibility cannot be checked.
```

Is there a plugin version for the entire workspace somewhere or is it my local config?

**EDIT:** Nevermind, you already opened another PR: https://github.com/tauri-apps/plugins-workspace/pull/3039 👍

FabianLars:
yeah i was a bit quick on the approve button. i forgot that sdk requirement changes aren't a breaking change in android land.

daniel-mader:
Just asking, why are you bumping to `8.11.0` instead of `8.13.0` (latest)?

FabianLars:
just matching tauri's version

Issue #2579, opened 2025-03-28
[bug] tauri-apps/updater check


AnthonyTB:
### Describe the bug

When putting a raw json endpoint in the updater.endpoints section of tauri.conf I get (error sending request for url ("endpoint url")). If I click on the link via the console output it opens just fine in another window. I then tried to switch to the github releases file download option and it gets the same error (error sending request for url ("its a weird objects.githubusercontent aws link that 401's"))

### Reproduction

tauri conf

"plugins": {
    "updater": {
      "active": true,
      "pubkey": "dW50cnVzdGVkIGNvbW1lbnQ6IG1pbmlzaWduIHB1YmxpYyBrZXk6IEQ1NDVEQ0NDNkE1Q0U0MQpSV1JCenFYR3pGMVVEYmJoZ1lvcFd1dDlxVk9TSHlZTGZva0plbnAvMjd4ODZQdDM3eHVmZEloOQo=",
      "endpoints": [
        "https://github.com/AnthonyTB/port-24/releases/latest/download/latest.json"
      ]
    }
  }

package.json

"@tauri-apps/plugin-process": "2.2.0",
"@tauri-apps/plugin-updater": "2.6.1",

js implementation

check({ timeout: 30000 })
        .then(async (data) => {
          console.log("j", { data });
          if (data) {
            data
              .downloadAndInstall()
              .then(async (res) => {
                console.log("h", { res });
                await relaunch();
              })
              .catch((i_err) => {
                console.log({ i_err });
              });
          } else {
            console.log("no update");
            setLoading(false);
          }
        })
        .catch((err) => {
          console.log({ err });
        });

### Expected behavior

Better error handling instead of just error requesting url and for it parse the json properly

### Full `tauri info` output

```text
[✔] Environment
    - OS: Windows 10.0.26100 x86_64 (X64)
    ✔ WebView2: 134.0.3124.93
    ✔ MSVC: Visual Studio Build Tools 2022
    ✔ rustc: 1.85.1 (4eb161250 2025-03-15)
    ✔ cargo: 1.85.1 (d73d2caf9 2024-12-31)
    ✔ rustup: 1.28.1 (f9edccde0 2025-03-05)
    ✔ Rust toolchain: stable-x86_64-pc-windows-msvc (default)
    - node: 23.9.0
    - npm: 10.9.2

[-] Packages
    - tauri 🦀: 2.4.0
    - tauri-build 🦀: 2.1.0
[...]

PR #3057, opened 2025-10-23
updater: allow accepting invalid TLS certs/hostnames via config

christian-leingang:
Problem: updater fails when contacting servers with self-signed or internal TLS certs and error message is generic.

Solution: add two plugin config flags `dangerousAcceptInvalidCerts` and `dangerousAcceptInvalidHostnames` and builder methods to override them. These mirror [http-plugin](https://v2.tauri.app/reference/javascript/http/#properties-1) and reqwest's `danger_accept_invalid_certs` and `danger_accept_invalid_hostnames`.

Security: these settings are dangerous and should only be used in trusted environments or testing.

### DIFF
--- a/.changes/updater-allow-invalid-tls.md
+++ b/.changes/updater-allow-invalid-tls.md
@@ -1,15 +1,6 @@
-```markdown
 ---
 "updater": minor
 "updater-js": minor
 ---
 
-Allow configuring the updater client to accept invalid TLS certificates and hostnames for
-internal/self-signed update servers. These options are available via the plugin config
-(`dangerousAcceptInvalidCerts`, `dangerousAcceptInvalidHostnames`) and via the
-`UpdaterBuilder` (`dangerous_accept_invalid_certs`, `dangerous_accept_invalid_hostnames`).
-
-These settings are gated behind the `dangerous-settings` Cargo feature and should only be
-used in trusted environments (tests, internal servers).
-
-```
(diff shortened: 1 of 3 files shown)


christian-leingang:
Would resolve the error ([Issue 2579](https://github.com/tauri-apps/plugins-workspace/issues/2579)) caused by strict TLS certificate validation.

FabianLars:
Thanks for the PR! would you mind adding a small changefile as well? There are some examples in the .changes dir :)

FabianLars:
argh, i just noticed that we export the Config struct which means that this is a breaking change........

Before thinking about alternative implementations, let us wait a few days and see how the discussions about an early v3 release we're currently having goes.

FabianLars (changes requested):
.

christian-leingang:
Thanks, I understand. I’ll wait for the v3 discussion and follow whatever the outcome is. If you prefer not to introduce a breaking change now, I can revert the Config edits and keep the functionality available only via the UpdaterBuilder (which should be non-breaking). Tell me which you’d like and I’ll update the PR accordingly.

FabianLars:
yeah, give us a few days to see if the v3 discussion is going anywhere. The UpdaterBuilder idea would be a good alternative indeed though a bit unfortunate that we'd then have 2 places for dangerous settings.

christian-leingang:
@FabianLars Just checking in on this since it's been a few weeks—how is the v3 discussion coming along?

If that timeline is still uncertain, would you prefer I switch this PR to the UpdaterBuilder approach (non-breaking) so we can get it merged into v2? I'm happy to do that if it helps unblock this.

FabianLars:
The discussion died before my comment apparently.

I also just realized that this Config object isn't actually used anywhere so perhaps we can be sneaky and just do this breaking change without anyone noticing.

The last hurdle is that i'm currently blocking updater releases, see https://github.com/tauri-apps/plugins-workspace/issues/3108 - maybe i'll revert those changes first before merging new ones, we'll see :/

FabianLars (approved):
fix for the linked issue is on the horizon so i'll start merging updater PRs

PR #3065, opened 2025-10-25
feat(log): Allow a log formatter per target

BinaryMuse:
This PR adds a `format` method to `Target`, allowing each target to have a different formatter. It also adds `reset_format` to the Builder to remove the default formatter.

# Motivation

When using the `with_colors` method with both a console-based and a file-based log target, the control codes for the colors are included in the log file. A custom formatter can be used when using `TargetKind::Dispatch`, but then you lose out on the automatic file rotation that the plugin provides.

Allowing a formatter per target solves this issue with techniques like the following:

```rust
Builder::new()
    .reset_format()
    .targets([
        Target::new(TargetKind::Stderr)
        .format(
            move |out, message, record| {
                // custom colorized output here
            },
        ),
        Target::new(TargetKind::LogDir { file_name: None })
        .format(move |out, message, record| {
            // custom non-colorized output here
        }),
    ])
    .build(),
```

### DIFF
--- a/.changes/allow-log-formatter-per-target.md
+++ b/.changes/allow-log-formatter-per-target.md
@@ -1,0 +1,5 @@
+---
+"log": "patch:enhance"
+---
+
+Allow specifying a log formatter per target using the `format` method on `Target`.
--- a/.changes/allow-log-formatter-per-target.md
+++ b/.changes/allow-log-formatter-per-target.md
@@ -1,6 +1,6 @@
 ---
-"log": "patch:enhance"
-"log-js": "patch"
+"log": "minor"
+"log-js": "minor"
 ---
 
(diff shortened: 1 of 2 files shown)


.changes/allow-log-formatter-per-target.md
```diff
@@ -0,0 +1,6 @@
+---
+"log": "patch:enhance"
+"log-js": "patch"
```
FabianLars:
```suggestion
"log": "minor"
"log-js": "minor"
```

Thanks for the PR! I think this counts as a minor since it's a new feature/api (can't push to your branch myself). The code itself looks good :)

FabianLars (approved):
thanks again

PR #3071, opened 2025-10-27
Android keystore biometric encrypt decrypt

bprucha:
I wanted to encrypt and decrypt a database PIN to add biometric login support for the [OPNManager](https://github.com/bprucha/OPNManager) Android app. I couldn't find any existing plugin to support this so I updated the biometric plugin. I found a year old discussion on Discord regarding similar functionality but for iOS. I commented on it suggesting we could combine our efforts to support both but haven't had any responses. Submitting this PR with support only for Android.


FabianLars:
> I found a year old discussion on Discord regarding similar functionality but for iOS. I commented on it suggesting we could combine our efforts to support both but haven't had any responses.

can you give me a link to that or ping me there?

bprucha:
> can you give me a link to that or ping me there?

Sure: [Native Biometrics](https://discord.com/channels/616186924390023171/1225154115311632586)

FabianLars:
thanks for the link and thanks for the PR!

there may be some overlap with https://github.com/tauri-apps/plugins-workspace/pull/2900 here that may need a bit more thinking. The linked thread is also very much about the keychain as well.
~~Maybe we can take some inspiration from the capawesome capacitor plugins https://capawesome.io/blog/how-to-securely-store-credentials-with-capacitor/ ~~ Nevermind, that doesn't really make sense for what we're trying to achieve here.

Then there's also the iOS keychain (for secure-storage) where you can configure it to request biometric authentication itself which obviously is the best approach to combine biometrics with secure _storage_ (no encrypt/decrypt api though).

Having a single security stuff plugin (for keychain, encryption, biometrics, and a mix of all) sounds a bit awful as well so maybe we'll need a bit of overlap between plugins (for example, both biometrics and secure-storage being able to encrypt/decrypt), but maybe i'm just overthinking it right now because the discord thread focused so much on keychain storage.

Sorry for this mess of a comment, i'll think a bit more about it the next couple of days and try to write something more cohesive.

plugins/biometric/guest-js/index.ts
```diff
@@ -64,13 +95,13 @@ export async function checkStatus(): Promise<Status> {
...
  * @param options
- * @returns
+ * @returns a promise resolving to an object containing the encrypted or decrypted data if any
  */
 export async function authenticate(
   reason: string,
   options?: AuthOptions
-): Promise<void> {
-  await invoke('plugin:biometric|authenticate', {
+): Promise<CipherData> {
```
FabianLars, line 103: i think this counts as a breaking change so maybe we'll need 2 different functions instead. One function for each AuthMode may be easier to understand anyway if that's possible.

bprucha: I noticed that too as I was reviewing at the changes I had made. I was think I should have made it Promise<CipherData?> instead. I wouldn't consider it a breaking change though because any attempt to use a return value based off the original API would be a compile time error. Using the new API that code would be 100% compatible. I can't think of any scenario where it wouldn't be compatible. Even if older code was plain old JS and somehow relied on a return value, that value would be treated as null which is the default behavior I had in mind when no MODE option is provided.

bprucha:
> Having a single security stuff plugin (for keychain, encryption, biometrics, and a mix of all) sounds a bit awful as well so maybe we'll need a bit of overlap between plugins (for example, both biometrics and secure-storage being able to encrypt/decrypt), but maybe i'm just overthinking it right now because the discord thread focused so much on keychain storage.

Yes, completely understand where you're going with this. I'm new to Tauri and didn't see an obvious way to tell a secure-storage type plugin how to hook into a biometric plugin. This PR just got the job done for me. I also noticed some discussion on Discord around ways to define plugin dependencies. Understood if this effort requires further framework build out to make it more robust.

PR #3073, opened 2025-10-28
feat(updater): add no_proxy config to disable system proxy

miaoshengkun:
Add a `no_proxy` configuration option to the updater plugin that calls reqwest's `no_proxy()` method. This allows users to prevent update requests from being affected by system proxy settings.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


FabianLars:
Thanks for the PR! would you mind also adding a small changefile (check the .changes dir for examples) as well?

miaoshengkun:
> Thanks for the PR! would you mind also adding a small changefile (check the .changes dir for examples) as well?

I have modified the commit.

plugins/updater/src/updater.rs
```diff
@@ -242,6 +244,11 @@ impl UpdaterBuilder {
         self
     }
 
+    pub fn no_proxy(mut self) -> Self {
```
FabianLars, line 248:
```suggestion
    /// Clear all proxies. See [`reqwest::ClientBuilder::no_proxy`](https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.no_proxy).
    pub fn no_proxy(mut self) -> Self {
```

FabianLars (approved):

PR #3097, opened 2025-11-11
chore: fix upload test on macos ci

FabianLars:
it works locally so idk what's happening here


FabianLars:
that was unnecessarily hard to find for me. also doesn't make sense that this is **only** a problem on macOS and only in CI(?) ...

Issue #2376, opened 2025-01-30
Upgrade `tauri-plugin-dialog` to `gtk4-rs` (potentially updating the upstream crate `rfd`)


tweidinger:
We need to upgrade from [`gtk`](https://crates.io/crates/gtk) to [`gtk4-rs`](https://crates.io/crates/gtk4) to migrate a unsoundness [issue](https://github.com/advisories/GHSA-wrw7-89jp-8q8g) found in glib < 0.2.
Please claim this issue if you want to support or work on this and keep in mind this is a prioritized issue and we try to migrate as soon as possible. If the issue is already assigned to someone you can either reach out with comments or questions reach out for collaboration to the assigned person.

This transition includes:

- [ ]  Upgrading [`gtk`](https://crates.io/crates/gtk) to [`gtk4-rs`](https://crates.io/crates/gtk4) and handling all errors caused by the dependency change
- [ ]  Handle all changes in a backwards compatible way if possible
    - [ ]  Raise awareness to the working group if breaking changes are required
- [ ]  Document migration errors that may be issues in the future or which were complicated to handle
- [ ]  Reach out to the working group (alternatively @tweidinger first) if blockers or upstream issues are hit

A possible alternative would be using the xdg feature instead, at least to unblock us and focus on the other upgrades. We'll lose the message dialogs on Linux but imo that can be a fair tradeoff. Further discussion on this should happen if no one claims and implements the transition including potential work on `rfd`.

Lepidopteran:
Hello,

I've come here to say that the status of the `rfd` documentation have been outdated for a bit and I've just help updated them PolyMeilex/rfd#272.

The `xdg_desktop` backend can actually do message dialogs using [`zenity`](https://gitlab.gnome.org/GNOME/zenity).

Additionally, there is an open issue for upgrading `rfd` to [`gtk4-rs`](https://crates.io/crates/gtk4) PolyMeilex/rfd#227 but it seems the [`gtk` backend isn't really recommended](https://github.com/PolyMeilex/rfd/issues/227#issuecomment-2560293223) and described as a [hacky feature](https://github.com/PolyMeilex/rfd/issues/227#issuecomment-2568542213).

I could see if I could try to create a backend for `gtk4` if it really is required, but I can't promise anything.

Personally, I'd prefer using `xdg_destkop` backend as that allows the user to use their file manager of their desktop environment, but I can see the case where a user doesn't have any desktop portal installed, though I'm pretty that would be rare case.

[...]

PR #3098, opened 2025-11-11
feat(dialog): add `xdg-portal` as an optional feature for `rfd`

Lepidopteran:
Adds a feature to be able to use `xdg-portal` instead of `gtk3` for `tauri-plugin-dialog`
Let me know if any changes are needed!

Continues #2376


FabianLars:
that was quick, thanks! For the plugin i don't think there's much more to do except for a small [changefile](https://github.com/tauri-apps/plugins-workspace/tree/v2/.changes) and maybeee some docs that relay some of rfd's xdg info.
The major thing is the zenity dependency but that's something to worry about in the tauri repo x)

FabianLars:
oh and ci may fail because we set --all-features

Lepidopteran:
> maybeee some docs that relay some of rfd's xdg info.

I can do that, by docs are you referring to [README.md](https://github.com/tauri-apps/plugins-workspace/blob/v2/plugins/dialog/README.md) in the dialog plugin folder or is it something else?

Lepidopteran:
If you don't mind me asking, Is there anything else I need to do @FabianLars?

I've added the `changefile` to `.changes` like you requested, but I'm not sure where to write the documentation about rfd's xdg info.

Should I be writing it in `dialog-plugin`'s README.md?

Additionally, when relaying some of the rfd's docs, am I explaining that `zenity` and an XDG portal needs to be installed on the users machine?

FabianLars:
I didn't have time for your pr yet (cause I wanted to try it but we can discard this)
We can leave it as is or ideally add just a few lines to the readme. Mentioning zenity and the portal would be nice. Anything more extensive should be on the website.

Lepidopteran:
I added a section about XDG desktop support to the README, lemme know if any changes need to be made!

FabianLars (dismissed):
thanks!

FabianLars:
good thing i waited for CI - mutually exclusive features need special handling there (see http) - it's too late for me now so if we can't figure it out i'll do it tomorrow :)

FabianLars (approved):

PR #3099, opened 2025-11-12
docs(updater): installer args and `on_before_exit`

Legend-Master:


plugins/updater/src/updater.rs
```diff
@@ -264,11 +266,17 @@ impl UpdaterBuilder {
...
+    ///
+    /// Note: this only removes the additional arguments added through
+    /// [`Self::installer_arg`], [`crate::Builder::installer_arg`]
+    /// and the `plugins > updater > windows > installerArgs` config
     pub fn clear_installer_args(mut self) -> Self {
         self.installer_args.clear();
         self
     }
 
+    /// Function to run before we run the installer and exist the app through `std::process::exit(0)` on Windows
```
```suggestion
    /// Function to run before we run the installer and exit the app through `std::process::exit(0)` on Windows
```

Legend-Master: oops 😂

plugins/updater/src/lib.rs
```diff
@@ -165,6 +167,9 @@ impl Builder {
         self
     }
 
+    /// Remove all the additional arguments to pass to the Windows installer.
+    ///
+    /// Note: this only removes the additional arguments added through [`Self::installer_args`]
```
FabianLars: should this be a bit more explicit and say that there are internal args?

FabianLars (dismissed):

FabianLars (approved):

PR #3101, opened 2025-11-12
Resolve boolean in `isAvailable` to agree with TypeScript API

### LINKED ISSUE #2830: wrong ts signature for NFC isAvailable return type
looking inside the dist.js of the plugin I can see the following signature:
`export declare function isAvailable(): Promise<boolean>;`

but when I console.log the actual value returned I get `{available: true}`

There is a mismatch here

bclarke123:
Description: Updates native plugins for Android / iOS to resolve a simple boolean in the `isAvailable()` functions of the NFC plugin

Fixes: #2830 

Tested: iOS / Android simulators; iPhone 8, Pixel 10


FabianLars:
Thanks for the PR! could you also add a small changefile please? (check .changes dir for examples)

FabianLars:
> Do it the other way to maintain expected functionality in Rust code

i mean, the rust code could be updated as well, i think it should support this as well no?

bclarke123:
> > Do it the other way to maintain expected functionality in Rust code
> 
> i mean, the rust code could be updated as well, i think it should support this as well no?

It absolutely can, after re-reading the original issue where it was suggested to update the TypeScript, I looked at that approach and it made for fewer changes, and only impacts TS usage. The original approach would have affected every `isAvailable` call in every downstream project.

FabianLars (approved):

PR #3110, opened 2025-11-19
fix(log): custom writer to rotate log file when exceeds max_file_size (fix #707)

### LINKED ISSUE #707: [log] LogTarget::Folder max_file_size does not affect
```
builder.plugin(
                tauri_plugin_log::Builder::default()
                    .max_file_size(100)
                    .targets([LogTarget::Folder(&path)])
                    .rotation_strategy(RotationStrategy::KeepAll)
                    .level(LevelFilter::Debug)
                    .build(),
            );
```

the log file becomes **~1500 KB** but still write there - no `*.bac` file created

liuzhch1:
Fix #707

### DIFF
--- a/plugins/log/src/lib.rs
+++ b/plugins/log/src/lib.rs
@@ -296,13 +296,11 @@ impl Write for RotatingFile {
             return Ok(());
         }
         if self.inner.is_none() {
-            self.open_file()
-                .map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, e))?;
+            self.open_file().map_err(std::io::Error::other)?;
         }
 
         if self.current_size + (self.buffer.len() as u128) > self.max_size {
-            self.rotate()
-                .map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, e))?;
+            self.rotate().map_err(std::io::Error::other)?;
         }
 
         if let Some(file) = self.inner.as_mut() {
--- a/plugins/log/src/lib.rs
+++ b/plugins/log/src/lib.rs
@@ -299,7 +299,8 @@ impl Write for RotatingFile {
             self.open_file().map_err(std::io::Error::other)?;
         }
 
-        if self.current_size + (self.buffer.len() as u64) > self.max_size {
+        if self.current_size != 0 && self.current_size + (self.buffer.len() as u64) > self.max_size
+        {
             self.rotate().map_err(std::io::Error::other)?;
         }
 
--- a/plugins/log/src/lib.rs
+++ b/plugins/log/src/lib.rs
@@ -14,6 +14,8 @@ use log::{LevelFilter, Record};
 use serde::Serialize;
 use serde_repr::{Deserialize_repr, Serialize_repr};
 use std::borrow::Cow;
+use std::fs::OpenOptions;
+use std::io::Write;
 use std::{
     fmt::Arguments,
     fs::{self, File},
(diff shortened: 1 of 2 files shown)


liuzhch1:
I have another related question: file's `len(&self) -> u64` returns u64, but the log module uses u128 throughout, which seems unnecessary.
Should we switch to u64?

FabianLars:
I'd still love to see this solved upstream https://github.com/daboross/fern/issues/80 but considering the activity of the maintainer we'd need something like your PR in the meantime anyway :)

> I have another related question: file's len(&self) -> u64 returns u64, but the log module uses u128 throughout, which seems unnecessary.

If this isn't part of the public api you can change it :)

plugins/log/src/lib.rs
```diff
@@ -142,6 +146,175 @@ impl TimezoneStrategy {
...
+        }
+        if self.inner.is_none() {
+            self.open_file()
+                .map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, e))?;
+        }
+
+        if self.current_size + (self.buffer.len() as u128) > self.max_size {
+            self.rotate()
+                .map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, e))?;
+        }
```
FabianLars: If the buffer length is bigger than max_size this will endlessly rotate. Maybe max_size should be a soft limit so that we only rotate after writing something past the max_size? could be bound to a comparison of buffer.lent() and max_size so that this really only happens in those cases.

liuzhch1:
Wow, thanks for your quick feedback.
I didn't aware it's a feature requested in upstream.

`pub fn max_file_size(mut self, max_file_size: u128) -> Self {`
I guess this is considered as a public api?...

FabianLars:
Yeah it is. But considering file.len() is u64 i think we can add a doc comment saying that it's cut off at u64::max and change the argument to u64 in v3

plugins/log/src/lib.rs
liuzhch1: The rotate is called at most one time for one logging call.
The current logic only results in the first log file potentially being rotated before any content is written to it; subsequent log files are written normally.
Perhaps we could check the log file size here: if self.current_size is 0, then we skip rotating check?
```suggestion
        if self.current_size != 0 && self.current_size + (self.buffer.len() as u128) > self.max_size {
            self.rotate()
                .map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, e))?;
        }
```

liuzhch1:
> Yeah it is. But considering file.len() is u64 i think we can add a doc comment saying that it's cut off at u64::max and change the argument to u64 in v3

I clamped it to u64::MAX, and use `self.max_file_size as u64` when it's used, which should be logically safe. If you prefer, I could change to use a min instead.

plugins/log/src/lib.rs
FabianLars: righttttt my bad. since log lines larger than max_size should not realistically happen the current code logic is fine but i wouldn't mind the 0 check so we really never end up with empty files - your call though.

FabianLars:
seems to work well, thanks!

one thing a low max_size made me realize though is that we need to think about the .bak files, my Logs folder is more than full rn x)

FabianLars:
ah and to be able to merge your PR i'll need a small changefile, check the .changes dir for examples. thanks :)

liuzhch1:
> one thing a low max_size made me realize though is that we need to think about the .bak files, my Logs folder is more than full rn x)

Yeah I've also noticed this issue; if `max_file_size` is too small, it generates many .bak files. I haven't figured out how to handle this problem yet. And IMO such case should rarely occur during actual use... As previous author said 'highly unlikely but defensively...' lol

FabianLars:
yeah whatever 🤷 better have many bk files than lost files

FabianLars (approved):

PR #3114, opened 2025-11-20
fix(updater): make pubkey optional in tauri config, check if set at runtime (fixes #2438)

### LINKED ISSUE #2438: [updater] Config fields required even if set at runtime
For [`updater`](https://tauri.app/plugin/updater/) it is possible to set all the required config fields at runtime: [endpoints](https://tauri.app/plugin/updater/#endpoints), [public key](https://tauri.app/plugin/updater/#public-key). However, it is still required that the `{ "plugins": { "updater": { "pubkey" } } }` field exists in the conf file.

You can get around this by setting the field to any value (e.g. `{ "plugins": { "updater": { "pubkey": "<SET AT RUNTIME>" } } }`), but it seems this should be unnecessary.

wovw:
If pubkey isn't set in tauri config, there's a new error to make sure it's set at least at runtime. 
Currently the pubkey set at runtime overwrites the one in config if it's set, shown with a warning.
Let me know if I need to make any edits anywhere.


plugins/updater/src/lib.rs
```diff
@@ -215,9 +215,16 @@ impl Builder {
...
             .setup(move |app, api| {
                 let mut config = api.config().clone();
-                if let Some(pubkey) = pubkey {
+
+                if config.pubkey.is_none() && pubkey.is_none() {
+                    return Err(Box::new(Error::MissingPubKey));
+                } else if pubkey.is_some() {
+                    log::warn!(
+                        "Updater pubkey is set at runtime. It will overwrite the config pubkey if set."
+                    );
```
FabianLars: i don't think there's much value in that log here, at least not on warn level. Maybe debug level but i really wouldn't mind removing it.

Instead we should think about having a warning at compile time though that likely has to happen in tauri's cli.
Speaking of which, we'd also need to change the cli https://github.com/tauri-apps/tauri/blob/afdd288eab7d8473fd6ad2796571a24c4873a8db/crates/tauri-cli/src/interface/rust.rs#L783

wovw: I nested the warning log further, it could serve as an incentive for the developer to keep it only in one or the other, but we could also just make it a debug log.

wovw:
I took at look at making changes in tauri-cli. From my understanding, it looks like in https://github.com/tauri-apps/tauri/blob/afdd288eab7d8473fd6ad2796571a24c4873a8db/crates/tauri-cli/src/bundle.rs#L223 the pubkey is validated against the private key when bundling. We could make this check optional if the pubkey doesn't exist in the config + show a warning. Should I go ahead with this?

PR #3115, opened 2025-11-20
feat: make `ExitStatus` code field public

niusia-ua:
All members of the `Output` struct are public. However, it is not possible to create its instance in the user code, since the code field of the `ExitStatus` struct is private.

### DIFF
--- a/plugins/shell/src/process/mod.rs
+++ b/plugins/shell/src/process/mod.rs
@@ -89,7 +89,7 @@ impl CommandChild {
 /// Describes the result of a process after it has terminated.
 #[derive(Debug)]
 pub struct ExitStatus {
-    code: Option<i32>,
+    pub code: Option<i32>,
 }
 
 impl ExitStatus {


FabianLars:
> However, it is not possible to create its instance in the user code

This sounds more like a feature to me as it would allow us to add more private fields. Just mentioning it because we've had issues with _not_ being able to do that the last few weeks.

Since we're moving towards v3 though that'd not be much of an issue i guess.

Anyway, thanks for the PR! To be able to merge it i'll need a small changefile as well (check the .changes folder)

niusia-ua:
> This sounds more like a feature to me as it would allow us to add more private fields.

Well, then it is better either make the `Output`'s fields private too (and replace direct access by getters) or add a `new` method for instantiating `Output` and `ExitStatus` with known fields.

If you really want to prevent manual instantiating of the `Output`, we should leave an explicit note about it.

FabianLars:
I don't disagree but one thing i forgot is that we're also mirroring rust std here: https://doc.rust-lang.org/std/process/struct.Output.html / https://doc.rust-lang.org/std/process/struct.ExitStatus.html - that's reason enough for me to have to decline this PR as-is, sorry :/

> If you really want to prevent manual instantiating of the Output, we should leave an explicit note about it.

fair enough!

niusia-ua:
Okay, no problem. You can close this PR. I will open another with a clarification comment instead.

PR #3121, opened 2025-11-22
fix(barcode-scanner): remove unnecessary permission checks on Android (fix #2312)

### LINKED ISSUE #2312: [BARCODE SCANNER] [BUG] Doesn't ask again when permission is set to ask every time on android.
When user choose "ask every time" on android device, then prompt to set permission doesn't pop up again, so he must open app settings. This very bad behavior, very bad user experience, should be changed.

I saw in android the source code it's made on purpose. It should be changed.

NKIPSC:
As reported in #2312, the barcode-scanner plugin currently checks if a permission request has already been made and never requests permissions a second time. This breaks Android's "ask every time" setting.

Additionally, the plugin is currently calling `shouldShowRequestPermissionRationale` to determine whether or not to request permissions again. The actual purpose of this function, however, is to display an in-app message to the user explaining why the permission is required ([documentation](https://developer.android.com/training/permissions/requesting#explain)).

I've removed both of these checks to make the plugin behave as expected according to user preference.

### DIFF
--- a/.changes/fix-android-barcode-scanner-permission-check.md
+++ b/.changes/fix-android-barcode-scanner-permission-check.md
@@ -1,0 +1,6 @@
+---
+"barcode-scanner": patch
+"barcode-scanner-js": patch
+---
+
+Remove unnecessary checks on Android when requesting camera permission.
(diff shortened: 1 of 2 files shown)


NKIPSC:
There was actually another instance of `shouldShowRequestPermissionRationale` used improperly, which would report permission as granted when it was in fact denied.
I've updated the PR to remove that one as well.

FabianLars (approved):
thank you :)

PR #3125, opened 2025-11-26
fix(sql): bindTypes hint to correctly bind binary data (fix #105)

### LINKED ISSUE #105: [sql] Can't save binary data with `execute` (SQLite)
First of all, thanks for your work on Tauri and essential plugins like this one!

I'm having trouble inserting `Uint8Array` data into `BLOB` columns in SQLite. Instead of being inserted as binary data, it's inserted as a stringified JSON object.

My main use case is a `uuid` column storing the 128-bit binary representation of a UUID.

```sql
CREATE TABLE node (
    uuid BLOB PRIMARY KEY
    -- ...
)
```

```typescript
const uuid = new Uint8Array(16)  // this would be a generated 128-bit UUID
const result = await db.execute('INSERT INTO node (uuid) VALUES ($1)', [uuid])
```

What's inserted into `node.uuid`:

```js
'{"0":0,"1":0,"10":0,"11":0,"12":0,"13":0,"14":0,"15":0,"2":0,"3":0,"4":0,"5":0,"6":0,"7":0,"8":0,"9":0}'
```

If I instead save the `ArrayBuffer` itself, I get an empty object:

```typescript
const uuid = new Uint8Array(16).buffer  // its underlying ArrayBuffer
const result = await db.execute('INSERT INTO node (uuid) VALUES ($1)', [uuid])
```

```js
'{}'
```

Am I doing something wrong, or does the binary data simply not survive serialization when sent from JS to Rust?

If it's the latter, maybe [serde_binary](https://docs.rs/serde-binary/latest/serde_binary/) can be of help?

liuzhch1:
fix #105

WIP
Currently only implemented the sqlite part. I'd add mysql and postgres parts once we decided to use this approach(just copy paste I think).


liuzhch1:
> We don't have to convert it to uint8array - since number[] is the most native type for this we might as well return that as-is and let the user decide whether they want uint8array - we do the same in tauri's invoke for example. (from [#105](https://github.com/tauri-apps/plugins-workspace/issues/105#issuecomment-3574767624))

I think we should convert it to uint8array. Consider if we change to other effective way to transmit data as binary(eg msg pack) in future, the BLOB data directly come to JS as uint8array. To keep api the same(return number[]), we need to convert Array.from(uint8array) then give it to user. However user might do need uint8array format since it is a blob, so they need to convert it back Uint8Array(number[]). Those converts do take time and memory

... just some thoughts comes to my mind, maybe it's overkill for now. api breaking change might be accepted in future version or additional api can be added..  and user can invoke `invoke<T>('plugin:sql|select'` themself, right? haha

FabianLars:
> Those converts do take time and memory

which was my argument to _not_ do these conversations on our side but i don't care too much since number[] -> uint8array should be relatively cheap.

FabianLars:
p.s. i'm a bit busy today so will take a look tomorrow, sry

liuzhch1:
yeah sure, no worries

let's use number[] for now.

PR #3129, opened 2025-11-28
fix(dialog): rfd feature flags

FabianLars:
wayland was missed in #3128 and tokio was always incorrect since it's not used for the gtk backend


Legend-Master (approved):
Oops, missed this one, I have looked up the rfd change log and seeing there's no `ashpd` in the diff in the last PR and assumed that we didn't use it at all 😂

PR #3136, opened 2025-12-01
feat(dialog) - Support fileAccessMode for open dialog (#3030)

onehumandev:
On iOS, when trying to access a file that exists outside of the app sandbox, one of 2 things need to happen to be able to perform any operations on said file:

* A copy of the file needs to be made to the internal app sandbox
* The method startAccessingSecurityScopedResource needs to be called.

Previously, a copy of the file was always being made when a file was selected through the picker dialog.

While this did ensure there were no file access exceptions when reading from the file, it does not scale well for large files.

To resolve this, we now support `fileAccessMode`, which allows a file handle to be returned without copying the file to the app sandbox.

This MR only supports this change for iOS; MacOS has a different set of needs for security scoped resources.

See discussion in #3716 for more discussion of the difference between iOS and MacOS.
See MR #3185 to see how these scoped files will be accessible using security scoping.


onehumandev:
c. @velocitysystems Here is the MR for the changes to support `startAccessingSecurityScopedResource ` for working with Cloud providers without requiring making a local copy of the file. Thanks!

plugins/dialog/ios/Sources/DialogPlugin.swift
```diff
@@ -52,11 +61,27 @@ class DialogPlugin: Plugin {
 
   var filePickerController: FilePickerController!
   var onFilePickerResult: ((FilePickerEvent) -> Void)? = nil
+  
+  // Dictionary to keep strong references to URLs with active security-scoped access
+  // Key: resource ID (UUID string), Value: URL object
+  private var activeSecurityScopedURLs: [String: URL] = [:]
```
velocitysystems: Possible race condition: `activeSecurityScopedURLs` dictionary is accessed from multiple threads (main thread via invoke handlers, potentially background threads) without synchronization.

Can we make this thread-safe?

plugins/dialog/ios/Sources/DialogPlugin.swift
```diff
@@ -292,6 +319,44 @@ class DialogPlugin: Plugin {
...
+    
+    // Look up the URL by resource ID and stop accessing it
+    if let url = activeSecurityScopedURLs[args.resourceId] {
+      url.stopAccessingSecurityScopedResource()
+      activeSecurityScopedURLs.removeValue(forKey: args.resourceId)
+      invoke.resolve(["success": true])
+    } else {
+      // Resource ID not found - might have already been cleaned up or never existed
+      // This is not necessarily an error, so we'll return success
+      invoke.resolve(["success": true])
```
velocitysystems: This _could_ mask bugs where developers pass incorrect resource IDs. Should we at least log a warning?

plugins/dialog/src/commands.rs
```diff
@@ -359,3 +367,80 @@ pub(crate) async fn confirm<R: Runtime>(
...
+
+#[derive(Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct EndFileAccessOptions {
+    pub resource_id: String,
+}
+
+#[command]
+pub(crate) async fn end_file_access<R: Runtime>(
+    _window: Window<R>,
```
velocitysystems: `_window` is unused here.

plugins/dialog/ios/Sources/DialogPlugin.swift
```diff
@@ -292,6 +319,44 @@ class DialogPlugin: Plugin {
...
+      activeSecurityScopedURLs[resourceId] = url
+      invoke.resolve(["resourceId": resourceId])
+    } else {
+      // Failed to start accessing - this might happen if the URL is not security-scoped
+      // or if access was already started. Still return success but without resource ID
+      invoke.reject("Failed to start accessing")
+    }
+  }
+
+  @objc public func endFileAccess(_ invoke: Invoke) throws {
```
velocitysystems: iOS may revoke security-scoped access when the app backgrounds. The current implementation doesn't handle this scenario. Should document this limitation and/or handle it?

velocitysystems (changes requested):

plugins/dialog/src/commands.rs
onehumandev: My understanding is that the window parameter is required for `command`s to be invoked; is that incorrect?

plugins/dialog/ios/Sources/DialogPlugin.swift
onehumandev: Good call; updated the method to return a distinct value stating that no matching resource was found with the given id, and to correctly error if anything goes wrong.

plugins/dialog/src/mobile.rs
```diff
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-License-Identifier: MIT
 
-use serde::{de::DeserializeOwned, Deserialize};
+use serde::{de::DeserializeOwned, Deserialize, Serialize};
```
velocitysystems: Nit: `Serialize` is added but not used.

plugins/dialog/guest-js/index.ts
```diff
@@ -76,6 +76,49 @@ interface OpenDialogOptions {
...
+   * ```ts
+   * const path = await open({ fileAccessMode: 'scoped' });
+   *
+   * startFileAccess({ path })
+   *   .then(() => {
+   *     // perform file operations
+   *   })
+   *   .catch(onMessage);
+   *
+   * endFileAccess({ path })
```
velocitysystems: Nit: `endFileAccess` takes `{ resourceId}` not `{ path }`.

plugins/dialog/ios/Sources/DialogPlugin.swift
```diff
@@ -292,6 +350,65 @@ class DialogPlugin: Plugin {
...
+        return nil
+      }
+    }
+    
+    if let resourceId = resourceId {
+      invoke.resolve(["resourceId": resourceId])
+    } else {
+      // Failed to start accessing - this might happen if the URL is not security-scoped
+      // or if access was already started, or if access was revoked (e.g., after app backgrounding)
+      invoke.reject("Failed to start accessing")
```
velocitysystems: Nit: The error message is generic. Consider distinguishing between:
- URL is not security-scoped (e.g., already in sandbox)
- Access was revoked
- Invalid URL format

velocitysystems (changes requested):

plugins/dialog/src/commands.rs
velocitysystems: According to Tauri's documentation, `Window`, `AppHandle`, and `State<T>` are "Special Parameters" that are automatically injected by the `#[tauri::command]` macro if present, but they're not required.

Since `end_file_access` doesn't use the window for anything (no scope management, no parent window needed), it can safely be removed.

onehumandev: When I remove the initial argument, we get a compilation error related to where we register the handler:
```ts
plugins/dialog/src/commands.rs:427:1
    |
427 |   #[command]
    |   ^^^^^^^^^^ cannot infer type
    |
   ::: plugins/dialog/src/lib.rs:203:25
    |
203 |           .invoke_handler(tauri::generate_handler![
    |  _________________________-
204 | |             commands::open,
205 | |             commands::save,
206 | |             commands::message,
...   |
210 | |             commands::end_file_access,
211 | |         ])
    | |_________- in this macro invocation
    |
    = note: cannot satisfy `_: tauri::Runtime`
note: required by a bound in `end_file_access`
   --> plugins/dialog/src/commands.rs:428:40
    |
428 | pub(crate) async fn end_file_access<R: Runtime>(
```

We can at least change the argument to an `AppHandle` instead of a `Window`, but removing the parameter will cause a compilation error.
Is there something I am missing with this?

onehumandev:
Updated MR per discussion in #3185 ; we now only include support for `fileAccessMode`.
c. @velocitysystems  @lucasfernog

lucasfernog (approved):

Issue #2039, opened 2024-11-12
[sql] Postgres UUID columns returning null 


glib-0:
Per my comment on #1375 
- PostgreSQL 16
- Fresh database with test table containing a uuid column with default `gen_random_uuid()`, added a single test row
- Frontend calls `await db.select("SELECT * FROM test")`
- Record returns fine, but uuid column value is `null`

### Tauri Info
[✔] Environment
    - OS: Windows 10.0.22631 x86_64 (X64)
    ✔ WebView2: 130.0.2849.80
    ✔ MSVC: Visual Studio Community 2022
    ✔ rustc: 1.80.0 (051478957 2024-07-21)
    ✔ cargo: 1.80.0 (376290515 2024-07-16)
    ✔ rustup: 1.27.1 (54dd3d00f 2024-04-24)
    ✔ Rust toolchain: stable-x86_64-pc-windows-msvc (default)
    - node: 22.8.0
    - npm: 10.8.2

[-] Packages
    - tauri 🦀: 2.0.6
    - tauri-build 🦀: 2.0.2
    - wry 🦀: 0.46.3
    - tao 🦀: 0.30.5
    - @tauri-apps/api : 2.1.1
    - @tauri-apps/cli : 2.1.0

[-] Plugins
    - tauri-plugin-fs 🦀: 2.0.3
    - @tauri-apps/plugin-fs : 2.0.0 (outdated, latest: 2.0.2)
    - tauri-plugin-sql 🦀: 2.0.2
    - @tauri-apps/plugin-sql : 2.0.1
    - tauri-plugin-dialog 🦀: 2.0.3
    - @tauri-apps/plugin-dialog : 2.0.1
    - tauri-plugin-shell 🦀: 2.0.2
    - @tauri-apps/plugin-shell : 2.0.1

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - devUrl: http://localhost:1420/
    - framework: React
    - bundler: Vite

zxcodes:
I'm facing the same issue as well. The current workaround I'm using is to cast it to a text type.

itsmadhusudhan:
I too have the same issue so make a PR https://github.com/tauri-apps/plugins-workspace/pull/3144

PR #3144, opened 2025-12-05
Fixes uuid columns returning null in postgres query results (#2039)

itsmadhusudhan:
When querying postgres db all the uuid columns were returning null.

This occurred because the UUID type was grouped with string types (CHAR, VARCHAR, TEXT, NAME) and attempted to decode as `String` instead of the proper `uuid::Uuid` type.

This PR adds support to return uuid columns by splitting UUID case from string types match arm.


plugins/sql/Cargo.toml
```diff
@@ -29,10 +29,11 @@ tauri = { workspace = true }
 log = { workspace = true }
 thiserror = { workspace = true }
 futures-core = "0.3"
-sqlx = { version = "0.8", features = ["json", "time"] }
+sqlx = { version = "0.8", features = ["json", "time", "uuid"] }
 time = "0.3"
 tokio = { version = "1", features = ["sync"] }
 indexmap = { version = "2", features = ["serde"] }
+uuid = "1.19.0"
```
FabianLars:
```suggestion
uuid = "1"
```

I'll try to set loose minimum versions if possible.

FabianLars (commented):
thanks for the pr! could you please also add a small changefile? (look in the .changes dir for examples)

plugins/sql/Cargo.toml
itsmadhusudhan: Done

.changes/fix-sql-query-results-uuid-returned-as-null.md
```diff
@@ -0,0 +1,5 @@
+---
+"sql": "patch"
```
FabianLars, line 2:
```suggestion
"sql": "patch"
"sql-js": "patch"
```

FabianLars (approved):

Issue #1471, opened 2024-06-16
[bug][shell][v2]: incorrect stdout encoding


thewh1teagle:
I use sidecar in v2 project to spawn `yt-dlp` from Rust with

```rust
let mut (rx, _) = app_handle.shell().sidecar(...);
spawn(async move {
  while let Some(event) = rx.recv().await {
    if let CommandEvent::Stdout(line: Vec<u8>) = event { ... } // line missing characters
  }
});
```

Example sentence:
`[download] Destination: PEDRO - Jaxomy, Agatino Romero, Raffaella Carrà (TikTok Song) [XzilCu9PcZk].f616.mp4`
Notice the special character in `Carrà` the `à` will be missing in the lines.

<details>
<summary>tauri info</summary>

```rust
[✔] Environment
    - OS: Windows 10.0.22631 X64
    ✔ WebView2: 125.0.2535.92
    ✔ MSVC: Visual Studio Community 2022
    ✔ rustc: 1.77.1 (7cf61ebde 2024-03-27)
    ✔ cargo: 1.77.1 (e52e36006 2024-03-26)
    ✔ rustup: 1.27.0 (bbb9276d2 2024-03-08)
    ✔ Rust toolchain: stable-x86_64-pc-windows-msvc (default)
    - node: 20.12.0
    - npm: 10.5.0
    - bun: 1.1.10

[-] Packages
    - tauri [RUST]: 2.0.0-beta.22
    - tauri-build [RUST]: 2.0.0-beta.17
    - wry [RUST]: 0.40.1
    - tao [RUST]: 0.28.1
    - @tauri-apps/api [NPM]: 2.0.0-beta.13
    - @tauri-apps/cli [NPM]: 2.0.0-beta.20

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - devUrl: http://localhost:1420/
    - framework: React
    - bundler: Vite
 ```
</details>

**Comment by Tunglies** [comment]
Might need use encoding_rs to decode manully
```rust
use encoding_rs::{Encoding, UTF_8};
let enc = Encoding::UTF_8;
let (text, _, had_errors) = enc.decode(line);
```
Or tauri plugin shell has re-expoprt encoding_rs, some usage like
```rust
/// commands.rs
use encoding_rs::Encoding;

/// encoding_rs
/// impl Encoding
[...]

PR #3148, opened 2025-12-05
fix(process): improve stdout encoding handling in read_line functions #1471

Tunglies:
I developed on macOS, need to make sure encoding behavior same as Windows.
Sidecar:
```rust
use encoding_rs;
use std::io::{self, Write};
use std::thread;
use std::time::Duration;

fn main() {
    let stdout = io::stdout();
    let mut handle = stdout.lock();
    let utf8_output = "[download] Destination: PEDRO - Jaxomy, Agatino Romero, Raffaella Carrà (TikTok Song) [XzilCu9PcZk].f616.mp4\n";
    const TARGET_ENCODING: &'static encoding_rs::Encoding = encoding_rs::WINDOWS_1252;
    let (encoded_bytes, _, malformed) = TARGET_ENCODING.encode(utf8_output);
    if malformed {
        eprintln!("[SIM] Warning: some characters cannot be encoded in Windows-1252!");
    }
    loop {
        match handle.write_all(encoded_bytes.as_ref()) {
            Ok(_) => {
                eprintln!(
                    "[SIM] Simulated output: successfully wrote {} bytes.",
                    encoded_bytes.len()
                );
            }
            Err(e) => {
                eprintln!("[SIM] Error writing to stdout: {}", e);
                break;
            }
        }
        thread::sleep(Duration::from_millis(1000));
    }
    let check_a_grave = TARGET_ENCODING.encode("à").0;
    eprintln!(
        "[SIM] Check: 'à' is encoded in Windows-1252 as bytes {:?}",
        check_a_grave
    );
}
```

User side, code snippet
```rust
let command = app_handle
            .clone()
            .shell()
            .sidecar(EXTERNAL_BIN)
            .expect("bin");
        let (mut rx, _) = command.spawn().expect("spawn");

        tauri::async_runtime::spawn(async move {
            while let Some(event) = rx.recv().await {
                if let CommandEvent::Stdout(line) = event {
                    let output = match String::from_utf8(line) {
                        Ok(s) => s,
                        Err(e) => {
                            println!("Invalid UTF-8 sequence: {}", e);
                            String::from_utf8_lossy(&e.into_bytes()).into_owned()
                        },
                    };
                    println!("{}", output);
                }
            }
        });
```

Output with mainline
```
[download] Destination: PEDRO - Jaxomy, Agatino Romero, Raffaella Carr� (TikTok Song) [XzilCu9PcZk].f616.mp4
```

Output with PR
```
[download] Destination: PEDRO - Jaxomy, Agatino Romero, Raffaella Carrà (TikTok Song) [XzilCu9PcZk].f616.mp4
```

### DIFF
--- a/.changes/fix-stdout-readline-encoding.md
+++ b/.changes/fix-stdout-readline-encoding.md
@@ -1,0 +1,6 @@
+---
+"shell": patch
+"shell-js": patch
+---
+
+Fix incorrect stdout encoding when readline.
--- a/plugins/shell/src/process/mod.rs
+++ b/plugins/shell/src/process/mod.rs
@@ -401,6 +401,7 @@ fn read_raw_bytes<F: Fn(Vec<u8>) -> CommandEvent + Send + Copy + 'static>(
     tx: Sender<CommandEvent>,
     wrapper: F,
 ) {
+    let encoder = get_system_encoding();
     loop {
         let result = reader.fill_buf();
         match result {
@@ -409,8 +410,10 @@ fn read_raw_bytes<F: Fn(Vec<u8>) -> CommandEvent + Send + Copy + 'static>(
                 if length == 0 {
                     break;
                 }
+                let (cow, _, _) = encoder.decode(buf);
+                let decode_bytes = cow.into_owned().into_bytes();
                 let tx_ = tx.clone();
-                let _ = block_on_task(async move { tx_.send(wrapper(buf.to_vec())).await });
+                let _ = block_on_task(async move { tx_.send(wrapper(decode_bytes)).await });
                 reader.consume(length);
             }
             Err(e) => {
@@ -428,6 +431,7 @@ fn read_line<F: Fn(Vec<u8>) -> CommandEvent + Send + Copy + 'static>(
     tx: Sender<CommandEvent>,
     wrapper: F,
 ) {
+    let encoder = get_system_encoding();
     loop {
         let mut buf = Vec::new();
         match tauri::utils::io::read_line(&mut reader, &mut buf) {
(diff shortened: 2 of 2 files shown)


Tunglies:
CC @FabianLars, this is initial start. Potential to continue improve with your advice.

> there's something for us to fix https://github.com/tauri-apps/plugins-workspace/issues/1471#issuecomment-3612072924

If auto encoding is required, current hardcoded for encoding need change to system encoding page detect, and brings more cross-platform support.

FabianLars:
@Legend-Master Do you know if this could also cause issues? This feels like it depends on the spawned process eg iirc node also uses utf8 on all platforms by default 🤔

Legend-Master:
Since this is the raw output of the command in `Vec<u8>`, I don't think we should alter/preprocess it here to be honest

And for the encoding, I think it really depends on the command/program you spawned, it's not something we could assume (by the way, windows-1252 code page / encoding isn't the one used in all applications, it's only for apps using ANSI with `ANSI Latin 1; Western European (Windows)`)

> https://learn.microsoft.com/en-us/windows/win32/intl/code-page-identifiers

Tunglies:
> I found re-export usage use tauri_plugin_shell::process::Encoding;, use that we can do
 ```rust
 let decoder = encoding_rs::WINDOWS_1252;
let (cow, _, had_errors) = recoder.decode(&line);
let output: String = cow.into_owned();
```

> it's not something we could assume

Shall we leave this and documents on the usage of re-export Encoding? #1471 isn't a bug, user should consider encoding when use.

**Comment by Legend-Master** [comment]
Our JavaScript API defaulting to assume UTF-8 which I think is more or less in line with the Rust behavior (since JS one defaults to `string` while the Rust one is `Vec<u8>`), I think maybe we mirror the `SpawnOptions.encoding` option from JS? (and of course documentations)

> https://github.com/tauri-apps/plugins-workspace/issues/1471 isn't a bug, user should consider encoding when use.

Yeah, the program can output whatever they want technically, we can only provide sensible defaults that works for most

**Comment by Tunglies** [comment]
```rust
sturcture Command {
 encoding: Encoding
 ...
}
app_handle.sidecar(BIN).output_line_encoding(Encoding::Type)
```
Some thing like this? The `.spawn` only returns rx and child. Additional field of Command structure needed. And this should only takes effect on not-raw output.

Legend-Master:
Ah, I see, let's just do the documentation work for now then

Tunglies:
The user should considering encoding on their own. Shold I close this PR or leave as reference?

Legend-Master:
We could close it or turn it into a PR that updates the documentation or if you want to open a new one that's fine too

Legend-Master:
Documentation added in #3183, thanks

Issue #1332, opened 2024-05-16
[shell] Add option to spawn command in process group


FabianLars:
this is basically the fix to issues like pyinstaller apps not being killed because pyinstaller wraps the actual app in a thin parent process that doesn't clean up after itself.

v1's command code was a bit too much for me to find motivation tbh but in this plugin it should be fairly straight forward.

Since may be use cases where this is not desirable, it will be a simple boolean option.

The implementation should use https://github.com/watchexec/process-wrap, the successor of command-group.

p.s. just opening this issue so i don't forget about it until next week, not asking the team to work on this (however, if a community member wants to work on this then i'd be quite happy about it)

mitulagr2:
Hi, is this in progress?

FabianLars:
No. I started it a few weeks/months ago but it was more annoying to implement than expected. Not even sure if I still have the local branch, gonna check on Monday.

If you're asking cause you want to pick it up then go for it :)

elibroftw:
Is there any API stub? I don't need this feature anymore since I came up with my own solution to the problem but if I ever get the time I'd be happy to help since it makes my code redundant. 

Like just add an option to spawn in a group basically?

FabianLars:
> Like just add an option to spawn in a group basically?

That was the idea, yeah. Perhaps not realistic though. I think I lost the very-wip branch I worked on a few months ago though, will check later.

Noremac11800:
@FabianLars any news on this? This is issue is the last piece of the puzzle I need to get my app up and running. I have an idea of how I'm going to get around it, but it requires significantly more Rust code than I was hoping I would need to write. Ideally, child processes would have a `kill_all()` option instead of just `kill()` which would recursively terminate child processes produced by my original parent.

I don't have the ability to make the parent process cleanup its children and their children because the grandchild processes are 3rd party. My best attempt so far, without writing a more robust solution, is to target killing PIDs by process name, but that approach is fragile and hacky.

aka-danielZhang:
[...]

PR #3149, opened 2025-12-05
feat(command): implement Drop for CommandChild with kill_on_drop option

Tunglies:
Related:  #1332, https://github.com/tauri-apps/tauri/pull/14443 https://github.com/tauri-apps/tauri/issues/14360
Not totally resolved for these.

Simplfy impl Drop for CommandChild, only takes effect on tauri-plugin-shell, whether if it is sidecar or normal command. Taks advantage of rust's liftime control. Before the impl, drop has no any operation. Has no any additional external tool usage or invoke helper program. Competiable with tauri dev.

Usage:

```rust
let command = app_handle
    .clone()
    .shell()
    .sidecar(EXTERNAL_BIN)?;
command.set_kill_on_drop(true);
```

### DIFF
--- a/.changes/feat-impl-drop-for-commmand-child-in-plugin-shell.md
+++ b/.changes/feat-impl-drop-for-commmand-child-in-plugin-shell.md
@@ -1,0 +1,6 @@
+---
+"shell": minor:feat
+"shell-js": minor:feat
+---
+
+Impl Drop for ChildCommand, and gving boolean to determine kill on drop.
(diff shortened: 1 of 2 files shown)


Tunglies:
CC @Legend-Master and @FabianLars

Legend-Master:
I don't mind adding this but it doesn't really sound like something related to the issues/PRs you linked?

Tunglies:
If more complex logic is introduced in the future, using drop(command) will make it easier to extend the data structures. No more external system tool need invoked.

PR #3155, opened 2025-12-09
fix(fs): off by one error in readTextFileLines

### LINKED ISSUE #3154: plugin-fs readTextFileLines appends null byte to the end of each line
### Describe the bug

Due to an off by one error in the implementation of `readTextFileLines` the `done` flag (which is 0 when not done) gets appended to each line. This is the offending line:
https://github.com/tauri-apps/plugins-workspace/blob/e4a40f44232834727334ceb4f8a0db30553eded9/plugins/fs/guest-js/index.ts#L841
The `bytes` buffer should be resliced to `bytes.byteLength-1` to remove the done flag, before decoding.

### Reproduction

```ts
const path = await open({
    title: "Select UTF-8 encoded multi-line text file without null bytes!",
    multiple: false,
    directory: false,
});

if (path === null)
    return;

const lines = await readTextFileLines(path);

for await (const line of lines) {
    console.log(line.endsWith("\u0000"))
}
```

### Expected behavior

Console should log `false`, but currently logs `true`.

### Full `tauri info` output

```text
[✔] Environment
    - OS: Fedora 43.0.0 x86_64 (X64) (KDE on wayland)
    ✔ webkit2gtk-4.1: 2.50.3
    ✔ rsvg2: 2.61.0
    ✔ rustc: 1.91.1 (ed61e7d7e 2025-11-07)
    ✔ cargo: 1.91.1 (ea2d97820 2025-10-10)
    ✔ rustup: 1.28.2
    ✔ Rust toolchain: stable-x86_64-unknown-linux-gnu (default)
    - node: 22.20.0
    - npm: 10.9.3

[-] Packages
    - tauri 🦀: 2.9.4
    - tauri-build 🦀: 2.5.3
    - wry 🦀: 0.53.5
    - tao 🦀: 0.34.5
    - @tauri-apps/api  ⱼₛ: not installed!
    - @tauri-apps/cli  ⱼₛ: 2.9.5

[-] Plugins
    - tauri-plugin-log 🦀: 2.7.1
    - @tauri-apps/plugin-log  ⱼₛ: not installed!
    - tauri-plugin-fs 🦀: 2.4.4
    - @tauri-apps/plugin-fs  ⱼₛ: 2.4.4
    - tauri-plugin-dialog 🦀: 2.4.2
    - @tauri-apps/plugin-dialog  ⱼₛ: 2.4.2

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - devUrl: http://localhost:8080/
    - framework: Vue.js
    - bundler: Vite
```

I'm gonna provide a fix shortly.

EliasStar:
Due to an off by one error in the implementation of `readTextFileLines` the `done` flag (which is 0 when not done) gets appended to each line. This PR fixes this by adding the missing `- 1` to the end index of the call to `bytes.slice`.

Fixes #3154


Legend-Master (approved):
Thanks!

Issue #3145, opened 2025-12-05
plugin  store panic bad resource??


yuyang-ok:
### Describe the bug

just get this error out of blue
~~~
[2025-12-05 13:37:00.097] Thread 'main' panicked at '/Users/yuyang/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tauri-plugin-store-2.4.1/src/store.rs:200:61': called `Result::unwrap()` on an `Err` value: BadResourceId(3396632397)
   0: std::backtrace_rs::backtrace::libunwind::trace
             at /rustc/ed61e7d7e242494fb7057f2657300d9e77bb4fcb/library/std/src/../../backtrace/src/backtrace/libunwind.rs:117:9
   1: std::backtrace_rs::backtrace::trace_unsynchronized
             at /rustc/ed61e7d7e242494fb7057f2657300d9e77bb4fcb/library/std/src/../../backtrace/src/backtrace/mod.rs:66:14
   2: std::backtrace::Backtrace::create
             at /rustc/ed61e7d7e242494fb7057f2657300d9e77bb4fcb/library/std/src/backtrace.rs:331:13
   3: std::backtrace::Backtrace::capture
             at /rustc/ed61e7d7e242494fb7057f2657300d9e77bb4fcb/library/std/src/backtrace.rs:296:9
   4: dbpal::setup_panic_hook::{{closure}}::{{closure}}
             at ./dbnexus-app/src-tauri/src/main.rs:106:52
   5: <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call
             at /rustc/ed61e7d7e242494fb7057f2657300d9e77bb4fcb/library/alloc/src/boxed.rs:1999:9
   6: std::panicking::panic_with_hook
             at /rustc/ed61e7d7e242494fb7057f2657300d9e77bb4fcb/library/std/src/panicking.rs:842:13
   7: std::panicking::panic_handler::{{closure}}
             at /rustc/ed61e7d7e242494fb7057f2657300d9e77bb4fcb/library/std/src/panicking.rs:707:13
   8: std::sys::backtrace::__rust_end_short_backtrace
             at /rustc/ed61e7d7e242494fb7057f2657300d9e77bb4fcb/library/std/src/sys/backtrace.rs:174:18
   9: __rustc::rust_begin_unwind
             at /rustc/ed61e7d7e242494fb7057f2657300d9e77bb4fcb/library/std/src/panicking.rs:698:5
[...]

PR #3156, opened 2025-12-09
refactor(store): hold a weak reference on store

Legend-Master:
Option 3 of https://github.com/tauri-apps/plugins-workspace/issues/3145#issuecomment-3626636292

We now always hold onto a weak reference on the `Store` until all the strong references to it get dropped, this ensures that we will only ever have 1 `Store` for a single path (except for when the user uses `create_new`)

This is similar to the implementation from the beginning of https://github.com/tauri-apps/plugins-workspace/pull/1860

Not sure if it's worth it to be honest though

### DIFF
--- a/plugins/store/src/store.rs
+++ b/plugins/store/src/store.rs
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-License-Identifier: MIT
 
-use crate::{ChangePayload, StoreState};
+use crate::{validate_or_add_resource, validate_resource_id, ChangePayload, StoreState};
 use serde_json::Value as JsonValue;
 use std::{
     collections::HashMap,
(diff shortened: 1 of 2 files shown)


plugins/store/src/store.rs
```diff
@@ -550,11 +547,9 @@ impl<R: Runtime> Store<R> {
...
         let app = store.app.clone();
-        let state = app.state::<StoreState>();
-        let stores = state.stores.lock().unwrap();
-        if let Some(rid) = stores.get(&store.path).copied() {
-            drop(store);
-            drop(stores);
+        let state = app.state::<StoreState<R>>();
+        let mut stores = state.stores.lock().unwrap();
+        if let Some(rid) = validate_resource_id(&app, stores.get_mut(&store.path)) {
             let _ = app.resources_table().close(rid);
```
Legend-Master, line 553: This is technically not safe since we didn't hold a lock through out and the resource id can be reused (https://github.com/tauri-apps/tauri/pull/9272) ...

PR #3183, opened 2025-12-27
docs(shell): update example to include Encoding usage in `Command::spawn`

Tunglies:
See context in issue #1471 and PR #3148

CC @Legend-Master

### DIFF
--- a/.changes/doc-to-Encoding-usage-in-shell.md
+++ b/.changes/doc-to-Encoding-usage-in-shell.md
@@ -1,4 +1,6 @@
 ---
-"shell": patch
+shell: patch
+shell-js: patch
 ---
+
 Docs on example to Encoding usage in `Command::spawn`. No user facing changes.
--- a/.changes/doc-to-Encoding-usage-in-shell.md
+++ b/.changes/doc-to-Encoding-usage-in-shell.md
@@ -1,0 +1,4 @@
+---
+"shell": patch
+---
+Docs on example to Encoding usage in `Command::spawn`. No user facing changes.
--- a/.changes/doc-to-Encoding-usage-in-shell.md
+++ b/.changes/doc-to-Encoding-usage-in-shell.md
@@ -1,4 +1,4 @@
 ---
-"shell": patch:minor
+"shell": patch
 ---
 Docs on example to Encoding usage in `Command::spawn`. No user facing changes.
--- a/.changes/doc-to-Encoding-usage-in-shell.md
+++ b/.changes/doc-to-Encoding-usage-in-shell.md
@@ -1,4 +1,4 @@
 ---
-"shell": patch
+"shell": patch:minor
 ---
 Docs on example to Encoding usage in `Command::spawn`. No user facing changes.
--- a/plugins/shell/src/process/mod.rs
+++ b/plugins/shell/src/process/mod.rs
@@ -271,8 +271,8 @@ impl Command {
     ///   .setup(|app| {
     ///     let handle = app.handle().clone();
     ///     tauri::async_runtime::spawn(async move {
-    ///       let (mut rx, mut child) = handle.shell().command("cargo")
-    ///         .args(["tauri", "dev"])
+    ///       let (mut rx, mut child) = handle.shell().command("some-program")
+    ///         .args(["some-arg"])
     ///         .spawn()
     ///         .expect("Failed to spawn cargo");
     ///
--- a/plugins/shell/src/process/mod.rs
+++ b/plugins/shell/src/process/mod.rs
@@ -237,6 +237,35 @@ impl Command {
     /// # Examples
     ///
     /// ```rust,no_run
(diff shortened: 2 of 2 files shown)


.changes/doc-to-Encoding-usage-in-shell.md
```diff
@@ -0,0 +1,4 @@
+---
+"shell": patch:minor
```
Legend-Master:
```suggestion
"shell": patch
```

plugins/shell/src/process/mod.rs
```diff
@@ -247,10 +247,13 @@ impl Command {
     ///         .spawn()
     ///         .expect("Failed to spawn cargo");
     ///
+    ///       let encoding = Encoding::for_label(b"utf-8").unwrap();
     ///       let mut i = 0;
     ///       while let Some(event) = rx.recv().await {
     ///         if let CommandEvent::Stdout(line) = event {
+    ///           let (decoded, _, _) = encoding.decode(&line);
     ///           println!("got: {}", String::from_utf8(line).unwrap());
```
Legend-Master, line 255: We're already decoding from utf-8 here right?

plugins/shell/src/process/mod.rs
```diff
@@ -237,7 +237,7 @@ impl Command {
     /// # Examples
     ///
     /// ```rust,no_run
-    /// use tauri_plugin_shell::{process::CommandEvent, ShellExt};
+    /// use tauri_plugin_shell::{process::{CommandEvent, Encoding}, ShellExt};
     /// tauri::Builder::default()
     ///   .setup(|app| {
     ///     let handle = app.handle().clone();
```
Legend-Master, line 243: We could maybe add an example below in this `.setup` closure instead, or as another example/code block below the current one

Tunglies:
Added on a new example block, should be better now.

plugins/shell/src/process/mod.rs
```diff
@@ -262,6 +262,31 @@ impl Command {
     ///     Ok(())
     /// });
     /// ```
+    ///
+    /// # Examples Mannual Encoding
```
Legend-Master:
```suggestion
    /// Depending on the command you spawn, it might output in a specific encoding, to parse the output lines in this case:
```

plugins/shell/src/process/mod.rs
```diff
@@ -262,6 +262,31 @@ impl Command {
...
+    /// tauri::Builder::default()
+    ///   .setup(|app| {
+    ///     let handle = app.handle().clone();
+    ///     tauri::async_runtime::spawn(async move {
+    ///       let (mut rx, mut child) = handle.shell().command("cargo")
+    ///         .args(["tauri", "dev"])
+    ///         .spawn()
+    ///         .expect("Failed to spawn cargo");
+    ///
+    ///      let encoding = Encoding::for_label(b"utf-8").unwrap();
```
Legend-Master: Probably show a command that doesn't use the utf-8 encoding so this is isn't so confusing (currently we just provided an example that could be simplified to the above)

Tunglies: Initially, I wanted to write WINDOWS_1252, thinking it would correspond to the above. Let me fix that.

plugins/shell/src/process/mod.rs
```diff
@@ -262,6 +262,31 @@ impl Command {
...
+    ///       let (mut rx, mut child) = handle.shell().command("cargo")
+    ///         .args(["tauri", "dev"])
+    ///         .spawn()
+    ///         .expect("Failed to spawn cargo");
+    ///
+    ///      let encoding = Encoding::for_label(b"windows-1252").unwrap();
+    ///       while let Some(event) = rx.recv().await {
+    ///         if let CommandEvent::Stdout(line) = event {
+    ///           let (decoded, _, _) = encoding.decode(&line);
+    ///           println!("got: {}", decoded);
```
Legend-Master:
```suggestion
    ///           println!("got: {decoded}");
```

plugins/shell/src/process/mod.rs
```diff
@@ -262,6 +262,31 @@ impl Command {
...
+    ///
+    /// Depending on the command you spawn, it might output in a specific encoding, to parse the output lines in this case:
+    ///
+    /// ```rust,no_run
+    /// use tauri_plugin_shell::{process::{CommandEvent, Encoding}, ShellExt};
+    /// tauri::Builder::default()
+    ///   .setup(|app| {
+    ///     let handle = app.handle().clone();
+    ///     tauri::async_runtime::spawn(async move {
+    ///       let (mut rx, mut child) = handle.shell().command("cargo")
```
Legend-Master: `cargo` doesn't output in `windows-1252`, let's change it to maybe `some-program`, and the `tauri dev` args

Legend-Master (commented):
Also you may want to fix the indentation of the example 🙃

Legend-Master (approved):
Thanks

PR #3185, opened 2025-12-29
feat(fs): access security scoped resources on iOS

lucasfernog:
# Add iOS security-scoped resource support

## Summary

Adds support for handling security-scoped URLs on iOS, enabling access to files outside the app's sandbox (e.g., files selected via file picker).

## Changes

- **iOS security-scoped resource handling**: The `open()` method in `ios.rs` now automatically calls `startAccessingSecurityScopedResource()` when opening files via `file://` URLs
- **New API**: Added `stopAccessingSecurityScopedResource()` command and guest-js binding to manually stop accessing security-scoped resources when done
- **Path resolution**: Updated `resolve_path()` to start accessing security-scoped resources early in the path resolution process
- **Code organization**: Split `mobile.rs` into separate `ios.rs` and `android.rs` files for better maintainability
- **Dependencies**: Added `objc2-foundation` dependency for iOS to interact with NSURL APIs

## Platform-specific behavior

- **iOS**: Automatically handles security-scoped URLs; provides API to stop access
- **Android**: No changes (continues using content URI handling)
- **Desktop**: No changes (no-op for security-scoped resource API)

## Breaking changes

None.


plugins/fs/src/ios.rs
```diff
@@ -0,0 +1,113 @@
...
+                // URLWithString may return None for invalid URLs, but file:// URLs should be valid
+                let ns_url = unsafe { NSURL::URLWithString(&url_nsstring) };
+                if let Some(ns_url) = ns_url {
+                    // Start accessing the security-scoped resource
+                    // This is required for files outside the app's sandbox (e.g., from file picker)
+                    // Note: We don't call stopAccessingSecurityScopedResource here because
+                    // the file handle needs to remain accessible while the File is in use.
+                    // The access will be automatically stopped when the app is backgrounded or terminated.
+                    unsafe {
+                        let _ = ns_url.startAccessingSecurityScopedResource();
```
velocitysystems: The return value indicates whether the URL is _actually_ a security-scoped resource. Ignoring it means:
- No way to know if the call was necessary
- No error handling if the resource can't be accessed

**Recommendation**: Check the return value and potentially log/warn if it returns false for a URL that was expected to be security-scoped.

plugins/fs/src/ios.rs
```diff
@@ -0,0 +1,113 @@
...
+
+                // Create NSURL from the URL string
+                // URLWithString may return None for invalid URLs, but file:// URLs should be valid
+                let ns_url = unsafe { NSURL::URLWithString(&url_nsstring) };
+                if let Some(ns_url) = ns_url {
+                    // Start accessing the security-scoped resource
+                    // This is required for files outside the app's sandbox (e.g., from file picker)
+                    // Note: We don't call stopAccessingSecurityScopedResource here because
+                    // the file handle needs to remain accessible while the File is in use.
+                    // The access will be automatically stopped when the app is backgrounded or terminated.
```
velocitysystems, line 51: This comment is not entirely correct. Apple's documentation states: "You must balance every call to this method with a call to `stopAccessingSecurityScopedResource()`."

While the OS will eventually reclaim resources on termination, this doesn't constitute proper resource management for several reasons:
- Resource exhaustion — The system limits how many security-scoped resources can be active
- Long-running apps — Apps that stay active may accumulate leaked handles.
- Bookmark invalidation — Improperly managed access can contribute to security-scoped bookmark staleness.

**Recommendation**: Consider implementing RAII-style cleanup or at minimum document the requirement to call `stopAccessingSecurityScopedResource()` explicitly.

plugins/fs/src/commands.rs
```diff
@@ -1095,6 +1122,27 @@ pub fn resolve_path<R: Runtime>(
...
+                use objc2_foundation::{NSString, NSURL};
+
+                let url_string = url.as_str();
+                let url_nsstring = NSString::from_str(url_string);
+                let ns_url = unsafe { NSURL::URLWithString(&url_nsstring) };
+                if let Some(ns_url) = ns_url {
+                    // Start accessing the security-scoped resource
+                    // This is required for files outside the app's sandbox (e.g., from file picker)
+                    unsafe {
+                        let _ = ns_url.startAccessingSecurityScopedResource();
```
velocitysystems: Double `startAccessingSecurityScopedResource` calls. The PR calls `startAccessingSecurityScopedResource` in two places:
- `resolve_path()` in commands.rs:1125-1141
- `open()` in ios.rs:38-47

If a path is resolved and then opened, this results in two start calls but potentially only one stop call, creating an imbalance.

**Recommendation**: Centralize the security-scoped resource handling or document the expected call pattern.

plugins/fs/src/ios.rs
```diff
@@ -0,0 +1,113 @@
...
+        match path.into() {
+            FilePath::Url(url) if url.scheme() == "file" => {
+                // Handle security-scoped URLs on iOS
+                let url_string = url.as_str();
+                let url_nsstring = NSString::from_str(url_string);
+
+                // Create NSURL from the URL string
+                // URLWithString may return None for invalid URLs, but file:// URLs should be valid
+                let ns_url = unsafe { NSURL::URLWithString(&url_nsstring) };
+                if let Some(ns_url) = ns_url {
```
velocitysystems, line 46: If `NSURL::URLWithString `returns `None`, the code silently continues without starting security-scoped access, which could lead to confusing permission errors later.

velocitysystems:
@onehumandev Can you please take a look at this and compare with changes in [PR 3136](https://github.com/tauri-apps/plugins-workspace/pull/3136)?

onehumandev:
c. @velocitysystems  Short version: AFAIK, for `startAccessingSecurityScopedResource` to hav an effect, FilePickerController.swift](https://github.com/tauri-apps/plugins-workspace/blob/31415effdf5a9ced19934a681cb044a732174088/plugins/dialog/ios/Sources/FilePickerController.swift#L122C1-L123C59) will need to be modified to not automatically save a copy of the selected file to the app sandbox.

Long version:
Currently, the current implementation of the [FilePickerController.swift](https://github.com/tauri-apps/plugins-workspace/blob/31415effdf5a9ced19934a681cb044a732174088/plugins/dialog/ios/Sources/FilePickerController.swift#L122C1-L123C59), copies any file selected from the document picker to the app sandbox.

This means that `startAccessingSecurityScopedResource` is not required for any file that is opened with the current implementation, at the expense of duplicated files. 

For `startAccessingSecurityScopedResource` to have an effect, an option would need to be included for files that are opened by the file picker to not copy the file to the sandbox, but rather allow access with `startAccessingSecurityScopedResource`.

In other words, it would essentially be an "open mode"; either "copy" (the current production behavior) or "scoped" (which would include a call to `startAccessingSecurityScopedResource`).

PR #3136 includes these changes; however, if that is not the path the project would like to take, I am happy to help with whatever alternative is preferred :) 

Further explanation, from comment in above linked PR:
```

    /// In which cases do we need to save a copy of a file selected by a user to the app sandbox?
	/// In short, only when the file is **not** selected using UIDocumentPickerDelegate.
	/// For the rest of the cases, we need to write a copy of the file to the app sandbox. 
	/// 
	/// For PHPicker (used for photos and videos), `NSItemProvider.loadFileRepresentation` returns a temporary file URL that is deleted after the completion handler.
	/// The recommendation is to [Persist](https://developer.apple.com/documentation/foundation/nsitemprovider/2888338-loadfilerepresentation) the file by moving/copying
	/// it to your app’s directory within the completion handler.
	/// 
	/// If available, `loadInPlaceFileRepresentation` can open a file in place; Photos assets typically do not support true in-place access,
	/// so fall back to persisting a local file.
	/// Ref: https://developer.apple.com/documentation/foundation/nsitemprovider/2888335-loadinplacefilerepresentation
	/// 
	/// For UIDocumentPicker, prefer “open in place” and avoid copying when possible.
	/// Ref: https://developer.apple.com/documentation/uikit/uidocumentpickerviewcontroller
```

plugins/fs/src/ios.rs
lucasfernog: direct cleanup can't be implemented without breaking changes in this case :/ we'll have to update the docs

plugins/fs/src/commands.rs
lucasfernog: that's two different code paths - resolve_file vs resolve_path, both of them are never called on the same API usage

lucasfernog:
@onehumandev I love the dialog plugin changes you suggested. What I don't really agree with is adding startFileAccess and endFileAccess to the dialog plugin - the same mechanism will be required to e.g. read file association input. That's why I decided to make some changes to the filesystem plugin instead.

If you can drop the file access APIs from your PR I can merge it to include the changes required to not create a file copy on the dialog APIs.

onehumandev:
@lucasfernog  Sounds good; updated #3136 in accord with latest comments

velocitysystems:
Thanks @lucasfernog @onehumandev. I think this is a great solution all around. 🎉 

So...We have new APIs for handling security scoped resources available in the `fs` plugin and these can be used for scoped access in the `dialog` plugin where required. Are any further changes required in the `dialog` plugin @onehumandev to make these work together?

velocitysystems (approved):
Looks good @lucasfernog. Please see my above comment. What further changes are required in the dialog plugin to make this work with this functionality in the `fs` plugin?
cc: @onehumandev

onehumandev:
> Looks good @lucasfernog. Please see my above comment. What further changes are required in the dialog plugin to make this work with this functionality in the `fs` plugin? cc: @onehumandev

I will deep dive into this tomorrow and Friday and make sure everything works as expected, and if there are any wires needing connecting.

lucasfernog:
> Looks good @lucasfernog. Please see my above comment. What further changes are required in the dialog plugin to make this work with this functionality in the `fs` plugin? cc: @onehumandev

@velocitysystems I tried this branch with #3136 and seems like everything is working fine now. Should be good to go.

onehumandev:
> > Looks good @lucasfernog. Please see my above comment. What further changes are required in the dialog plugin to make this work with this functionality in the `fs` plugin? cc: @onehumandev
> 
> @velocitysystems I tried this branch with #3136 and seems like everything is working fine now. Should be good to go.

@velocitysystems  Confirmed; I tested it locally as well with our current use cases, and all works as expected.

plugins/fs/src/ios.rs
FabianLars: > direct cleanup can't be implemented without breaking changes in this case

How would these look like? Mind adding a TODO comment (or issue) so we can follow up in v3? Edit: or would it be too ugly and therefore won't be implemented?

FabianLars (dismissed):

FabianLars (dismissed):

plugins/fs/src/ios.rs
lucasfernog: https://github.com/tauri-apps/plugins-workspace/pull/3185/changes/526e2e13358cfc70975210597fcfeee827ba5b2b

FabianLars (approved):

PR #3186, opened 2025-12-29
feat(deep-link): validate new intent URLs against configured deep links

lucasfernog:
the onNewIntent can be triggered for notifications and upcoming file associations, so we must validate the intent data against the configured deep links

### DIFF
--- a/plugins/deep-link/android/src/main/java/DeepLinkPlugin.kt
+++ b/plugins/deep-link/android/src/main/java/DeepLinkPlugin.kt
@@ -6,9 +6,9 @@ package app.tauri.deep_link
 
 import android.app.Activity
 import android.content.Intent
-import android.os.Bundle
+import android.net.Uri
+import android.os.PatternMatcher
 import android.webkit.WebView
-import app.tauri.Logger
 import app.tauri.annotation.InvokeArg
 import app.tauri.annotation.Command
 import app.tauri.annotation.TauriPlugin
@@ -22,12 +22,28 @@ class SetEventHandlerArgs {
     lateinit var handler: Channel
 }
 
+@InvokeArg
+class AssociatedDomain {
+    var scheme: List<String> = listOf("https", "http")
+    var host: String? = null
+    var path: List<String> = listOf()
+    var pathPattern: List<String> = listOf()
+    var pathPrefix: List<String> = listOf()
+    var pathSuffix: List<String> = listOf()
+}
+
+@InvokeArg
+class PluginConfig {
+    var mobile: List<AssociatedDomain> = listOf()
+}
+
 @TauriPlugin
 class DeepLinkPlugin(private val activity: Activity): Plugin(activity) {
     //private val implementation = Example()
     private var webView: WebView? = null
     private var currentUrl: String? = null
     private var channel: Channel? = null
+    private var config: PluginConfig? = null
 
     companion object {
         var instance: DeepLinkPlugin? = null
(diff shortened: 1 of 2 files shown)


plugins/deep-link/android/src/main/java/DeepLinkPlugin.kt
```diff
@@ -51,27 +67,105 @@ class DeepLinkPlugin(private val activity: Activity): Plugin(activity) {
...
+        }
+
+        val scheme = uri.scheme ?: return false
+        val host = uri.host
+        val path = uri.path ?: ""
+
+        // Check if URL matches any configured mobile deep link
+        for (domain in config.mobile) {
+            // Check scheme
+            if (!domain.scheme.contains(scheme)) {
```
velocitysystems: Scheme comparison is case-sensitive. URI schemes are technically case-insensitive per RFC 3986. Consider normalizing to lowercase.

plugins/deep-link/android/src/main/java/DeepLinkPlugin.kt
```diff
@@ -51,27 +67,105 @@ class DeepLinkPlugin(private val activity: Activity): Plugin(activity) {
...
+                continue
+            }
+
+            // Check host (if configured)
+            if (domain.host != null) {
```
velocitysystems, line 127: Host comparison is case-sensitive. DNS hostnames are case-insensitive. Should use `host.equals(domain.host, ignoreCase = true)`.

plugins/deep-link/android/src/main/java/DeepLinkPlugin.kt
```diff
@@ -51,27 +67,105 @@ class DeepLinkPlugin(private val activity: Activity): Plugin(activity) {
...
+
+        val uri = try {
+            Uri.parse(url)
+        } catch (e: Exception) {
+            // not a URL
+            return false
+        }
+
+        val scheme = uri.scheme ?: return false
+        val host = uri.host
```
velocitysystems, line 116: Null safety on `uri.host`. The code gets `val host = uri.host` which can be `null`, but then compares `host != domain.host` without null handling.

plugins/deep-link/android/src/main/java/DeepLinkPlugin.kt
```diff
@@ -51,27 +67,105 @@ class DeepLinkPlugin(private val activity: Activity): Plugin(activity) {
 
     override fun load(webView: WebView) {
         instance = this
+        config = getConfig(PluginConfig::class.java)
+
+        super.load(webView)
```
velocitysystems, line 72: `super.load(webView)` moved before intent handling. Is this intentional? Verify this doesn't change behavior unexpectedly.

plugins/deep-link/android/src/main/java/DeepLinkPlugin.kt
```diff
@@ -22,12 +22,28 @@ class SetEventHandlerArgs {
     lateinit var handler: Channel
 }
 
+@InvokeArg
+class AssociatedDomain {
+    var scheme: List<String> = listOf("https", "http")
```
velocitysystems, line 27: What happens for custom scheme deep links (e.g., myapp://)?

lucasfernog: that's just the default value, custom schemes are configured manually by the user and will overwrite this

plugins/deep-link/android/src/main/java/DeepLinkPlugin.kt
lucasfernog: super.load does nothing, this doesn't change behavior, thanks for catching this!

plugins/deep-link/android/src/main/java/DeepLinkPlugin.kt
lucasfernog: we only compare hosts when domain.host is not null (meaning the user requested a host validation rather than just scheme check - app link instead of custom scheme deep link). In this case if `host`  is null then `host != domain.host` is true and the intent is filtered out.

plugins/deep-link/android/src/main/java/DeepLinkPlugin.kt
lucasfernog: thanks, pushed

plugins/deep-link/android/src/main/java/DeepLinkPlugin.kt
lucasfernog: thanks, pushed

velocitysystems (approved):

FabianLars (approved):
this feels like something we may want to move to the rust side

PR #3194, opened 2026-01-04
fix(single-instance): unconventional dbus names (fix #3184)

### LINKED ISSUE #3184: Single-instance plugin D-Bus naming violates Flatpak/D-Bus conventions - causing Flathub rejections
## Summary

The Tauri single-instance plugin's D-Bus service naming pattern is incompatible with Flatpak/Flathub requirements and violates D-Bus naming conventions, causing Tauri apps to be rejected from Flathub.

### Current Behavior
The single-instance plugin uses D-Bus service names in the format: `org.{appId}.SingleInstance`

For example, an app with:
- Flatpak ID: `me.amankhanna.opendeck`
- Tauri app ID: `opendeck`

Creates D-Bus service: `org.opendeck.SingleInstance`

### Problem
Flathub's flatpak-builder-lint tool rejects this pattern because:
1. The D-Bus service name doesn't match the app's reverse-DNS Flatpak ID
2. It violates D-Bus naming conventions which require owned names to be under the application's namespace
3. It creates potential naming conflicts between different apps

### Impact
- Flatpak maintainers have rejected exception requests for this pattern
- Two existing Tauri apps with exceptions (`com.bilingify.readest` and `dev.mrquantumoff.mcmodpackmanager`) will have their exceptions removed
- Flathub maintainers have stated they will block all Tauri apps if this isn't fixed
- Affects all Tauri apps using the single-instance plugin on Flatpak/Snap

## References

- Original PR discussion: https://github.com/flathub-infra/flatpak-builder-lint/pull/819
- Tauri single-instance docs: https://v2.tauri.app/plugin/single-instance/#usage-in-snap-and-flatpak
- D-Bus naming conventions: https://dbus.freedesktop.org/doc/dbus-specification.html#message-bus-names

## Additional Context

Quote from Flathub maintainer (bbhtt):
> "It needs to follow the appid not use a random dbus name that starts with `org.`. That's literally how everything else works including it is how things should be done per the dbus spec. [...] We will block all Tauri apps if we have to."

This is a critical issue for the Tauri ecosystem's Flatpak/Linux distribution story.

### Reproduction

_No response_

### Expected behavior

_No response_

### Full `tauri info` output

```text
[✔] Environment
    - OS: Mac OS 15.6.1 arm64 (X64)
    ✔ Xcode: 26.2
    ✔ rustc: 1.92.0 (ded5c06cf 2025-12-08)
    ✔ cargo: 1.92.0 (344c4567c 2025-10-21)
    - node: 22.9.0
    - pnpm: 10.25.0
    - yarn: 1.22.22
    - npm: 10.9.1

[-] Packages
    - tauri 🦀: 2.9.5
    - tauri-build 🦀: 2.5.3
    - wry 🦀: 0.53.5
    - tao 🦀: 0.34.5
    - @tauri-apps/api  ⱼₛ: 2.9.1
    - @tauri-apps/cli  ⱼₛ: 2.9.6

[-] Plugins
    - tauri-plugin-persisted-scope 🦀: 2.3.4
    - @tauri-apps/plugin-persisted-scope  ⱼₛ: not installed!
    - tauri-plugin-fs 🦀: 2.4.4
    - @tauri-apps/plugin-fs  ⱼₛ: 2.4.4
    - tauri-plugin-cli 🦀: 2.4.1
    - @tauri-apps/plugin-cli  ⱼₛ: 2.4.1
    - tauri-plugin-dialog 🦀: 2.4.2
    - @tauri-apps/plugin-dialog  ⱼₛ: 2.4.2
    - tauri-plugin-log 🦀: 2.7.1
    - @tauri-apps/plugin-log  ⱼₛ: 2.7.1
    - tauri-plugin-updater 🦀: 2.9.0
    - @tauri-apps/plugin-updater  ⱼₛ: 2.9.0
    - tauri-plugin-window-state 🦀: 2.4.1
    - @tauri-apps/plugin-window-state  ⱼₛ: not installed!
    - tauri-plugin-haptics 🦀: 2.3.2
    - @tauri-apps/plugin-haptics  ⱼₛ: 2.3.2
    - tauri-plugin-websocket 🦀: 2.4.1
    - @tauri-apps/plugin-websocket  ⱼₛ: 2.4.1
    - tauri-plugin-process 🦀: 2.3.1
    - @tauri-apps/plugin-process  ⱼₛ: 2.3.1
    - tauri-plugin-http 🦀: 2.5.4
    - @tauri-apps/plugin-http  ⱼₛ: 2.5.4
    - tauri-plugin-os 🦀: 2.3.2
    - @tauri-apps/plugin-os  ⱼₛ: 2.3.2
    - tauri-plugin-shell 🦀: 2.3.3
    - @tauri-apps/plugin-shell  ⱼₛ: 2.3.3
    - tauri-plugin-opener 🦀: 2.5.2
    - @tauri-apps/plugin-opener  ⱼₛ: 2.5.2
    - tauri-plugin-deep-link 🦀: 2.4.5
    - @tauri-apps/plugin-deep-link  ⱼₛ: 2.4.5
    - tauri-plugin-single-instance 🦀: 2.3.6
    - @tauri-apps/plugin-single-instance  ⱼₛ: not installed!

[-] App
    - build-type: bundle
    - CSP: frame-src 'self' blob: asset: http://asset.localhost https://*.stripe.com; font-src 'self' blob: data: asset: http://asset.localhost tauri: https://db.onlinewebfonts.com https://cdn.jsdelivr.net https://fonts.gstatic.com https://chinese-fonts-cdn.netlify.app https://cdnjs.cloudflare.com; default-src 'self' 'unsafe-inline' blob: data: customprotocol: asset: http://asset.localhost ipc: http://ipc.localhost; connect-src 'self' blob: data: asset: http://asset.localhost ipc: http://ipc.localhost http://*:* https://*:* https://*.sentry.io https://*.posthog.com https://*.deepl.com https://*.wikipedia.org https://*.wiktionary.org https://*.supabase.co https://*.readest.com wss://speech.platform.bing.com https://*.cloudflarestorage.com https://translate.googleapis.com https://translate.toil.cc https://*.microsofttranslator.com https://edge.microsoft.com https://*.googleusercontent.com; img-src 'self' blob: data: asset: http://asset.localhost https://* https://*:* http://* http://*:*; style-src 'self' 'unsafe-inline' blob: asset: http://asset.localhost https://cdn.jsdelivr.net https://fonts.googleapis.com https://chinese-fonts-cdn.netlify.app https://cdnjs.cloudflare.com; script-src 'self' 'unsafe-inline' 'unsafe-eval' blob: asset: http://asset.localhost https://*.sentry.io https://*.posthog.com  https://*.stripe.com
    - frontendDist: ../out
    - devUrl: http://localhost:3000/
    - framework: React (Next.js)
    - bundler: Webpack
```

### Stack trace

```text

```

### Additional context

_No response_

mrquantumoff:
Flathub no longer allows the "org.app_id.SingleInstance" dbus id that was allowed as an exception previously. Thus, I implemented an optional feature that would allow developers to use a more conventional dbus name that would pass the Flathub inspections without causing extra issues.

The Flathub discussion
https://github.com/flathub-infra/flatpak-builder-lint/pull/819#issuecomment-3695837478


FabianLars:
I really appreciate the PR but please check out https://github.com/tauri-apps/plugins-workspace/issues/3184, we've had a slightly different approach in mind.

mrquantumoff:
Thank you for informing me about your approach to this issue. I suppose your way of solving the problem is better than just adding an another feature to the crate. In this case, I'll wait for the official implementation and use my own implementation for the time being.

FabianLars:
nono your implementation wasn't that far off. I just wouldn't put it behind a feature flag and instead force the new ids on all users 🤷

FabianLars:
one thing that's a bit of an annoyance with zbus is the interface name, i'd like to have that match the dbus name but it needs to be defined at build time if we don't want to code everything ourselves. The spec says "should ..." when it comes to that though so ig it doesn't matter that much

mrquantumoff:
In that case, I'll rework some part of it, to match the intended vision a bit more

mrquantumoff:
I added an option to set the DBus ID at compile time. Made the conventional dbus naming the standard one. I chose to check for ``DBUS_ID`` instead of ``FLATPAK_ID`` in order to keep better compatibility with people who want to package snaps.

amrbashir:
Please document that env var in lib.rs documentation and README.md

amrbashir:
Please document that env var in lib.rs documentation and README.md

FabianLars:
As far as i understood it, FLATPAK_ID is set by flatpak _at runtime_. A compile time env var in my opinion makes little sense and if we want that we should rather have a setter in rust.

amrbashir:
Yup, that's actually better

chrox:
👍 I’ve confirmed that this approach works for releasing new updates on Flathub.

Here’s what needs to be done:

1. When building the Tauri app, set an environment variable, for example: ```export DBUS_ID=com.bilingify.readest```
The `DBUS_ID` must exactly match the Flathub app ID defined in the Flatpak manifest YAML file.

2. In the Flatpak manifest, you no longer need to define own-name or talk-name for the D-Bus names.

See https://github.com/flathub/com.bilingify.readest

mrquantumoff:
Sorry for not signing my commits beforehand and for messing up some of the existing ones. I am not used to signing commits with git.

FabianLars:
looks like you meant to force-push? your old unsigned commits are still there. I'm very bad with git operations either so i can't really help. I personally would probably reset the branch and copy over the changes manually 😅 

On the PR changes themselves. I think we should remove the env var reading. If devs want that they can use the setter with the `env!()` macro themselves.
For the setter we should not rely on LazyLock (or the once_cell crate if we were to get down the MSRV), but instead use the Builder pattern, with the init function being a shortcut for those not wanting to set the id. I think the opener plugin's code is a fairly good example https://github.com/tauri-apps/plugins-workspace/blob/v2/plugins/opener/src/lib.rs#L178-L244

FabianLars:
@amrbashir Sorry but i think i'll need your review here since my changes were a bit much x)

plugins/single-instance/src/platform_impl/linux.rs
```diff
@@ -29,35 +26,49 @@ impl<R: Runtime> SingleInstanceDBus<R> {
...
+) -> TauriPlugin<R> {
     plugin::Builder::new("single-instance")
-        .setup(|app, _api| {
-            #[cfg(feature = "semver")]
-            let id = dbus_id(app.config(), app.package_info().version.clone());
-            #[cfg(not(feature = "semver"))]
-            let id = dbus_id(app.config());
+        .setup(move |app, _api| {
+            let dbus_name = if let Some(dbus_id) = dbus_id {
+                dbus_id
```
amrbashir: why aren't we appending `SingleInstance` suffix here as well?

mrquantumoff: Because developers should decide for themselver whether they want it. I think that's what most people would want when overriding the ID

FabianLars: i was very indecisive about this change as well. In theory I'd prefer your approach of adding the suffix but the dbus_id setter felt like it should overwrite it as a whole.

For fixing the flatpak issues both approaches should work.

amrbashir: I think we should add the suffix and document that in the `dbus_id` method

mrquantumoff: Respectfully, I disagree. I believe that if a developer wants to overwrite the dbus id they should have full control of it as there could be cases where people might want to use a fully custom dbus id (despite how illogical they may be). However, in the end you're the maintainer, so it's up to you.

amrbashir: My reasoning is that user might think we want their `dbus_id` and we are gonna use it as is which might cause conflicts if they use the same dbus_id for another service or something. 

If we are not appending a suffix, then I would still recommend in the documentation to append a unique suffix or something.

FabianLars: i mean i get the sentiment but i really can't think of a reason why you'd also want to overwrite the subname or component name, whatever you wanna call it. They don't have the same limitations as the main part (the one that broke flatpaks) and are very specific to the "component", in this case the single instance plugin (but any other plugin could also add their own sub)

FabianLars: and we can add another setter if there arises a need for this 🤷

plugins/single-instance/src/platform_impl/macos.rs
```diff
@@ -63,7 +63,7 @@ fn socket_path(config: &Config, _package_info: &tauri::PackageInfo) -> PathBuf {
     #[cfg(feature = "semver")]
     let identifier = format!(
         "{identifier}_{}",
-        semver_compat_string(_package_info.version.clone()),
+        &semver_compat_string(_package_info.version),
```
amrbashir:
```suggestion
        semver_compat_string(&_package_info.version),
```

amrbashir (approved):

plugins/single-instance/src/platform_impl/macos.rs
FabianLars: hahaha, i should have gone to bed instead 🤣

nekename:
Thank you to all involved. This now brings my app to 99% feature parity when running on Flatpak :)

PR #3195, opened 2026-01-05
Bump reqwest to 0.13.1

Ludea:
Bump reqwest to 0.13.

Since this version, rustls use aws-lc-rs instead of ring as default crypto provider


FabianLars:
We can't do this. It's a breaking change because we re-export reqwest in the http client. On top of that, your feature flag changes are also breaking changes.

Also, as discussed in https://github.com/tauri-apps/tauri/pull/14724 / https://github.com/tauri-apps/tauri/pull/14726 we should keep ring as the default.

Thanks for the PR nonetheless :)

PR #3210, opened 2026-01-14
fix(http): manually set origin header to tauri://localhost

FabianLars:
https://discord.com/channels/616186924390023171/1454227211325538456


FabianLars:
why did vscode not tell me about that. i actually expected that error and was surprised that it didn't fail

PR #3214, opened 2026-01-15
fix(deep-link): ChromeOS deep link calls filtered and ignored by plugin (fix #3207)

### LINKED ISSUE #3207: [bug][deep-link] ChromeOS deep links never processed by plugin
### Describe the bug

The deep-link plugin never processes callbacks with custom-schemes for Android applications running on ChromeOS. This potentially would also apply to http/s applinks too.

Using `adb logcat`, I was able to confirm that the custom scheme is correctly registered and passed to my application from the browser. Additionally, the "intent" is processed in the Kotlin code. The issue seems to be that it passes a ChromeOS-specific action: `org.chromium.arc.intent.action.VIEW`

The plugin, in this [file](https://github.com/tauri-apps/plugins-workspace/blob/f3d75f7abb536ba70d1746801096af6d9be7f8b1/plugins/deep-link/android/src/main/java/DeepLinkPlugin.kt#L57), checks specifically for `android.intent.action.VIEW` and would therefore ignore the chromium variant of the action, filtering away this callback.

When I get time, I will test a fork with the proposed change to see if it resolves the issue with my Chromebook and submit a PR accordingly.

### Reproduction

Build a Tauri v2 app for Android with deep-link plugin configured
Install and run on a Chromebook (device bridging should work, though I was forced to use the PlayStore's internal release process).
Trigger a deep link (e.g., myapp://callback from a browser)
The `deep_link().on_open_url()` callback is never invoked.

### Platform and versions
I have the latest ChromeOS versions but as per the context below, this seems to be quite an old issue in terms of the OS behaviour.
OS: ChromeOS with ARC (143.0.7499.196 Chrome, 16463.72.0 Nissa)
Tauri: 2.9.5
tauri-plugin-deep-link: 2.4.5

### Expected behavior

_No response_

### Full `tauri info` output

```text
Note: I build tauri on a Macbook and ship to the Chromebook via the Playstore.

[✔] Environment
    - OS: Mac OS 26.1.0 arm64 (X64)
    ✔ Xcode: 26.2
    ✔ rustc: 1.90.0 (1159e78c4 2025-09-14)
    ✔ cargo: 1.90.0 (840b83a10 2025-07-30)
    - node: 20.19.0
    - npm: 10.8.2

[-] Packages
    - tauri 🦀: 2.9.5
    - tauri-build 🦀: 2.5.3
    - wry 🦀: 0.53.4, (outdated, latest: 0.53.5)
    - tao 🦀: 0.34.5
    - @tauri-apps/api : not installed!
    - @tauri-apps/cli : 2.8.4 (outdated, latest: 2.9.6)

[-] Plugins
    - tauri-plugin-opener 🦀: 2.5.2, (outdated, latest: 2.5.3)
    - @tauri-apps/plugin-opener : 2.5.2 (outdated, latest: 2.5.3)
    - tauri-plugin-log 🦀: 2.8.0
    - @tauri-apps/plugin-log : 2.8.0
    - tauri-plugin-deep-link 🦀: 2.4.5, (outdated, latest: 2.4.6)
    - @tauri-apps/plugin-deep-link : 2.4.6
    - tauri-plugin-notification 🦀: 2.3.3
    - @tauri-apps/plugin-notification : 2.3.1 (outdated, latest: 2.3.3)
    - tauri-plugin-os 🦀: 2.3.2
    - @tauri-apps/plugin-os : 2.3.2
    - tauri-plugin-single-instance 🦀: 2.3.4, (outdated, latest: 2.3.7)

[-] App
    - build-type: bundle
    - CSP: connect-src ipc: http://ipc.localhost 'self' https://vtkrnbpgtvtbhkqwcwqm.supabase.co wss://vtkrnbpgtvtbhkqwcwqm.supabase.co;; img-src 'self' data:; style-src 'self' https://fonts.googleapis.com; font-src https://fonts.gstatic.com
    - frontendDist: ../dist
    - bundler: Vite
```

### Stack trace

```text

```

### Additional context

This is a known ChromeOS/ARC++ behaviour documented in [React Native issue #20301](https://github.com/facebook/react-native/issues/20301)

joshIsCoding:
ChromeOS with ARC uses a different intent view action which is presently excluded by the tauri deep-link plugin. This PR explicitly checks for and includes this variation. I've verified the behaviour resolves deep-link functionality in an Android application of my own deployed to a Chromebook via the Playstore.

Intended to resolve https://github.com/tauri-apps/plugins-workspace/issues/3207


joshIsCoding:
~Just noticed my commit signing didn't work; will try again~

Resolved.

plugins/deep-link/build.rs
```diff
@@ -25,6 +25,8 @@ fn intent_filter(domain: &AssociatedDomain) -> String {
     format!(
         r#"<intent-filter {auto_verify}>
     <action android:name="android.intent.action.VIEW" />
+    <!-- ChromeOS ARC++ uses a different action for deep links -->
+    <action android:name="org.chromium.arc.intent.action.VIEW" />
```
FabianLars, line 29: is this really needed? https://github.com/facebook/react-native/issues/20301 sounds like it isn't for some reason. afaik multiple actions in a single intent-filter is fine but just to be sure.

plugins/deep-link/android/src/main/java/DeepLinkPlugin.kt
```diff
@@ -31,6 +31,11 @@ class DeepLinkPlugin(private val activity: Activity): Plugin(activity) {
 
     companion object {
         var instance: DeepLinkPlugin? = null
+        private const val CHROMEOS_ACTION_VIEW = "org.chromium.arc.intent.action.VIEW"
```
FabianLars: i try to avoid using the companion object (dunno why but i always read opinions to avoid it 🤷 )

can the const not be moved outside of the class declaration? If not i wouldn't mind hardcoding it in isViewIntent directly since it's only used there anyway.

FabianLars:
thanks for the PR! we'll also need a small changefiles (check the .changes folder)  :)

plugins/deep-link/build.rs
joshIsCoding: Yeah this was warranted to avoid essentially the same issue, though it potentially depends on how dependent apps apply the intent filters

plugins/deep-link/android/src/main/java/DeepLinkPlugin.kt
joshIsCoding: Sure will move it 👍

joshIsCoding:
@FabianLars please forgive the extended delay to my engagement with your feedback - work priorities took over, and we made do with my local fork to unblock our deep-link development. 

All requested changes are now present; please let me know if you have any further requirements

plugins/deep-link/android/src/main/java/DeepLinkPlugin.kt
```diff
@@ -74,7 +77,7 @@ class DeepLinkPlugin(private val activity: Activity): Plugin(activity) {
 
         val intent = activity.intent
 
-        if (intent.action == Intent.ACTION_VIEW && intent.data != null) {
+        if (isViewIntent(Intent.ACTION_VIEW) && intent.data != null) {
```
FabianLars: looks like you meant to do `isViewIntent(intent.action)` instead? (same in onNewIntent)

joshIsCoding: Oh thank you, good spot, sorry!

joshIsCoding: Now fixed - this fell out of the merge conflict resolution, my bad

FabianLars (approved):
thanks :)

PR #3228, opened 2026-01-21
fix(http): ensure body resources are released on stream cancellation

### LINKED ISSUE #3227: [http] Resource leak: dropBody is not invoked upon stream cancellation
### Describe the bug

The current `ReadableStream` implementation in the `http-js` packages lacks a `cancel` lifecycle hook. Consequently, when a consumer manually terminates the stream, the internal cleanup function `dropBody()` is never executed.

#### **The Problem**

In the Web Streams API, the `cancel` method is the standard signal from the consumer that they are no longer interested in the data. Without the `cancel` hook, terminating a stream does not release the underlying HTTP resources.

#### **Scenarios affected:**

1. Manually calling `reader.cancel()`.
2. Breaking out of a `for await...of` loop early.

### Reproduction

```typescript
/**
 * SCENARIO 1: Explicit reader cancellation
 * Previously, this never triggers dropBody() to release resources
 */
async function testExplicitCancel() {
  const response = await http.get('https://api.example.com/stream');

  const { value } = await reader.read();
  console.log("Read initial chunk, now cancelling...");

  // This call triggers the 'cancel' hook in ReadableStream
  await reader.cancel("Consumer no longer needs data");
}

/**
 * SCENARIO 2: Early break from async iteration
 * Async iterators automatically call .cancel() when the loop is broken.
 */
async function testAsyncIterationBreak() {
  const response = await http.get('https://api.example.com/large-data');

  for await (const chunk of response.body) {
    if (shouldStopEarly(chunk)) {
      console.log("Condition met, breaking loop...");
      break; // The 'break' triggers stream cancellation
    }
  }
}

```

### Expected behavior

The cancel hook should be implemented to ensure that any stream termination from the consumer side triggers the necessary cleanup logic.

### Full `tauri info` output

```text
[✔] Environment
    - OS: Mac OS 15.7.3 arm64 (X64)
    ✔ rustc: 1.92.0 (ded5c06cf 2025-12-08)
    ✔ cargo: 1.92.0 (344c4567c 2025-10-21)
    - node: 22.14.0
    - pnpm: 10.0.0
    - npm: 10.9.2
    - bun: 1.2.21
    - deno: deno 2.5.6

[-] Packages
    - tauri 🦀: 2.9.5
    - tauri-build 🦀: 2.5.3
    - wry 🦀: 0.53.5, (outdated, latest: 0.54.1)
    - tao 🦀: 0.34.5
    - @tauri-apps/api : not installed!
    - @tauri-apps/cli : 2.5.0 (outdated, latest: 2.9.6)

[-] Plugins
    - tauri-plugin-opener 🦀: 2.5.3
    - @tauri-apps/plugin-opener : not installed!

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../src
```

### Stack trace

```text

```

### Additional context

https://github.com/tauri-apps/plugins-workspace/blob/25ad21beff11689cf453df0f6fc7dfd9e2f5e6d0/plugins/http/guest-js/index.ts#L270

Image

idootop:
### Summary

This PR addresses a resource management issue where HTTP response bodies were not being properly released if the stream was cancelled by the consumer. By implementing the `cancel` hook in the `ReadableStream` constructor, we ensure that the internal `dropBody()` logic is triggered, preventing memory leaks and hanging connections.

### Problem

Previously, the implementation only handled `AbortSignal` events. However, if a user:

1. Manually called `reader.cancel()`, or
2. Exited a `for await...of` loop early via `break` or `return`,
The `ReadableStream` would stop, but the underlying resources managed by `dropBody()` remained active.

### Solution

Added the `cancel` property to the `ReadableStream` source object to bridge the gap between the Web Stream lifecycle and our internal resource management.

```typescript
// Ensure cleanup if the consumer cancels the stream
cancel: () => dropBody()

```

Fixes https://github.com/tauri-apps/plugins-workspace/issues/3227


.changes/fix-http-stream-cancellation.md
```diff
@@ -0,0 +1,5 @@
+---
+http-js: patch
+---
+
+Added the cancel hook to ReadableStream to ensure resources are released via dropBody() when a consumer stops reading.
```
amrbashir: The change file should focus on the user facing changes and not mention internals like `dropBody`

```suggestion
---
http-js: patch:bug
---

Cleanup resource when the returned `ReadableStream.cancel` is called to avoid memory leaks
```

amrbashir (commented):
Thank you, I just left one comment to fix the change file.

.changes/fix-http-stream-cancellation.md
idootop: Done.

amrbashir (approved):
Thank you

PR #3236, opened 2026-01-26
fix(fs): fix crash on getting file descriptor from android content uri

aiueo13:
Fixes a crash that occurs when opening a file from an Android Content URI due to missing permissions or the file not existing.

# Details
Before the fix, Kotlin’s [getFileDescriptor](https://github.com/tauri-apps/plugins-workspace/blob/69a9d5771cc888b93ce71162b280b91408decc87/plugins/fs/android/src/main/java/FsPlugin.kt#L40) function would return `null` when the file was unavailable.

https://github.com/tauri-apps/plugins-workspace/blob/69a9d5771cc888b93ce71162b280b91408decc87/plugins/fs/android/src/main/java/FsPlugin.kt#L63-L67

However, on the Rust side that receives it, `null` triggers `unimplemented!()`, causing a panic.

https://github.com/tauri-apps/plugins-workspace/blob/69a9d5771cc888b93ce71162b280b91408decc87/plugins/fs/src/mobile.rs#L86-L93

And with this PR, the getFileDescriptor function will throw an error instead of returning null when the file is not available.

### DIFF
--- a/.changes/fix-fs-android-file-may-crash-for-content-uri.md
+++ b/.changes/fix-fs-android-file-may-crash-for-content-uri.md
@@ -1,0 +1,6 @@
+---
+fs: patch:bug
+fs-js: patch:bug
+---
+
+Fixes a crash that occurs when opening a file from an Android Content URI due to missing permissions or the file not existing.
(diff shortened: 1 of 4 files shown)


aiueo13:
https://github.com/tauri-apps/plugins-workspace/pull/3636

PR #3244, opened 2026-02-01
feat(fs): #3243 add encoding option to readTextFile functions

### LINKED ISSUE #3243: feat(fs): add encoding option to readTextFile (support non-UTF-8 files)
### Describe the problem

@tauri-apps/plugin-fs/readTextFile 
Currently, `fs.readTextFile` only works correctly with UTF-8 encoded text files.

On Windows (especially in Japanese environments), many text files are still encoded in legacy encodings such as Shift_JIS (CP932). When such files are opened with `readTextFile`, the content becomes garbled (mojibake), making the API unusable for these files.

This makes it difficult to build simple text editors or viewers with Tauri that need to handle existing non-UTF-8 text files.

It should be possible to specify the text encoding when reading a file, for example:

```ts
await readTextFile(path, { encoding: "shift_jis" });
```

### Alternatives considered

_No response_

### Additional context

1. Prepare a text file encoded in Shift_JIS (e.g. a typical Japanese `.txt` file created by Notepad).
2. Call `readTextFile(path)` from the Tauri fs plugin.
3. The returned string is garbled because it is decoded as UTF-8.

SeijiOkuda:
## Summary

Added support for reading text files encoded in character sets other than UTF-8 (e.g. Shift_JIS).

Previously, the application assumed UTF-8 encoding, which caused garbled text in some cases. With this change, multiple character encodings are now supported and such files can be read correctly.

Closes #3243 

## Changes

- Added support for reading files saved in encodings other than UTF-8  
- Added logic to specify and/or detect the character encoding and decode accordingly  
- Confirmed that existing behavior for UTF-8 files is not affected  

## Background / Motivation

- To support existing memo/text files that are saved in encodings such as Shift_JIS  
- To allow users to handle text files without character corruption  

## Verification

- Confirmed that UTF-8 files can still be read as before  
- Confirmed that files encoded in Shift_JIS and other encodings are displayed correctly  

## Usage

You can specify the encoding when reading a text file:

```ts
await readTextFile(path, { encoding: "shift_jis" });

### CODE REVIEW

File: plugins/fs/guest-js/index.ts [hunk 1]
```diff
@@ -722,7 +722,8 @@ async function readDir(
  */
 interface ReadFileOptions {
   /** Base directory for `path` */
-  baseDir?: BaseDirectory
+  baseDir?: BaseDirectory,
+  encoding?: string
```
amrbashir, line 727: Missing documentation

amrbashir (commented):
Can you please add a change file and also update `readTextFileLines` to do the same decoding

plugins/fs/guest-js/index.ts
SeijiOkuda: @amrbashir 
Thanks for the review!
Added documentation comments.

SeijiOkuda:
> Can you please add a change file and also update `readTextFileLines` to do the same decoding

@amrbashir 
Thanks for the review!
Done! I added the change file and updated `readTextFileLines` to apply the same decoding as well.

aiueo13:
`readTextFileLines` detects line breaks on the Rust side using a byte-level check of `\n` and `\r\n` for `0x0A` and `0x0D 0x0A`. Therefore, for encodings that are not compatible with this approach. it may be necessary to either return an error or change the implementation.😱

https://github.com/tauri-apps/plugins-workspace/blob/05045f9a7218fa023bf7260e497c1548682e5c96/plugins/fs/src/commands.rs#L1208-L1227

amrbashir (approved):

amrbashir:
I think that could be fixed in another PR

Issue #1470, opened 2024-06-16
How to get http response headers?


hengkx:
```js
// tauri-plugin-http = {version = "2.0.0-beta.9", features = ["cookies", "unsafe-headers"] }
import { fetch } from "@tauri-apps/plugin-http";
const response = await fetch();
console.log(response.headers.getSetCookie());// always []
```

4aiman:
I second this.
After installing the log plugin I can clearly see that cookies *are* being set:
![](https://github.com/tauri-apps/plugins-workspace/assets/2913210/d9fcc4fd-cc3e-4bc6-ad0d-4820ebd1a34e)

However, cookies in devtools are empty and show only localhost Tauri address 
(guess this is expected, as I've used the fetch from Rust backend to make the request)
![изображение](https://github.com/tauri-apps/plugins-workspace/assets/2913210/e3eb8d23-1e75-4eae-b0e2-2e710531a508)

I've tried iterating the headers, hoping to see a `set-cookie` header, but it's been stripped on the backend, it seems, as this code
```JS
  app.fetch = window.__TAURI__.http.fetch
  result1 = await app.fetch(url, request) // url = an url path; request is an instance of clientConfig 
  for (let pair of result1.headers.entries()) {
      console.log(pair)
  }
```
yields this
![изображение](https://github.com/tauri-apps/plugins-workspace/assets/2913210/ecaffa5f-0edf-40b4-893f-3564c31a9a2e)

amrbashir:
Since the `http` plugin operates using Rust and not the browser built-in fetch, cookies are stored on the Rust side and are not accessible through JavaScript normal cookie store.

amrbashir:
Could you explain a bit more about your use case and I will see if we can support it?

hengkx:
Some parameters are parsed through set cookies, and we need to obtain the value of set cookies to proceed with the next request

PR #3252, opened 2026-02-04
fix(http): correct Response header initialization to support cloning (fix #14892)

NVolcz:
Previously, headers were patched onto the Response object after construction, which bypassed the internal header list and caused `response.clone().headers` to be empty. This change passes the headers directly to the Response constructor, ensuring they are properly stored and clonable.


plugins/http/guest-js/index.ts
```diff
@@ -284,18 +284,16 @@ export async function fetch(
...
+  // constructor, like url.
   //
-  // Since url and headers are read only properties
-  // this is the only way to set them.
+  // Since url is a read only property
+  // this is the only way to set it.
   Object.defineProperty(res, 'url', { value: url })
-  Object.defineProperty(res, 'headers', {
-    value: new Headers(responseHeaders)
-  })
```
amrbashir, line 298: This was done in order to fix https://github.com/tauri-apps/plugins-workspace/issues/1470 , removing it here would make it regress

NVolcz: The headers are set in the Response constructor instead and will keep to be available through `response.headers`: https://github.com/tauri-apps/plugins-workspace/pull/3252/changes#diff-a65350a6e52e983dc2499033bc29a4177adc84fb0d7702fdef62f56306f5a010R288

amrbashir: The browser will still remove it, you can test the following in a chromium based browser:
```js
let res = new Response("Asd", {
  headers: new Headers({"X-Header": 123, "Set-Cookie": "hey!"})   
});
console.log(res.headers.getSetCookie()) //-> []
console.log(Array.from(res.headers)) // -> [['content-type', 'text/plain;charset=UTF-8'], ['x-header', '123']]
```
```js
let res = new Response("Asd");
Object.defineProperty(res, 'headers', {
  value: new Headers({"X-Header": 123, "Set-Cookie": "hey!"})
})
console.log(res.headers.getSetCookie()) // -> ['hey!']
console.log(Array.from(res.headers)) // -> [['set-cookie', 'hey!'], ['x-header', '123']]
```

You can see that in the first `set-cookie` is omitted and `getSetcookie()` will return an empty array.

NVolcz: Ah! Then I understand it better. Will look into it a bit more and see if there is a way to satisfy both use cases

NVolcz: I restored the previous code for setting the headers with `Object.defineProperty` and added code to override the clone for both url and headers instead. Also made the properties read-only as according to spec.

amrbashir (dismissed):

amrbashir (commented):
thank you, could you also add a change file in `.changes` directory?

amrbashir (approved):

PR #3265, opened 2026-02-08
fix(single-instance): recover as primary when mutex exists but IPC window is gone

goosewobbler:
When a previous instance crashes or is killed, its mutex may persist but the IPC window is destroyed by the OS. Currently, when a new instance detects ERROR_ALREADY_EXISTS but FindWindowW returns null, it falls through without registering as primary — leaving the app in a broken state with no IPC window and no mutex ownership.

Restructure init() so this case recovers by becoming the primary instance, creating the IPC window and managing the mutex normally.


Legend-Master (changes requested):
> When a previous instance crashes or is killed, its mutex may persist

Could you provide an example where this might happen? The mutex should be released once the creating process exits

goosewobbler:
@Legend-Master 

Windows-only: This recovery mechanism only affects Windows. Linux uses D-Bus service names and macOS uses Unix domain sockets, both of which are automatically released when the process terminates.

The issue: When a Windows process crashes or is killed (not gracefully terminated), the named mutex can briefly persist while the hidden IPC window is immediately destroyed by the OS.

Example scenario:
1. User opens your Tauri app → primary instance starts, creates mutex, creates IPC window
2. App crashes (segfault, OOM) OR user force-kills it via Task Manager → process terminates abnormally
3. Within seconds, user tries to reopen the app:
   - CreateMutexW returns existing mutex → ERROR_ALREADY_EXISTS
   - FindWindowW returns NULL (window was destroyed immediately)
   - Without this fix: The new instance falls through both branches → no mutex ownership, no IPC window, app runs in broken state
   - With this fix: Detects mutex exists but window gone → assumes previous owner died → becomes primary instance

Why the mutex persists:
Windows defers mutex cleanup in some cases (handle inheritance, debugging scenarios, or just async kernel cleanup). The key point is the window is definitely gone; so we know the previous instance is dead, regardless of mutex state.

Legend-Master:
> Windows defers mutex cleanup in some cases (handle inheritance, debugging scenarios, or just async kernel cleanup). The key point is the window is definitely gone; so we know the previous instance is dead, regardless of mutex state.

I have never encountered this before where it gets delayed, and even if say it happens, in such cases, the delay should be acceptable

The window can be killed by external programs as well, and should not be solely relied on

PR #3266, opened 2026-02-08
fix(single-instance): skip single-instance check for WebView2 child processes

goosewobbler:
WebView2 spawns child processes (renderer, gpu-process, etc.) with `--type=` arguments that re-enter `main()`. Without detection, these trigger the single-instance mutex check, causing incorrect IPC forwarding or premature exits.


Legend-Master (changes requested):
The plugin setup hook only runs for the tauri process, never the webview2 processes, could you explain the issue that you're facing a bit more?

goosewobbler:
@Legend-Master 
The setup hook does run in WebView2 child processes.  On Windows, WebView2 spawns separate processes (renderer, GPU process, etc.) by re-executing the same binary with `--type=renderer` style arguments. These child processes do re-enter main() and execute the full Tauri initialization, including plugin setup hooks.

The problem without this fix:
1. Child process starts → creates/locks mutex → ERROR_ALREADY_EXISTS
2. Without the `--type=` check, it attempts to:
   - Find the window (FindWindowW) which doesn't exist in child processes
   - Send WM_COPYDATA to a non-existent window
   - Call app.cleanup_before_exit() and process::exit(0), killing the renderer/GPU process

This causes either:
- Silent process termination (WebView2 fails to render anything)
- Incorrect IPC forwarding (child process trying to send messages as if it were a user launching a second instance)

The fix:
We detect `--type=` arguments early and skip the entire single-instance logic for these processes, allowing WebView2 child processes to start normally without interfering with the single-instance mechanism.

Legend-Master:
> On Windows, WebView2 spawns separate processes (renderer, GPU process, etc.) by re-executing the same binary with --type=renderer style arguments.

It should be spawning `msedgewebview2.exe`, not the tauri app executable, do you have a reproducible example where this happens?

Please fact check LLM outputs before opening PRs

PR #3270, opened 2026-02-10
fix(window-state): Window coordinates depend on the monitor’s scale_factor because logical pixels are being interpreted as physical pixels

teux:
Workaround for window shrinking on MacBook with external non-Retina (1x) monitor.

When the app closes, the window is on a 1x monitor where logical and physical pixels match. The coordinates are saved to disk with scale_factor=1.0.

On next launch, the Window object is created before the WebView exists. At this point, tao/macOS determines scale_factor from the default display (the built-in Retina screen with scale_factor=2.0), not from the monitor where the window will actually be positioned.

When saved coordinates are applied, they are treated as physical pixels on a 2x display and get divided by 2 to convert to logical pixels, causing the window to shrink to half its intended size.

This workaround uses the scale_factor of the target monitor (determined by saved coordinates) instead of the current window's scale_factor.

### DIFF
--- a/plugins/window-state/src/lib.rs
+++ b/plugins/window-state/src/lib.rs
@@ -87,6 +87,8 @@ struct WindowState {
     visible: bool,
     decorated: bool,
     fullscreen: bool,
+    #[serde(default)]
+    scale_factor: Option<f64>,
 }
 
 impl Default for WindowState {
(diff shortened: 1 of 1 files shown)


Legend-Master (changes requested):
This will not work when the window's physical position is returned correctly, please see the discussion in #2620

PR #3272, opened 2026-02-11
fix(fs): change to reject if unsupported encodings in readTextFileLines

aiueo13:
Change `readTextFileLines` to throw an error when an encoding is specified that cannot be handled by the existing newline detection logic.

This is related to <https://github.com/tauri-apps/plugins-workspace/pull/3244#issuecomment-3869568792>.


amrbashir:
Instead of rejection, I would rather change line detection on rust to read until the line separator according to the passed encoding.

aiueo13:
> Instead of rejection, I would rather change line detection on rust to read until the line separator according to the passed encoding.

Got it, thank you.

PR #3273, opened 2026-02-12
fix(fs): change to use correct line detection for encodings in `readTextFileLines`

aiueo13:
Previously, line detection assumed ASCII-compatible.
This change makes it use the line break representation for the passed encoding.

This is related https://github.com/tauri-apps/plugins-workspace/pull/3244#issuecomment-3869568792


.changes/fix-fs-line-detection-logic.md
```diff
@@ -0,0 +1,6 @@
+---
+"fs": patch
+"fs-js": patch
+---
+
+Fix an issue where the line detection logic in `readTextFileLines` does not work correctly for certain encodings.
```
amrbashir: No need to add a change file since the decoding feature is not released yet.

plugins/fs/src/commands.rs
```diff
@@ -1203,34 +1235,69 @@ impl StdFileResource {
...
+    }
+}
+
 struct StdLinesResource(Mutex<LinesBytes<BufReader<File>>>);
 
 impl StdLinesResource {
-    fn new(lines: BufReader<File>) -> Self {
-        Self(Mutex::new(LinesBytes(lines)))
+    fn new(lines: BufReader<File>, lf_bytes: Vec<u8>, cr_bytes: Vec<u8>) -> Self {
+        Self(Mutex::new(LinesBytes {
```
amrbashir: Let's add a `LinesBytes::new` method instead of creating the struct directly.

plugins/fs/src/commands.rs
```diff
@@ -437,12 +445,36 @@ pub fn read_text_file_lines<R: Runtime>(
...
 
     Ok(rid)
 }
 
+/// Returns the byte sequences for LF (`\n`) and CR (`\r`) in the encoding label.
+/// 
+/// <https://developer.mozilla.org/ja/docs/Web/API/Encoding_API/Encodings>
+fn lf_cr_bytes_for_encoding_label(label: Option<&str>) -> Result<(Vec<u8>, Vec<u8>), Error> {
+    // https://developer.mozilla.org/ja/docs/Web/API/TextDecoder/TextDecoder#label
+    let label = label.unwrap_or("utf-8");
```
amrbashir, line 467: instead of unwrap_or, we can just skip the rest of the function

```rs
let Some(label) = label else {
    return Ok((vec![b"\n"], vec![b"\r"]));
};

```

amrbashir: we could even go a little bit further and check if the passed encoding is `utf8` or `utf-8` and skip for these as well

```rs
if matches!(label, None | Some ("utf8") | Some("utf-8")) {
    return Ok((vec![b"\n"], vec![b"\r"]));
}

let label = label.unwrap();
```

plugins/fs/src/commands.rs
```diff
@@ -437,12 +445,36 @@ pub fn read_text_file_lines<R: Runtime>(
...
+    // So, we handle these cases directly.
+    if encoding == encoding_rs::UTF_16LE {
+        Ok((vec![0x0A, 0x00], vec![0x0D, 0x00]))
+    } else if encoding == encoding_rs::UTF_16BE {
+        Ok((vec![0x00, 0x0A], vec![0x00, 0x0D]))
+    } else {
+        let (lf, _, _) = encoding.encode("\n");
+        let (cr, _, _) = encoding.encode("\r");
+        Ok((lf.to_vec(), cr.to_vec()))
+    }
```
amrbashir: I think we can skip going through encoding_rs here and just return the bytes directly

```rs
Ok((vec![b"\n"], vec![b"\r"]))
```

afaik, only UTF-16 (LE or BE) and UTF-32 have different line feed and carriage return, aka 2 or 4 bytes, other encodings all use single byte `\n` and `\r` 

Please correct me if I am wrong on this one.

aiueo13: Yes. According to the Web Standard, the ASCII-incompatible encodings are UTF-16LE/BE and ISO-2022-JP. However, ISO-2022-JP can still detect line breaks in the same way as ASCII.

https://encoding.spec.whatwg.org/#security-background

plugins/fs/guest-js/index.ts
```diff
@@ -813,9 +813,15 @@ async function readTextFileLines(
 
     async next(): Promise<IteratorResult<string>> {
       if (this.rid === null) {
+        // Normalize the encoding label.
+        // If it is invalid, a RangeError is thrown here.
+        const encoding = options?.encoding != null
+          ? (new TextDecoder(options.encoding)).encoding
+          : undefined
```
amrbashir: We are creating another decoder down below to actually decode the bytes, let's create one here and re-move the other one.

plugins/fs/src/commands.rs
```diff
@@ -1362,10 +1439,14 @@ mod test {
             .lines()
             .map_while(Result::ok)
             .collect::<String>();
-        let string3 = LinesBytes(BufReader::new(bytes))
-            .flatten()
-            .flat_map(String::from_utf8)
-            .collect::<String>();
+        let string3 = LinesBytes {
```
amrbashir: use LineBytes::new

plugins/fs/src/commands.rs
```diff
@@ -1362,10 +1439,14 @@ mod test {
...
+            bytes: BufReader::new(bytes), 
+            lf_bytes: vec![b'\n'],
+            cr_bytes: vec![b'\r'] 
+        }
+        .flatten()
+        .flat_map(String::from_utf8)
+        .collect::<String>();
```
amrbashir: let's add tests for UTF16-LE/BE

amrbashir (commented):
You also need to run `pnpm build` to generate `plugins/fs/api-iife.js`

amrbashir (dismissed):
Thank you, this is awesome

amrbashir:
looks like you need to do `pnpm build` again?

aiueo13:
Oh😱 Thank you.

amrbashir (approved):
Thanks

aiueo13:
Thanks for the review

PR #3277, opened 2026-02-18
refactor(store): replace load_from_store with TryFrom impl in example

gzordrai:
Replace `load_from_store` with a `TryFrom` impl for more idiomatic conversion from `Store` to `AppSettings`.

Note: since all fields currently have default fallbacks, this conversion never actually fails. A `From` impl might be more appropriate here, but kept `TryFrom` to match the existing error handling.

### DIFF
--- a/plugins/store/examples/AppSettingsManager/src-tauri/src/app/settings.rs
+++ b/plugins/store/examples/AppSettingsManager/src-tauri/src/app/settings.rs
@@ -10,10 +10,8 @@ pub struct AppSettings {
     pub theme: String,
 }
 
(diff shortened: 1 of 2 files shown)


FabianLars:
> A From impl might be more appropriate here, but kept TryFrom to match the existing error handling.

I mean, it's just an example so you can change it if you prefer 🤷

gzordrai:
> > A From impl might be more appropriate here, but kept TryFrom to match the existing error handling.
> 
> I mean, it's just an example so you can change it if you prefer 🤷

Done

FabianLars (approved):

PR #3287, opened 2026-02-28
refactor(dialog): reuse `message` command for confirm

Legend-Master:


amrbashir (commented):
this is breaking because if users are not using `dialog:default` or don't have `dialog:allow-message` permission

if a breaking change is desired, then I'd suggest updating the change log to reflect that.

Otherwise, I'd suggest keeping `allow-ask`, `allow-confirm` and include `allow-message` as a nested command for them, as we did in fs plugin

Legend-Master:
Nice catch! Never thought about that possibility. Changed the confirm and ask permissions to be aliases of message permissions

amrbashir (approved):
That's clever, LGTM now

PR #3299, opened 2026-03-01
feat(updater): option to not restart after install

### LINKED ISSUE #3076: Updater on Windows: close and install - don't restart
On Windows. 
I'm finding myself in the need to run `update.install(bytes)` without wanting the installer to restart my app. 

**Use case**: I'm installing updates from `RunEvent::ExitRequested` so that I don't have to interrupt the user. 

The app checks and downloads the update in the background and then installs it when the user either clicks on a `Relaunch to apply update` button (and the app correctly restarts) OR when the user closes the app, where I don't want to have it restarted. 

Ideally there would be something like a `Update::install_without_restart` function or something to set on `Update`, and not so much something on `UpdateBuilder` as I don't know how the update will be installed when I "build" it. 

## Additional Context

I've seen a [`/R` flag](https://github.com/tauri-apps/plugins-workspace/blob/ea172bfa3cef8ba4021095a033d1a6f152c28c3e/plugins/updater/src/config.rs#L39-L40) is [passed to the NSIS installer](https://github.com/tauri-apps/plugins-workspace/blob/ea172bfa3cef8ba4021095a033d1a6f152c28c3e/plugins/updater/src/updater.rs#L765C21-L765C33), and a [`AUTOLAUNCHAPP=True` to MSI](https://github.com/tauri-apps/plugins-workspace/blob/ea172bfa3cef8ba4021095a033d1a6f152c28c3e/plugins/updater/src/updater.rs#L787).

Legend-Master:
Closes #3076


amrbashir:
how would this work on Windows, where you have to kill the app before applying the update, the app stays close? why would someone need that? 

I think if we already closed the app to apply updates, we should open it again

Legend-Master:
As mentioned in https://github.com/tauri-apps/plugins-workspace/issues/3076#issue-3566948352, it's used for installing updates on close, or if you registered a background task (like chrome's update task), it can be used for that as well

amrbashir:
yeah that's a valid case.

Prospector:
hey, is there a blocker or something this PR is waiting on? This is something a ton of our Windows users have been complaining about since we started pushing updates more regularly

Legend-Master:
Mainly just maintainer/reviewer time, it takes a lot of time to review and test changes for anything updater related

denjell-crabnebula:
Hey @lucasfernog can I ask you to prioritize the review and merge of this one?

Legend-Master:
It seems like the .msi updater couldn't launch the app with or without this change for me. I honestly couldn't remember if it worked before either, @FabianLars do you still remember it? (sorry I have a hard time understanding what https://github.com/tauri-apps/tauri/pull/10117 really did, would be super helpful if you could shed some light)


The command line flags are `/passive /promptrestart AUTOLAUNCHAPP=True LAUNCHAPPARGS=""` by the way, unchanged with or without this PR.

Legend-Master:
Oh never mind, I was trying to re-install the same version which triggered `NOT Installed`. Sorry!

plugins/updater/src/updater.rs
```diff
@@ -733,6 +761,16 @@ impl Update {
     fn install_inner(&self, _bytes: &[u8]) -> Result<()> {
         Ok(())
     }
+
+    /// If the Windows installer should restart the app after installed, default is `true`
```
```suggestion
    /// Whether the Windows installer should restart the app after installed, default is `true`
```

FabianLars (dismissed):

plugins/updater/src/updater.rs
Legend-Master: image

this button is so hard to click 😂

FabianLars (approved):

PR #3305, opened 2026-03-02
chore(deps): disable renovate bot on v1

Legend-Master:
We're not really actively working on v1 anymore, not sure if it's worth it to still keep updating its dependencies

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,6 +1,5 @@
 {
   "extends": ["config:recommended"],
-  "baseBranches": ["v2", "v1"],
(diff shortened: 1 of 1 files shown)


FabianLars:
i enabled it cause we were asked for it but in the last year it's been only prettier and eslint which is absoltely not worth it

FabianLars (approved):

FabianLars:
also be prepared for a bunch of pr spam. every time we touch the config it forgets which prs we closed 😂

PR #3335, opened 2026-03-08
fix(opener): strip UNC prefix before passing path to ILCreateFromPathW on Windows

### LINKED ISSUE #3304: tauri-plugin-opener bug - revealItemInDir fails for UNC network paths on Windows
### Describe the bug

revealItemInDir fails on Windows when given a UNC network path (e.g., \\server\share\file.mkv) with:

Failed to convert path '\\?\UNC\server\share\filename.ext' to ITEMIDLIST

Cause: Canonicalization in reveal_item_in_dir.rs (lines 13 and 51) converts UNC paths to \\?\UNC\... format. The \\?\ prefix is stripped for local drives but not for UNC paths. The prefixed path is then passed to ILCreateFromPathW, which doesn't support \\?\ prefixes and returns NULL.

I fixed this locally by stripping the \\?\UNC\ prefix back to \\ after canonicalization, before passing to ILCreateFromPathW. Affects both reveal_item_in_dir (line 13) and reveal_items_in_dir (line 51).

Environment: Plugin 2.5.3, Windows 11, SMB network paths

### Reproduction

_No response_

### Expected behavior

_No response_

### Full `tauri info` output

```text
Failed to convert path '\\?\UNC\server\share\filename.ext' to ITEMIDLIST
```

### Stack trace

```text

```

### Additional context

_No response_

arelove:
Fixes #3304

`dunce::canonicalize` converts UNC paths like `\\server\share\file.mkv`
to `\\?\UNC\server\share\file.mkv` on Windows. The `\\?\` prefix was
already being stripped for local paths, but not for UNC — so
`ILCreateFromPathW` was receiving an unsupported prefix and returning NULL.

## Fix

Added `normalize_path_for_shell()` in the Windows `imp` module to strip
`\\?\UNC\` and replace it with `\\` before passing to `ILCreateFromPathW`.

## Tests

Running unittests src\lib.rs (target\debug\deps\tauri_plugin_opener-88af2df7c99491c7.exe)

running 4 tests

```
test reveal_item_in_dir::tests::local_path_unchanged ... ok
test reveal_item_in_dir::tests::plain_unc_unchanged ... ok
test reveal_item_in_dir::tests::unc_path_converts_correctly ... ok
test reveal_item_in_dir::tests::unc_path_should_not_have_extended_prefix ... ok
test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

Note: the 4 failing doc-tests are pre-existing and unrelated to this change.``

### DIFF
--- a/plugins/opener/src/reveal_item_in_dir.rs
+++ b/plugins/opener/src/reveal_item_in_dir.rs
@@ -96,6 +96,17 @@ mod imp {
         },
     };
 
+    pub(super) fn normalize_path_for_shell(path: &Path) -> std::borrow::Cow<'_, Path> {
+        if let Some(s) = path.to_str() {
+            if let Some(unc) = s.strip_prefix(r"\\?\UNC\") {
+                return std::borrow::Cow::Owned(
+                    PathBuf::from(format!(r"\\{}", unc))
+                );
+            }
+        }
+        std::borrow::Cow::Borrowed(path)
+    }
+
     pub fn reveal_items_in_dir(paths: &[PathBuf]) -> crate::Result<()> {
         if paths.is_empty() {
             return Ok(());
@@ -166,11 +177,12 @@ mod imp {
 
     impl OwnedItemIdList {
         fn new(path: &Path) -> crate::Result<Self> {
-            let path_hstring = HSTRING::from(path);
+            let path = normalize_path_for_shell(path);
+            let path_hstring = HSTRING::from(path.as_ref());
             let item_id_list = unsafe { ILCreateFromPathW(&path_hstring) };
             if item_id_list.is_null() {
                 Err(crate::Error::FailedToConvertPathToItemIdList(
-                    path.to_owned(),
+                    path.to_path_buf(),
                 ))
             } else {
                 Ok(Self {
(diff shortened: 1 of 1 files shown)


plugins/opener/src/reveal_item_in_dir.rs
```diff
@@ -96,6 +96,15 @@ mod imp {
         },
     };
 
+    pub(super) fn normalize_path_for_shell(path: &Path) -> std::borrow::Cow<'_, Path> {
+        if let Some(s) = path.to_str() {
+            if let Some(unc) = s.strip_prefix(r"\\?\UNC\") {
+                return std::borrow::Cow::Owned(PathBuf::from(format!(r"\\{}", unc)));
```
Legend-Master, line 102: Genuine question, is it safe to always convert `\\?\UNC\` to `\\`?

arelove: Yes, it's safe specifically for ILCreateFromPathW since that Shell API doesn't support the \\?\ prefix at all. The one edge case worth noting: if the UNC path exceeds MAX_PATH (260 chars), stripping \\?\UNC\ could cause issues in other parts of the code that rely on extended-length path support. For the typical use case of revealing a file in Explorer this should be fine though.
But maybe I missed something, I'm not sure.

Legend-Master: Sounds reasonable, we could go with either this or use `absolute` instead of `canonicalize` (#3343)

PR #3339, opened 2026-03-09
fix: remove extra lambda brackets (fix #3338)

### LINKED ISSUE #3338: [android] Invoke NFC write without session do nothing
### Describe the bug

At NfcPlugin.kt:358 has extra lambda brackets 

 args.kind?.let { kind -> {
      ...
  }}

### Reproduction

_No response_

### Expected behavior

_No response_

### Full `tauri info` output

```text
It do nothing actually
```

### Stack trace

```text

```

### Additional context

_No response_

crazy-genius:


FabianLars:
Thanks for the PR! Since this is fairly small i'll make an exception but for following PRs please set up commit signing https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits as we have branch protection rules that require that :)

FabianLars (approved):

Issue #35, opened 2022-12-17
[log] Using Tauri Plugin Log with a Yew frontend


ClementGre:
Is there a way to use this plugin with yew, sending logs from rust in the frontend ?

Otherwise, is it a good  practice to send a message from the frontend to the backend, and then calling appropriate info!, warn! etc. functions like described in the documentation : https://tauri.app/fr/v1/guides/features/command.
This is the only alternative I've considered.

JonasKruckenberg:
Once `tauri-bindgen` is ready we will provide Rustwasm bindings to the log plugin too, but for the time being you gotta implement this functionality yourself. https://github.com/tauri-apps/tauri-bindgen/blob/main/examples/test/src/tauri_log.rs is an example of this (you just gotta switch out the actual sending logic for something using `invoke`)

ClementGre:
Thank's, then I don't have to use .wit files and I can just use a command in the backend:
```Rust
#[tauri::command]
fn send_log(message: str, type: u16) {
  // match type
  info!(message);
}
```
And from Yew, sending commands using invoke.

I achieved that with this code in the backend:
```Rust
use log::{debug, error, info, trace, warn};
use serde::{Serialize, Deserialize};

#[repr(usize)]
#[derive(Debug, Hash, Serialize, Deserialize)]
pub enum Level {
    Error,
    Warn,
    Info,
    Debug,
    Trace,
}

#[tauri::command]
pub fn log_from_front(message: &str, level: Level) {
    match level {
        Level::Error => error!("{}", message),
        Level::Warn => warn!("{}", message),
        Level::Info => info!("{}", message),
        Level::Debug => debug!("{}", message),
        Level::Trace => trace!("{}", message),
    }
}
```


And this code in the frontend:
```Rust
use serde::{Serialize, Deserialize};
use serde_wasm_bindgen::to_value;
use wasm_bindgen::{prelude::wasm_bindgen, JsValue};

#[repr(usize)]
#[derive(Debug, Hash, Serialize, Deserialize)]
pub enum Level {
    Error,
    Warn,
    Info,
    Debug,
    Trace,
}

#[wasm_bindgen]
extern "C" {
    #[wasm_bindgen(js_namespace = ["window", "__TAURI__", "tauri"])]
    fn invoke(cmd: &str, args: JsValue) -> JsValue;
}

#[derive(Serialize, Deserialize)]
struct LoggingMessage<'a> {
    message: &'a str,
    level: Level
}

[...]

PR #3351, opened 2026-03-19
feat(shell): add process group spawn option (fix #1332)

### LINKED ISSUE #1332: [shell] Add option to spawn command in process group

rterence:
## Summary

- Adds a `processGroup` boolean option to the shell plugin's `spawn` command
- When enabled, spawns the child in a new process group (POSIX) or job object (Windows) using the [`process-wrap`](https://github.com/watchexec/process-wrap) crate
- Killing the child process now kills the entire process tree, fixing orphaned grandchild processes (e.g. PyInstaller-wrapped apps)

Closes #1332
cc. @FabianLars 

## Changes

- **Rust** (`process/mod.rs`): `Command::set_process_group()` builder method. `CommandChild` now holds a `process_wrap::std::StdChildWrapper`, so both the plain and process-group spawn paths go through `process-wrap` uniformly — `process_group` is just an optional wrapper on that path. `kill()` delegates to `process-wrap`, reaping the whole process group (POSIX) or job object (Windows) instead of only the direct child.
- **Rust** (`commands.rs`): `processGroup` field on `CommandOptions`, plumbed through `prepare_cmd`
- **TypeScript** (`index.ts`): `processGroup?: boolean` on `SpawnOptions`
- **Example app** (`examples/api/src/views/Shell.svelte`): a "Process group" checkbox, plus the spawned pid in the output log, so the new option can be exercised from the API demo
- **Dependencies**: adds `process-wrap` 8.2 (`default-features = false`, features `std` / `process-group` / `creation-flags` / `job-object` / `tracing`) and `libc` 0.2 (unix, tests only). Removes `shared_child` 1 — superseded by `process-wrap`, which now handles child supervision on every platform.

## Testing

- All 12 unit tests pass, including 5 new process group tests
- PyInstaller simulation tests reproduce the exact issue scenario: without `processGroup` killing the wrapper orphans the grandchild; with `processGroup` killing the wrapper also kills the grandchild
- Windows not tested locally (uses `process-wrap` `JobObject` with polling `try_wait`)

## Demo

Recorded in the `examples/api` app's Shell view. The script backgrounds a long-running grandchild and prints its pid, reproducing the same orphaning shape as a PyInstaller bootloader:

```sh
sleep 300 & echo "grandchild=$!"; wait
```

A direct child is killed either way, so the grandchild pid is what matters — survival is checked from a terminal with `ps -p <pid>`.

**First half — "Process group" unchecked (the bug):**

1. Run the script, note the `grandchild=<pid>` line in the output pane
2. Press Kill
3. `ps -p <pid>` in the terminal **still lists the process** — the grandchild is orphaned and survives

**Second half — "Process group" checked (the fix):**

1. Run the script again, note the new grandchild pid
2. Press Kill
3. `ps -p <pid>` returns nothing — the **entire process tree is gone**

https://github.com/user-attachments/assets/0c25ad4d-b513-4d7b-a77d-7319f0da8893


Noremac11800:
Awesome additions @rterence! May I ask what's holding this up? This would go a long way to simplify one of my projects.

rterence:
> Awesome additions @rterence! May I ask what's holding this up? This would go a long way to simplify one of my projects.
@FabianLars can I get a review on this please?

FabianLars:
Sorry sorry. My backlog is a bit insane ATM. Will try to take a look this week.

plugins/shell/src/process/mod.rs
```diff
@@ -58,15 +58,79 @@ pub enum CommandEvent {
...
-#[derive(Debug)]
 pub struct CommandChild {
-    inner: Arc<SharedChild>,
+    inner: ChildKind,
     stdin_writer: PipeWriter,
 }
 
+enum ChildKind {
+    Direct(Arc<SharedChild>),
+    #[cfg(any(unix, windows))]
```
FabianLars: This covers everything, except wasm right? Might as well remove it then. Same for all other uses of the same cfg further down of course

plugins/shell/Cargo.toml
```diff
@@ -33,6 +33,10 @@ regex = "1"
 open = { version = "5", features = ["shellexecute-on-windows"] }
 encoding_rs = "0.8"
 os_pipe = "1"
+process-wrap = { version = "8.2", features = ["std"] }
```
FabianLars: do we need all default features?

plugins/shell/src/process/mod.rs
```diff
@@ -58,15 +58,79 @@ pub enum CommandEvent {
...
     cmd: StdCommand,
     raw_out: bool,
+    process_group: bool,
 }
 
 /// Spawned child process.
-#[derive(Debug)]
 pub struct CommandChild {
-    inner: Arc<SharedChild>,
+    inner: ChildKind,
```
FabianLars: I wonder if process-wrap supports spawning CommandWrap that did not add any wrappers? In that case i assume it should be possible to drop SharedChild completely in favor of CommandWrap? (Edit: Or rather, get rid of just the enum since we still use SharedChild on Unix here?) Then we'd just have to conditionally add the wrapper instead of having the whole child be the switch.

idk if i'm asking for too much here for this pr 😅

plugins/shell/Cargo.toml
FabianLars: Also fyi, we'll increase our rust version next minor (in tauri at least) so we can update to v9 _then_ here. Dunno if there are any changes that would help us

plugins/shell/src/process/mod.rs
```diff
@@ -58,15 +58,79 @@ pub enum CommandEvent {
...
+            Ok(())
+        } else {
+            Err(std::io::Error::last_os_error())
+        }
+    }
+}
+
+#[cfg(windows)]
+struct GroupChild {
+    inner: Arc<std::sync::Mutex<Box<dyn process_wrap::std::StdChildWrapper>>>,
```
FabianLars: if we have to keep SharedChild for unix, does it make sense to reuse it here? (If that's even possible)

FabianLars (commented):
didn't try it yet, just some messy thoughts before going to bed 😅

plugins/shell/src/process/mod.rs
rterence: Good catch. Removed the `#[cfg(any(unix, windows))]` gate now

plugins/shell/Cargo.toml
rterence: switched to `default-features = false` and enabled only what is used now

plugins/shell/src/process/mod.rs
rterence: Same constraint as below. The Windows child can't be reduced to a raw Child without losing tree-kill. Happy to tidy the spawn path in a follow-up leaving a TODO?

plugins/shell/src/process/mod.rs
rterence: On Windows the kill needs the Job Object handle inside the wrapper, and `into_inner()` discards it, so a raw Child would only kill the direct process. Unix can use SharedChild because its kill is just `killpg` on the pgid.

FabianLars: yeah makes sense, disregard my comment here. shouldn't have commented half asleep

plugins/shell/src/process/mod.rs
FabianLars: I can't follow my own text there lol so let me ask it differently. Can we drop SharedChild on Unix and use the same struct GroupChild as on Windows? If not, what is process-wrap's ProcessGroupChild missing here?

plugins/shell/src/process/mod.rs
```diff
@@ -58,15 +58,78 @@ pub enum CommandEvent {
...
+    fn kill(&self) -> std::io::Result<()> {
+        // SAFETY: killpg is a standard POSIX syscall. The pgid was obtained from
+        // the child's PID, which equals its PGID since it was spawned as a group leader.
+        let ret = unsafe { libc::killpg(self.pgid, libc::SIGKILL) };
+        if ret == 0 {
+            Ok(())
+        } else {
+            Err(std::io::Error::last_os_error())
+        }
+    }
```
FabianLars: opposed to https://docs.rs/process-wrap/latest/process_wrap/std/struct.ProcessGroupChild.html in https://github.com/watchexec/process-wrap/blob/main/src/std/process_group.rs#L169-L182 (kill = start_kill + wait) we do not wait for the children to exit here

plugins/shell/src/process/mod.rs
rterence: Addressed in e8d08018.  Dropped SharedChild and the ChildKind enum entirely, so there's now a single process-wrap child type for both platforms and process_group is just an optional wrapper. Lmk what you think, happy to tidy up the PR.

rterence: @FabianLars are you happy with this if I tidy up the PR and attach an updated demo?

rterence: @FabianLars if you get a moment could I get another review on this please? Or if this is being fixed elsewhere can you link me? Thanks 😄

rterence: @FabianLars any update on this?

FabianLars: sorryyyyyyyyy, having a hard time staying on top of everything lately.

yes, i think this approach looks good (assuming it works :D )

rterence: Hey @FabianLars. I've updated the PR desc to match the latest implementation and added a fresh demo vid from validating locally, all works as intended. Can I get an approval on this?

rterence: Hey @FabianLars. Just bumping this incase you have time.

FabianLars: i've been looking at it today actually. was thinking about how to get rid of the sleep loop but i think that's all too complicated. for this PR at least. i'll probably merge it as-is tomorrow.

rterence: Nice one! Sorry for the pestering, I'm keen for it to go in so that I can remove the workaround I have for an app.

rterence: @FabianLars did you get anywhere with the sleep loop?

rterence: Any update on this @FabianLars ?

FabianLars: not really, i did look a bit into it but got mad 😂 I still have one or two ideas but since my time's very limited as you've noticed already i'd look into that after your pr was already merged.

one thing that's left for this PR is a quick CPU usage comparison to see how "bad" the sleep loop is.

rterence: Got it, I'll do the CPU usage comparison this evening. Can you drop your ideas for the sleep loop. I'm happy to look into this once this PR is merged.

FabianLars: the only "idea" which i think is still realistic and hopefully maintainable is essentially how you started the PR. We go back to shared_child on all platforms and copy the logic of process-wrap into here. We saw this is possible on Linux since you already did that but not sure if it works on windows as well. I'd hate to have to copy/modify shared_child into our code as well...

rterence: Did the CPU comparison. 10 idle spawned children over 5s burns ~75ms CPU with the sleep loop vs ~11ms without, so roughly 0.15% of a core per child. Not terrible but its ~100 wakeups/sec per child for nothing.

The without number is because I also took a crack at removing the sleep loop entirely, without going back to shared_child. Instead of waiting through the lock, block on the raw process outside it with a wait that doesnt reap (waitid + WNOWAIT on unix, WaitForSingleObject on windows, same trick shared_child uses internally), then take the lock once and let process-wrap collect the status via try_wait. One subtlety, kill() reaps under the lock so the raw wait can lose that race and get ECHILD, which just means the status is already cached in the wrapper so its treated as success. Stress tested 900 spawn+kill cycles with no lost Terminated events. output() latency also drops from ~13ms to ~2ms since it doesnt eat a poll interval anymore.

Commit here: rterence/plugins-workspace@b6e68d7e. Happy to fold it into this PR or open as a follow up after this merges, whatever you prefer.

FabianLars: Interesting! I also looked into this idea but hit the wall on Windows cause I couldn't get a handle for the JobObject and thought the main process wouldn't be enough.

I see in the commit message that you tested it on macOS, did you also test it on Windows? If not i can try hopefully later today.

rterence: Yeah I tested on my mac, if you can test on windows that could be great!

rterence: @FabianLars did you manage to get a look at this?

FabianLars: Doesn't seem to work on Windows. It seemingly gets stuck on the `kill()` line.

rterence: Couldn't repro this on a windows-2025 runner. I tested both process-wrap alone with this branch's locking pattern and the plugin's actual spawn/kill on the commit you tested — kill returns in ~2ms every time ([runs](https://github.com/rterence/plugins-workspace/actions/runs/31625979843), [harness](https://github.com/rterence/plugins-workspace/tree/ci/windows-kill-repro)).

The only call on that path that can block forever is process-wrap's `kill()`, whose blocking `wait()` half runs with the child mutex held — on Windows that parks in `GetQueuedCompletionStatus(INFINITE)` on the job's completion port. Removed in rterence/plugins-workspace@e2586b99: `kill()` now does the non-blocking `start_kill()` under the lock, then waits on the raw process handle outside it and lets `try_wait` reap, same as the wait thread. The kill scenarios are `#[cfg(windows)]` tests now, so `cargo test -p tauri-plugin-shell` covers them in CI.

Also noticed the job object path was dropping CREATE_NO_WINDOW (process-wrap's CreationFlags/JobObject coordination is broken during spawn, watchexec/process-wrap#35), so process-group children flashed a console window. Fixed in rterence/plugins-workspace@9e6579d5.

Can you retest on the branch?

FabianLars: Hmm, i wonder i tested the correct commit 😅 i was in a bit of a hurry...

Now that i tested https://github.com/rterence/plugins-workspace/commit/b6e68d7e again it indeed seems to work in a contained test. No real world tests done yet.

Maybe we can focus on that one for now since https://github.com/rterence/plugins-workspace/commit/e2586b99 re-introduces a sleep loop (that ideally doesn't loop often if at all, but still) and adds even more code (i'm still trying to keep it minimal if possible). The testing looks a bit sloppy at first glance. I'd assume a powershell (or cmd) variant of your bash script would be enough for now? That's what i used mostly for testing tbh.

> Also noticed the job object path was dropping CREATE_NO_WINDOW (process-wrap's CreationFlags/JobObject coordination is broken during spawn, https://github.com/watchexec/process-wrap/issues/35), so process-group children flashed a console window. Fixed in https://github.com/rterence/plugins-workspace/commit/9e6579d5.

Good catch but damn, that's unfortunate. Hopefully they fix it soon. Anyway, good catch but i think you can use upstream's CreationFlags wrapper instead of a custom one. That should also accept `CREATE_SUSPENDED | CREATE_NO_WINDOW` as the input according to the readme. The important thing here is that this is called after the JobObject wrapper.

rterence: Done and pushed. Went with your plan:

- Dropped e2586b99 entirely — `kill()` is unchanged from the b6e68d7e you tested (now rterence/plugins-workspace@3b54fc40 on the branch). Happy to have another crack at getting rid of the wait-thread loop as a follow-up after this merges.
- rterence/plugins-workspace@c1f04fb2 — the CREATE_NO_WINDOW fix now uses upstream's `CreationFlags(CREATE_SUSPENDED | CREATE_NO_WINDOW)` registered after `JobObject`, custom wrapper deleted. One caveat I want on record: this only works *because* of watchexec/process-wrap#35. `get_wrap` can't see other wrappers during spawn, so `JobObject::wrap_child`'s "did the user ask for CREATE_SUSPENDED?" check misses our flag and still resumes the child. If they fix #35 in a semver-compatible release, that check starts seeing our CREATE_SUSPENDED and every job-object child stays suspended forever. There's a comment in `spawn` explaining this, and `test_cmd_process_group_output` runs on windows now so a bad bump fails CI instead of shipping. If you'd rather not depend on the bug at all I can go back to a tiny custom wrapper (distinct type, so `get_wrap` never finds it) — your call.
- rterence/plugins-workspace@3856ba03 — added `test/pyinstaller_sim.ps1` and lifted the unix-only gates on the process-group tests, so the windows job in test-rust exercises the real spawn/kill/job-object path instead of skipping it. Both pyinstaller tests also assert a `Terminated` event still arrives after `kill()`, since that's what apps actually depend on. The watchdog/diag harness stays out of the PR — that was throwaway to chase the hang report.

Green windows run of this branch (plus a scratch workflow commit to trigger it): https://github.com/rterence/plugins-workspace/actions/runs/31782370025

rterence: @FabianLars are you ok with this?

FabianLars: yep, looks good at first glance.

rterence: Any chance we can get this merged this weekend ?

plugins/shell/src/process/mod.rs
```diff
@@ -334,32 +346,52 @@ impl Command {
...
+                // CREATE_SUSPENDED, dropping CREATE_NO_WINDOW and flashing a
+                // console window. Registered last, our `pre_spawn` wins.
+                // CREATE_SUSPENDED must be kept: the child may only start
+                // running once `JobObject::wrap_child` has assigned it to the
+                // job, otherwise early grandchildren escape the job. The same
+                // bug makes `wrap_child`'s "user asked for CREATE_SUSPENDED"
+                // check miss this flag, so it still resumes the child. If
+                // process-wrap ever fixes #35, that check will start seeing
+                // our CREATE_SUSPENDED and leave the child suspended forever;
+                // `test_cmd_process_group_output` catches that on a bump.
```
FabianLars:
```suggestion
                // Intentionally using the wrong order as a workaround for <https://github.com/watchexec/process-wrap/issues/35>.
```

plugins/shell/src/process/mod.rs
```diff
@@ -334,32 +346,52 @@ impl Command {
...
-                }
-                Err(e) => {
-                    let _l = guard.write().unwrap();
-                    block_on_task(async move { tx.send(CommandEvent::Error(e.to_string())).await })
-                }
-            };
-        });
+        // Always go through process-wrap so the spawn path is uniform across
+        // platforms; the `process_group` switch is just an optional wrapper
+        // rather than a separate child type.
```
FabianLars:
```suggestion
```

.changes/shell-wait-no-poll.md
```diff
@@ -0,0 +1,5 @@
+---
+"shell": patch
+---
+
+Replace the polling loop in the child wait thread with a blocking wait on the process itself, removing a 10ms poll per spawned child and the corresponding latency on `Terminated` events.
```
FabianLars: Delete this file, it's relative to old changes in this PR, not related to the current dev/v2 branch.

.changes/shell-process-group.md
```diff
@@ -0,0 +1,6 @@
+---
+"shell": minor:feat
```
FabianLars:
```suggestion
"shell": minor
```

FabianLars (commented):
I'm fairly happy with the code now, thanks for sticking through it. Just a few docs changes as i don't like the overly verbose generated text.

rterence:
> I'm fairly happy with the code now, thanks for sticking through it. Just a few docs changes as i don't like the overly verbose generated text.

Made your requested updates, looks like there is an existing ci failure not related to this PR

FabianLars (dismissed):

rterence:
@FabianLars do you have any insight on this failing check?

FabianLars:
I didn't update all deps yet. Nothing for you to worry about in this pr.

FabianLars:
Okay so i have potentially bad news. My colleague is back from vacation and we've had a conversation about this PR.
1) We're concerned about the recent-ish development situation of process-wrap, i did not check it out in the recent months so i didn't notice.
2) Earlier in this PR my assumption was that we have to kill the JobObject in a special way and killing its parent wouldn't be enough which would make usage of shared_child impossible
3) shared_child's code is more complicated code compared to process-wrap, especially since rust std has process group support on unix nowadays which hopefully is enough for our use-case.

The consequence of this is that we agreed to either inline the JobObject logic into the plugin, or to use the seemingly more simple/sane https://github.com/ohadravid/win32job-rs (i think inlining is more reasonable here), and then return to shared_child if possible to handle the wait/kill problem again.

I know i put you already through much here so the question is, do you still want to work on this?

rterence:
> Okay so i have potentially bad news. My colleague is back from vacation and we've had a conversation about this PR.
> 
> 1. We're concerned about the recent-ish development situation of process-wrap, i did not check it out in the recent months so i didn't notice.
> 2. Earlier in this PR my assumption was that we have to kill the JobObject in a special way and killing its parent wouldn't be enough which would make usage of shared_child impossible
> 3. shared_child's code is more complicated code compared to process-wrap, especially since rust std has process group support on unix nowadays which hopefully is enough for our use-case.
> 
> The consequence of this is that we agreed to either inline the JobObject logic into the plugin, or to use the seemingly more simple/sane https://github.com/ohadravid/win32job-rs (i think inlining is more reasonable here), and then return to shared_child if possible to handle the wait/kill problem again.
> 
> I know i put you already through much here so the question is, do you still want to work on this?

Ok no problem, I feel like I'm committed to the cause now 😆 I'll take a look later today if I get a chance

rterence:
I'll drop process-wrap, go back to shared_child on all platforms with std process groups on unix, and inline the job object on windows. I'll try to get the windows behaviour to match unix as closely as possible.

rterence:
@FabianLars how's this?

rterence:
@FabianLars any update on this?

PR #3359, opened 2026-03-24
fix error that byte could be undefined

elwerene:
I get this error:

```
vite v7.3.1 building client environment for production...
transforming (3400) node_modules/@tabler/icons-vue/dist/esm/icons/IconLayout.mjsvendor/cargo/tauri-plugin-fs-2.4.5/guest-js/index.ts:272:10 - error TS18048: 'byte' is possibly 'undefined'.

272     x += byte
             ~~~~


Found 1 error in vendor/cargo/tauri-plugin-fs-2.4.5/guest-js/index.ts:277
```

I don't know why typescript reports this error, but here is an easy fix for it

### DIFF
--- a/plugins/fs/guest-js/index.ts
+++ b/plugins/fs/guest-js/index.ts
@@ -280,7 +280,7 @@ function fromBytes(buffer: FixedSizeArray<number, 8>): number {
   let x = 0
   for (let i = 0; i < size; i++) {
     // eslint-disable-next-line security/detect-object-injection
-    const byte = bytes[i]
+    const byte = bytes[i] ?? 0
     x *= 0x100
     x += byte
   }


Legend-Master:
I don't think you're supposed to type check library code here, try adding `skipLibCheck` to your tsconfig.json

> https://www.typescriptlang.org/tsconfig/#skipLibCheck

We should probably enable [`noUncheckedIndexedAccess`](https://www.typescriptlang.org/tsconfig/#noUncheckedIndexedAccess) in this repo though

Thanks for the PR anyways!

elwerene:
@Legend-Master  `skipLibCheck` somehow does not help here. I'm not a Typescript expert, do you have an idea what I can use to fix this? At the moment I'm using my fork of plugins-workspace to produce releases

Legend-Master:
The error looks like it's from vite but vite usually wouldn't do the type checking at all so I'm not sure 
I'll need more information on the build process to say where the problem is

PR #3368, opened 2026-03-29
feat(opener): `open_dir` fn to open dir directly in explorer

wovw:
Added `open_dir` and `openDir` with similar logic and capabilities to `reveal_item_in_dir` for opening a dir directly in the default explorer.


Legend-Master:
Doesn't `openPath` already cover this?

wovw:
`openPath` relies on running the `xdg-open` binary. My default file manager is `yazi`, a terminal application, so `openPath` fails silently because the Tauri app has no TTY attached for the terminal app to render to. `openDir` uses standard D-Bus calls like `revealItemInDir`, so a request goes directly to the desktop environment which handles launching the relevant `.desktop` file like nautilus. (On Windows and macOS `openPath` is fine).

Legend-Master:
Well, we can extend `openPath` in that case, a new dedicated command here doesn't really make much sense

And if you think it fits, you can open a PR upstream at https://github.com/Byron/open-rs so this could benefit more than just tauri

wovw:
I decided not to upstream to `open-rs` for now since it's outside the scope of that crate, and it would introduce another dependency.

PR #3374, opened 2026-04-01
fix(deep-link): remove quotes from Exec path in .desktop file on Linux (fix tauri-apps/tauri#10570)

### LINKED ISSUE #10570: [bug] Deep links doesn't open app on linux
### Describe the bug

Tauri deep links doesn't work on Linux on v2 project. The handler is not called.

Eg. the link `vibe://download/?url=https://huggingface.co/ggerganov/whisper.cpp/resolve/main/ggml-tiny.bin?download=true`

### Reproduction

Use deep link plugin in v2 project on Linux

### Expected behavior

It should open the app with the parameters

### Full `tauri info` output

```text
tauri beta.v2.25
```

### Stack trace

_No response_

### Additional context

https://github.com/thewh1teagle/vibe/issues/212

0xPxt:
The `register()` function wraps the `Exec` path in literal double quotes when generating `.desktop` files on Linux:

```rust
let qualified_exec = format!("\"{}\" %u", exec);
// produces: Exec="/usr/bin/kunobi" %u
```

Per the [freedesktop Desktop Entry Spec](https://specifications.freedesktop.org/desktop-entry-spec/latest/), `Exec` values are not shell-parsed — the quotes become part of the binary path. When `xdg-open` handles a deep link URL, it extracts the first token (including the literal `"` characters) and fails to find it:

```
$ XDG_UTILS_DEBUG_LEVEL=2 xdg-open "myapp://callback"
/usr/bin/xdg-open: 811: : Permission denied
```

Removing the quotes fixes deep link delivery on Linux.

Intended to resolve https://github.com/tauri-apps/tauri/issues/10570

### DIFF
--- a/.changes/fix-deep-link-linux-exec-quoting.md
+++ b/.changes/fix-deep-link-linux-exec-quoting.md
@@ -1,0 +1,6 @@
+---
+"deep-link": patch
+"deep-link-js": patch
+---
+
+Remove literal quotes around `Exec` path in generated `.desktop` file on Linux, fixing deep link registration via `xdg-open`.
--- a/plugins/deep-link/src/lib.rs
+++ b/plugins/deep-link/src/lib.rs
@@ -293,7 +293,7 @@ mod imp {
                     .unwrap_or_else(|| bin.into_os_string())
                     .to_string_lossy()
                     .to_string();
-                let qualified_exec = format!("\"{}\" %u", exec);
+                let qualified_exec = format!("{} %u", exec);
 
                 let target = self.app.path().data_dir()?.join("applications");
 


FabianLars:
> Intended to resolve https://github.com/tauri-apps/tauri/issues/10570

I doubt it helps. The quotes were added a month ago. This issue is 2 years old.

Now you're not the first one (unless you were the one on discord) to report having issues with the quotes but they were added because we had issues with paths with spaces.
[`desktop-file-validate`](https://manpages.debian.org/trixie/desktop-file-utils/desktop-file-validate.1.en.html) also does not complain about the quotes.
The [linked spec](https://specifications.freedesktop.org/desktop-entry/latest/exec-variables.html) also says the following
> Arguments may be quoted in whole. If an argument contains a reserved character the argument must be quoted. The rules for quoting of arguments is also applicable to the executable name or path of the executable program as provided. 

Looks like this is a bug in xdg-open https://gitlab.freedesktop.org/xdg/xdg-utils/-/work_items/151 - we should double check that spaces are broken (i'll try to get the repro that prompted us to add the quotes) and then maybe set the quotes only if we detect quotes.

PR #3377, opened 2026-04-03
feat(geolocation): support for non-GMS devices

### LINKED ISSUE #2115: [feat][Geolocation] Add support for getCurrentPosition via Android Framework Location API
Geolocation plugin implements Google Play Services Location API version only, not implementing Android Framework Location API.
It becomes a problem for those users who have no GMS supports.

cfpwastaken:
Closes #2115 
I also fixed an issue where the `watchPosition()` command wouldn't actually resolve (and thus never returning a channel id) and added the watchPosition/clearWatch commands to the example app.
I can split these changes off if needed.


tomkel:
Thank you for starting this, I don't use Google Play Services and this has been blocking me from adopting Tauri.

Legend-Master:
Seem like all the APIs used here will require API level 31? We'll need to guard them since the plugin still wants to support older devices.

cfpwastaken:
Good point. Will look into that.
API Level 31 is for Android 12, 2021. Makes sense to guard that.
I'm also just noticing that the CI is screaming at the changeset file.

.changes/geolocation-non-gms.md
```diff
@@ -0,0 +1,5 @@
+---
+"geolocation": "minor"
```
FabianLars, line 2:
```suggestion
"geolocation": "minor"
"geolocation-js": "minor"
```
that's why ci is complaining btw

cfpwastaken:
Is this gonna be a problem?
image

FabianLars:
no, all good

cfpwastaken:
That should do it.

Issue #3328, opened 2026-03-04
plugin-barcode-scanner iOS crash: "io error: No route to host (os error 65)"


ildella:
### Describe the bug

The barcode plugin is working fine in my Sveltekit2 app on Android since months. 
Never an issue. 

All dependencies of any kind are the latest. 

Trying now in iOS it crashes "badly", with this error:
iOS I'm trying on is an iphone SE3, but the same happend on other users iphones too. 

`io error: No route to host (os error 65)`

That error is catched by some "global error cateher" when I do not try to "catch" the error in the Svelte component that runs it. 
But, when I run the code down there, the app just crashes and no error is reported and the "signal" inside the catch are *never* called. 
It just vanishes the whole app. 

To be more clear: 
1. the app goes in scan mode as soon as the onMount calls the "scan()" function. 
2. if I focus on the QR code --> crashes badly
3. in a previous version of the code I had a cancel button that was calling the plugin "cancel()" function. That was also crashing the whole app badly in the same way, with the same error. 

This is the code. I always get to `scanner_permission_granted` but I never see `scanner_scan_completed` - but that's ok. I also never see the signal sent in the "catch" and not even the one in "finally". The app crashes in Rust code and that is it. 

```js
  onMount(async () => {
    setTransparency('add')
    const granted = await ensurePermissions()
    if (!granted) {
      error = {message: 'Camera permission denied'}
      signal('scanner_permission_refused')
      setTimeout(onCancel, DELAY_BEFORE_CANCEL)
      return
    }
    signal('scanner_permission_granted')
    try {
      const result = await scan({
        windowed: true,
        formats: [Format.QRCode],
      })
      signal('scanner_scan_completed')
      if (!result.content)
        throw new Error('Scanner returned empty result')
      const {content} = result
      signal('scanner_scan_succesfull', {content})
      onResult(content)
    } catch (error) {
      // console.error(error)
      signal('scanner_scan_failed', {})
      const message = getErrorMessage(error)
      signal('scanner_scan_error', {message})
    } finally {
      signal('scanner_scan_finished', {})
    }
  })
```

Bundle config

```json5
  bundle: {
    iOS: {
      infoPlist: './Info.ios.plist',
      bundleVersion: '22',
    },
  },
```

and full Info.ios.plist

```xml
<?xml version="1.0" encoding="UTF-8"?>
[...]

Issue #3081, opened 2025-11-05
Barcode scanner crashes app on iOS if `cancel` is called while no longer reading


simonwst:
Tested on iPhone 14 (physical device) with iOS 18.5 and barcode-scanner plugin `v2.4.2` / tauri `v2.9.2`. After calling `scan`, waiting for the result and then calling `cancel` the whole app freezes / the camera output turns black. Looks like the plugin tries to release a camera that is no longer attached / accessed causing it to crash and the overlay for the camera to turn black and block the UI.

niko-ng:
Fixed in PR #3107

PR #3393, opened 2026-04-17
fix(barcode-scanner): dispatch iOS cancel() cleanup to the main thread

AlexisZankowitch:
## Summary

Wraps the body of `BarcodeScannerPlugin.cancel(_:)` on iOS in `DispatchQueue.main.async` so the UIKit teardown runs on the main thread.

## Context — please review carefully ⚠️

I'm not an iOS developer. I hit this crash while building a Tauri iOS app and worked through the diagnosis and fix with Claude Opus 4.7. I want to be upfront about my limited native iOS experience so a maintainer familiar with the plugin internals can validate the approach before merging. This code fixed the issue I was facing, so I wanted to share it here. 

### The fix the agent proposed
Wrap the body of `cancel(_:)` in `DispatchQueue.main.async`, mirroring the pattern already used by [`scan()`](https://github.com/tauri-apps/plugins-workspace/blob/9228419e1327c6642a26478826dd9b3500d140cd/plugins/barcode-scanner/ios/Sources/BarcodeScannerPlugin.swift#L328) and [`openAppSettings()`](https://github.com/tauri-apps/plugins-workspace/blob/9228419e1327c6642a26478826dd9b3500d140cd/plugins/barcode-scanner/ios/Sources/BarcodeScannerPlugin.swift#L252)
```swift 
@objc private func cancel(_ invoke: Invoke) {
  DispatchQueue.main.async { [self] in
    self.invoke?.reject("cancelled")
    self.destroy()
    invoke.resolve()
  }
}
```
_⬇️ Methodology explained in [my comment below](https://github.com/tauri-apps/plugins-workspace/pull/3393#issuecomment-4266655971) ⬇️_ 

## How I validated the fix
- Pointed my Tauri app at a local path-dep clone of this repo with the patch applied.
- Rebuilt and deployed to the same iPhone 13 / iOS 26.4.
- Tapped scan → camera opened in windowed mode.
- Tapped my cancel button → the scanner closed cleanly, no crash. The pending scan promise in JS rejected with "cancelled" as expected.
- Re-tested the happy path: scan → point at a real book barcode → decoded correctly, webView opacity/background restored normally.
- ⚠️ I would  probably need someone else to validate this fix is working fine. 

## Potential related issues
Issues that could be related, please let me know if I should mention them in the PR title.
- #3081 (however this one has been already closed)
- #3328 

## Final notes 
I reviewed the best I could what the agent proposed and ran some tests and it seems to be working fine for me. That's why I wanted to propose the fix to you guys as well. But I'm perfectly fine if you guys want to take those changes and push it from another PR if you think it's more appropriate. Hopefully I didn't fail your [AI policy](https://github.com/tauri-apps/tauri/blob/dev/.github/CONTRIBUTING.md#ai-tool-policy). If so I apologize and can delete my PR 😅


AlexisZankowitch:
## Methodology used to diagnose and "fix" the issue

1. In my app, I added the plugin @tauri-apps/plugin-barcode-scanner and wired a scan icon to my svelte kit app. I then tried it on my actual iPhone (iPhone 13 / iOS 26.4). I am using `windowed: true` so I could render my own UI (including a cancel button). Tapping the scan icon opened the camera fine and scanning a real barcode worked perfectly. But when I added a cancel button that calls `cancel()` from JS to close the scanner without scanning anything, tapping it reliably **crashed the app** every time.

2. I managed to find some crash report (`.ips`) from the phone's using the `Open Recent Logs` button of the `Devices and Simulators` window from Xcode. Where I saw some: `Crash Reporter Key` and other words that made me think I could give that file to the LLM to see if it can make sense of it.

3. **I fed the crash report to the agent** and asked it to help diagnose. **From that point on, it led the analysis.**  

**From my crash report:**
```
Last Exception Backtrace:
0   CoreFoundation                	       0x193103c70 __exceptionPreprocess + 164
1   libobjc.A.dylib               	       0x18fbd9224 objc_exception_throw + 87
2   Foundation                    	       0x190c39f50 -[NSAssertionHandler handleFailureInMethod:object:file:lineNumber:description:] + 287
3   UIKitCore                     	       0x198c38198 -[UIView(Internal) _didMoveFromWindow:toWindow:] + 2103
4   UIKitCore                     	       0x198bd55f0 __45-[UIView(Hierarchy) _postMovedFromSuperview:]_block_invoke + 131
5   UIKitCore                     	       0x198bd54f8 -[UIView _postMovedFromSuperview:] + 507
6   UIKitCore                     	       0x198be3f84 __UIViewWasRemovedFromSuperview + 135
7   UIKitCore                     	       0x198be4eb4 -[UIView(Hierarchy) removeFromSuperview] + 255
8   <app>                        	       0x102e97fe0 BarcodeScannerPlugin.dismantleCamera() + 836
9   <app>                        	       0x102e98124 BarcodeScannerPlugin.destroy() + 76
10  <app>                        	       0x102e9ae5c BarcodeScannerPlugin.cancel(_:) + 240
11  <app>                        	       0x102e9aedc @objc BarcodeScannerPlugin.cancel(_:) + 68 //  ⬅️Apparently this is the sweat sugar (from me)
12  <app>                        	       0x102e8a558 closure #1 in PluginManager.invoke(name:invoke:) + 1932
13  <app>                       	       0x102e8ab0c thunk for @escaping @callee_guaranteed @Sendable () -> () + 48
14  libdispatch.dylib             	       0x1cd22e9a8 _dispatch_call_block_and_release + 31
15  libdispatch.dylib             	       0x1cd2481e4 _dispatch_client_callout + 15
16  libdispatch.dylib             	       0x1cd236fb0 _dispatch_lane_serial_drain + 739
17  libdispatch.dylib             	       0x1cd237aac _dispatch_lane_invoke + 391
18  libdispatch.dylib             	       0x1cd241dac _dispatch_root_queue_drain_deferred_wlh + 283
19  libdispatch.dylib             	       0x1cd2416ac _dispatch_workloop_worker_thread + 719
20  libsystem_pthread.dylib       	       0x1f1e303b0 _pthread_wqthread + 291
21  libsystem_pthread.dylib       	       0x1f1e2f8c0 start_wqthread + 7
```

### Diagnosis from Claude Opus 4.7
Looking at `BarcodeScannerPlugin.swift`, the `cancel()` handler synchronously mutates UIViews on whichever thread Tauri happens to dispatch plugin commands on — in this case thread 12 with `"queue": "ipc"`, not `com.apple.main-thread`. UIKit enforces main- thread-only mutations with an internal assertion that aborts the process with SIGABRT:

```
┌─  UIKit assertion fires here
│   -[NSAssertionHandler handleFailureInMethod:...]
│   -[UIView(Internal) _didMoveFromWindow:toWindow:]
│   -[UIView(Hierarchy) removeFromSuperview]     ← UIKit call off-main
│
│   BarcodeScannerPlugin.dismantleCamera()
│   BarcodeScannerPlugin.destroy()
│   BarcodeScannerPlugin.cancel(_:)              ← our handler
│   @objc BarcodeScannerPlugin.cancel(_:)
│
│   closure #1 in PluginManager.invoke(name:invoke:)
│   _dispatch_call_block_and_release
│   _dispatch_lane_serial_drain
└─  _dispatch_workloop_worker_thread              ← background IPC queue
```

The other commands in the same file (`scan()`, `openAppSettings()`) already wrap their UIKit work in `DispatchQueue.main.async`. The `cancel()` handler doesn't — that looks like the bug.

oscarardevol:
Confirming this fix works.
     
  Without it, calling `cancel()` from JS in windowed mode crashes the app on iPhone 12 Pro with iOS 26.5:
  
  ```
Assertion failure in <private>, UIView.m:22709
  Modifying properties of a view's layer off the main thread is not allowed
  ```
  
  I'm vendoring 2.4.4 locally with the same patch and the crash is gone. Happy to drop the vendor once this lands.
  
  **Environment**
  
  - `tauri` `2.11.2`
  - `tauri-plugin-barcode-scanner` `2.4.4`

plugins/barcode-scanner/ios/Sources/BarcodeScannerPlugin.swift
```diff
@@ -337,10 +337,11 @@ class BarcodeScannerPlugin: Plugin, AVCaptureMetadataOutputObjectsDelegate {
   }
 
   @objc private func cancel(_ invoke: Invoke) {
-    self.invoke?.reject("cancelled")
-
-    destroy()
-    invoke.resolve()
+    DispatchQueue.main.async { [self] in
+      self.invoke?.reject("cancelled")
```
FabianLars, line 341: i wonder if the invoke call should stay outside but probably doesn't make a difference 🤔

FabianLars (approved):

plugins/barcode-scanner/ios/Sources/BarcodeScannerPlugin.swift
oscarardevol: Tested both placements locally on iPhone 12 Pro / iOS 26.5:

  - `invoke.resolve()` inside the closure: works, no crash, scanner dismisses cleanly.
  - `invoke.resolve()` outside the closure (as you suggested): also works, no crash, scanner dismisses cleanly. No visible difference.
  
  Pulling it out also feels right and keeps main-thread work to a minimum. Though I guess in this case the difference is imperceptible.

FabianLars: thanks for testing! i'll keep it as is for now but will keep this in mind for a later pr.

PR #3398, opened 2026-04-18
[Window State] implement fullscreen & minimized logic

polarove:
1. Add new member `StateFlags::MINIMIZED` to record window.is_minimized() since we have [StateFlag::MAXIMIZED](https://github.com/tauri-apps/plugins-workspace/blob/c1fd33b3a2735f2e25c1d026dc524af932db3315/plugins/window-state/src/lib.rs#L55)

2. Implement [StatesFlags::FULLSCREEN](https://github.com/tauri-apps/plugins-workspace/blob/c1fd33b3a2735f2e25c1d026dc524af932db3315/plugins/window-state/src/lib.rs#L58), records & recover its state accordingly

### DIFF
--- a/plugins/window-state/src/lib.rs
+++ b/plugins/window-state/src/lib.rs
@@ -56,6 +56,7 @@ bitflags! {
         const VISIBLE     = 1 << 3;
         const DECORATIONS = 1 << 4;
         const FULLSCREEN  = 1 << 5;
+        const MINIMIZED   = 1 << 6;
     }
 }
 
@@ -84,6 +85,7 @@ struct WindowState {
     prev_x: i32,
     prev_y: i32,
     maximized: bool,
+    minimized: bool,
     visible: bool,
     decorated: bool,
     fullscreen: bool,
@@ -99,6 +101,7 @@ impl Default for WindowState {
             prev_x: Default::default(),
             prev_y: Default::default(),
             maximized: Default::default(),
+            minimized: Default::default(),
             visible: true,
             decorated: true,
             fullscreen: Default::default(),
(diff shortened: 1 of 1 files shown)


plugins/window-state/src/lib.rs
```diff
@@ -216,12 +219,13 @@ impl<R: Runtime> WindowExt for Window<R> {
                 })?;
             }
 
-            if flags.contains(StateFlags::MAXIMIZED) && state.maximized {
+            // fullscreen > maximized > minimized
```
Legend-Master, line 222: These 3 can co-exist.

plugins/window-state/src/lib.rs
```diff
@@ -256,9 +256,23 @@ impl<R: Runtime> WindowExt for Window<R> {
                 metadata.fullscreen = self.is_fullscreen()?;
```
Legend-Master, line 256: Don't we already have fullscreen implemented?

plugins/window-state/src/lib.rs
```diff
@@ -280,18 +294,33 @@ impl<R: Runtime> WindowExtInternal for WebviewWindow<R> {
...
+            | StateFlags::SIZE
+            | StateFlags::FULLSCREEN;
+
+        let is_concern = flags.intersects(concern_flags);
+
+        let is_maximized = is_concern && self.is_maximized()?;
 
+        let is_minimized = is_concern && self.is_minimized()?;
+
+        let is_fullscreen = is_concern && self.is_fullscreen()?;
```
Legend-Master, line 309: I don't think this change is correct

Legend-Master (commented):
Also, some of your commits are not signed + your seem not to have the signing public key uploaded to GitHub yet.

PR #3406, opened 2026-04-26
fix(upload): wrong progress callback parameter names in README

Zhoucheng133:
The current documentation/examples for the download and upload functions use an incorrect destructuring pattern in the progress callback.

Currently, the examples suggest using ({ progress, total }), but according to the [ProgressPayload](https://github.com/tauri-apps/plugins-workspace/blob/c463d8ab1422a4cf44f627a7b4e6ba9d3553f334/plugins/upload/guest-js/index.ts#L9) interface in the source code, progress represents the chunk size (delta), while progressTotal represents the accumulated downloaded bytes.

### DIFF
--- a/plugins/upload/README.md
+++ b/plugins/upload/README.md
@@ -66,7 +66,7 @@ import { upload, HttpMethod } from '@tauri-apps/plugin-upload'
 upload(
   'https://example.com/file-upload',
   './path/to/my/file.txt',
-  (progress, total) => console.log(`Uploaded ${progress} of ${total} bytes`), // a callback that will be called with the upload progress
+  (progressTotal, total) => console.log(`Uploaded ${progressTotal} of ${total} bytes`), // a callback that will be called with the upload progress
   { 'Content-Type': 'text/plain' } // optional headers to send with the request
 )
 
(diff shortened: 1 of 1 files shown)


plugins/upload/README.md
```diff
@@ -66,15 +66,17 @@ import { upload, HttpMethod } from '@tauri-apps/plugin-upload'
 upload(
   'https://example.com/file-upload',
   './path/to/my/file.txt',
-  (progress, total) => console.log(`Uploaded ${progress} of ${total} bytes`), // a callback that will be called with the upload progress
+  (progressTotal, total) =>
```
Legend-Master: shouldn't this be

```suggestion
  ({ progressTotal, total }) =>
```

Legend-Master (changes requested):
Looks like examples are all outdated, let's also fix the header params

plugins/upload/README.md
Zhoucheng133: Oops, I missed the destructuring braces. Thanks for catching that!

Legend-Master (approved):
Thanks, and you'll need to sign your commits for me to merge this

> https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

Zhoucheng133:
@Legend-Master I've re-signed all the commits! They should all be verified now.

PR #3419, opened 2026-05-11
fix(nfc): adapt backend invoke to frontend

### LINKED ISSUE #2272: Invoking an NFC scan from Rust causes serialization errors
For the NFC plugin there appears to be a mismatch in the messaging between the Kotlin code and the Rust interface. The Rust side wants to receive a ScanResponse, but the Kotlin code replies back with only a NfcTag type object.
There is also a mismatch in the NfcTag struct. The id should probably be a Vec<u8> and the kind should (I think) be a Vec<String>.
Finally, it looks like Java's byte is signed, so that causes further problems for deserializing into a u8.

I don't have an iOS device so I'm not sure if any of these issues occur there.

pascalauroboa:
Should resolve #2272. There is a mismatch between the frontend and backend API, as well as the native implementations. Javascript works but Rust does not. This should change the API to match the working frontend.


FabianLars:
thanks for the PR! Unfortunately, changing the function signature and remove a pub struct are both breaking changes so this has to wait for v3. until then, does this work?
```rs
    pub fn scan(&self, payload: ScanRequest) -> crate::Result<ScanResponse> {
        self.0
            .run_mobile_plugin("scan", payload)
            .map(|v| ScanResponse { tag: v })
            .map_err(Into::into)
    }
```

Lastly, please also add a changefile similar to https://raw.githubusercontent.com/tauri-apps/plugins-workspace/refs/heads/v2/.changes/material-dialog.md , thanks :)

.changes/nfc.md
```diff
@@ -0,0 +1,5 @@
+---
+nfc: fix
```
FabianLars:
```suggestion
nfc: patch
nfc-js: patch
```

FabianLars (approved):

Issue #1583, opened 2024-07-24
Tauri-plugin-positioner panics whenever I try to use Position::Tray*


Bamboooz:
I wanted to use the Tray::BottomRight in my application, but whenever I try to use it and start my application, it throws a panic message:

```
thread 'main' panicked at C:\Users\Bambu\.cargo\registry\src\index.crates.io-6f17d22bba15001f\tauri-plugin-positioner-1.0.5\src\ext.rs:198:21:
Tray position not set
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[0724/101207.681:ERROR:window_impl.cc(121)] Failed to unregister class Chrome_WidgetWin_0. Error = 1412
```

I have the system-tray feature included, I also found the same error in an issue from 2 years ago, which had a link to a fix, but it is no longer available: https://github.com/tauri-apps/plugins-workspace/issues/173

This is how I call the tray positioning
```rs
fn initialize(app: &mut App) -> Result<(), Box<dyn std::error::Error>> {
  let window = app.get_window("main").unwrap();

  let _ = set_shadow(&window, true);
  let _ = window.move_window(Position::TrayBottomRight);

  Ok(())
}
```

And the initialize is called in the tauri builder default setup function

FabianLars:
> And the initialize is called in the tauri builder default setup function

The tray position apis currently require a click on the tray icon before it can position the window. That's what the `on_tray_event(app, &event);` line mentioned in the linked issue is for.

UrsDeSwardt:
@FabianLars would it not make more sense to return an error instead of panicking?

FabianLars:
yes

PR #3420, opened 2026-05-14
fix: replace panics with error returns

UrsDeSwardt:
These panics are causing all sorts of havoc in our app. This PR just removes the panics and returns `Err` instead.

Related issue: https://github.com/tauri-apps/plugins-workspace/issues/1583#issuecomment-2681486868

### DIFF
--- a/.changes/positioner-panic.md
+++ b/.changes/positioner-panic.md
@@ -1,0 +1,6 @@
+---
+positioner: patch
+positioner-js: patch
+---
+
+Removed panics and replaced them with error handling.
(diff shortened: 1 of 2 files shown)


UrsDeSwardt:
@Legend-Master can I get another pass please?

Legend-Master (dismissed):
Thanks! Could you also add a change file?

UrsDeSwardt:
@Legend-Master Done! :)

Legend-Master (approved):
~~You'll also need to sign your commits for me to merge this~~ Or maybe not I guess

Thanks!

PR #3445, opened 2026-06-10
feat(log): rotate log file on each session

bajoca05:
In the Tauri Log plugin, file rotation occurs when the file has reached a max size. Otherwise, new lines are appended to the existing file from a previous session, if any. In some scenarios, it is very convenient to have one log file per session.

This PR proposes, on top of the rotation, a new strategy for opening the existing file, with two variants:
- Append: new log lines are appended to the existing file, if any. (current approach)
- Rotate: any existing file is rotated, and the rotator starts a new file.

### DIFF
--- a/.changes/log-file-open-strategy.md
+++ b/.changes/log-file-open-strategy.md
@@ -1,0 +1,6 @@
+---
+"log": minor:feat
+"log-js": minor
+---
+
+Added the `FileOpenStrategy` for log rotation. It defaults to append into existing file if any (previous behaviour), and brings a new feature to create a new file per session: `FileOpenStrategy::Rotate`. 
--- a/plugins/log/src/lib.rs
+++ b/plugins/log/src/lib.rs
@@ -472,6 +472,9 @@ impl Builder {
         self
     }
 
+    /// Sets the strategy to open the log file.
+    ///
+    /// The default is to append to the existing log file, if any.
     pub fn file_open_strategy(mut self, file_open_strategy: FileOpenStrategy) -> Self {
         self.file_open_strategy = file_open_strategy;
         self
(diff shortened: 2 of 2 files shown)


Legend-Master (dismissed):
Thanks! Could you also add a change file?

plugins/log/src/lib.rs
```diff
@@ -456,6 +472,11 @@ impl Builder {
         self
     }
 
+    pub fn file_open_strategy(mut self, file_open_strategy: FileOpenStrategy) -> Self {
```
Legend-Master, line 478: It would be nice to also have a bit of docs on the defaults here

bajoca05: Done!

plugins/log/src/lib.rs
```diff
@@ -456,6 +472,14 @@ impl Builder {
         self
     }
 
+    /// Sets the strategy to open the log file.
+    ///
+    /// The default is to append to the existing log file, if any.
```
Legend-Master:
```suggestion
    /// The default is [`FileOpenStrategy::Append`].
```

bajoca05:
> Thanks! Could you also add a change file?
> 
> > https://github.com/tauri-apps/plugins-workspace/blob/v2/.changes/readme.md

Added!

.changes/log-file-open-strategy.md
```diff
@@ -0,0 +1,6 @@
+---
+"log": minor:feat
+"log-js": minor
```
Legend-Master, line 3:
```suggestion
"log": minor:feat
"log-js": minor:feat
```

Legend-Master: Seem like it doesn't really matter in this repo

Legend-Master (approved):
Thanks!

PR #3449, opened 2026-06-11
fix(positioner): return error instead of panicking when window has no monitor

skkap:
`calculate_position` calls `window.current_monitor()?.unwrap()`. `current_monitor()` legitimately returns `Ok(None)` when the window is transiently not associated with any monitor — e.g. during display sleep or monitor reconfiguration — so the `unwrap` panics and takes the whole app down. We hit this in production: a tray click during a display transition crashed the app inside `move_window(Position::TrayCenter)`.

#3420 already converted the sibling `panic!("Tray position not set")` branches in this function to error returns; this PR applies the same treatment to the remaining `unwrap`, returning `tauri::Error::Io` in the same shape.

Includes a `.changes` patch entry for `positioner` and `positioner-js`, matching #3420.

### DIFF
--- a/.changes/positioner-current-monitor-panic.md
+++ b/.changes/positioner-current-monitor-panic.md
@@ -1,0 +1,6 @@
+---
+positioner: patch
+positioner-js: patch
+---
+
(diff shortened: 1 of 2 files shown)


Legend-Master (approved):
Thanks!

PR #3466, opened 2026-06-25
fix(single-instance): Yield unix listener in macos

bajoca05:
In the single-instance plugin, the macos implementation spawns a Tauri task that never yields. It uses a [standard UnixListener](https://doc.rust-lang.org/std/os/unix/net/struct.UnixListener.html) that blocks the thread while awaiting for incoming connections.

This PR proposes to use a non blocking `tokio::net::UnixListener` instead, so the task is yielded until the next incoming connection. This time, the thread will be available for other tasks in the runtime.

### DIFF
--- a/.changes/fix-yield-single-instance-macos.md
+++ b/.changes/fix-yield-single-instance-macos.md
@@ -1,0 +1,5 @@
+---
+"single-instance": minor-fix
+"single-instance": minor-fix
+---
+Fix blocked thread on the single-instance plugin for MacOS: replace standard `UnixListener` with `tokio::net::UnixListener`, so the task can yield.
--- a/.changes/fix-yield-single-instance-macos.md
+++ b/.changes/fix-yield-single-instance-macos.md
@@ -1,5 +1,4 @@
 ---
-"single-instance": minor-fix
-"single-instance": minor-fix
+"single-instance": patch
 ---
 Fix blocked thread on the single-instance plugin for MacOS: replace standard `UnixListener` with `tokio::net::UnixListener`, so the task can yield.
--- a/plugins/single-instance/Cargo.toml
+++ b/plugins/single-instance/Cargo.toml
@@ -24,7 +24,7 @@ tracing = { workspace = true }
 thiserror = { workspace = true }
 tauri-plugin-deep-link = { path = "../deep-link", version = "2.4.9", optional = true }
 semver = { version = "1", optional = true }
-tokio = "1"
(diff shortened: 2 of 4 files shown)


plugins/single-instance/src/platform_impl/macos.rs
```diff
@@ -97,13 +98,16 @@ fn listen_for_other_instances<A: Runtime>(
...
     match UnixListener::bind(socket) {
-        Ok(listener) => {
+        Ok(std_listener) => {
+            std_listener.set_nonblocking(true).ok();
             tauri::async_runtime::spawn(async move {
-                for stream in listener.incoming() {
-                    match stream {
-                        Ok(mut stream) => {
+                let listener = tokio::net::UnixListener::from_std(std_listener)
+                    .expect("failed to convert std UnixListener to tokio yielding UnixListener");
```
FabianLars: we should not expect here, tracing::error + return is safer - you can change the initial UnixListener::bind to tokio's UnixListener::bind to keep the code simple and reuse the existing error print

plugins/single-instance/Cargo.toml
```diff
@@ -24,6 +24,7 @@ tracing = { workspace = true }
 thiserror = { workspace = true }
 tauri-plugin-deep-link = { path = "../deep-link", version = "2.4.9", optional = true }
 semver = { version = "1", optional = true }
+tokio = "1"
```
FabianLars: we should probably explicitly enable the `net` feature here since tauri doesn't do that. Edit: this should also fix the CI error i didn't see

.changes/fix-yield-single-instance-macos.md
```diff
@@ -0,0 +1,5 @@
+---
+"single-instance": minor-fix
+"single-instance": minor-fix
```
FabianLars:
```suggestion
"single-instance": patch
```

FabianLars (changes requested):
thanks for the pr :) you're right, we should use std blocking in a tokio task

plugins/single-instance/Cargo.toml
bajoca05: Done

plugins/single-instance/src/platform_impl/macos.rs
bajoca05: Tokio UnixListener::bind requires to be done inside the runtime, so I moved it inside the task in 29ae77542073, which should be fine?

FabianLars (approved):

PR #3477, opened 2026-07-10
patch(log): use kv stable feature rather than unstable

bajoca05:
Log plugin is using the `kv_unstable` feature from the `log` crate. That feature was stabilized in [log v0.4.21](https://github.com/rust-lang/log/compare/0.4.20...0.4.21). And this workspace is using [v0.4.27](https://github.com/bajoca05/plugins-workspace/blob/v2/Cargo.lock#L3463), no need to keep using it.

### DIFF
--- a/.changes/log-kv-stable-feature.md
+++ b/.changes/log-kv-stable-feature.md
@@ -3,4 +3,4 @@
 "log-js": patch
 ---
 
-Log plugin is using the `kv_unstable` feature from the `log` crate. That feature was stabilized in [log v0.4.21](https://github.com/rust-lang/log/compare/0.4.20...0.4.21). And this workspace is using [v0.4.27](https://github.com/bajoca05/plugins-workspace/blob/v2/Cargo.lock#L3463), no need to keep using it.
+Migrated from the `log` crate's `kv_unstable` feature flag to the `kv` feature flag stabilized in `0.4.21`.
--- a/.changes/log-kv-stable-feature.md
+++ b/.changes/log-kv-stable-feature.md
@@ -1,0 +1,6 @@
+---
+"log": patch
+"log-js": patch
+---
+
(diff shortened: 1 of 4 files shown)


Legend-Master (dismissed):
Thanks!

Let's change the workspace log version to the minimum supported version for the `kv` feature so cargo doesn't select a lower version.

Also could you add a change file so we can release this?

Legend-Master (dismissed):
Thanks!

Cargo.toml
```diff
@@ -11,7 +11,7 @@ resolver = "2"
 [workspace.dependencies]
 serde = { version = "1", features = ["derive"] }
 tracing = "0.1"
-log = "0.4"
+log = "0.4.27"
```
FabianLars:
```suggestion
log = "0.4.21"
```

shouldn't we keep 0.4.21 ? The lockfile is only used in this repo and ignored on the user side so in theory they could use 0.4.23 or whatever.

I highly doubt anyone needs >0.4.21, <0.4.27 but technically we can support it.

.changes/log-kv-stable-feature.md
```diff
@@ -0,0 +1,6 @@
+---
+"log": patch
+"log-js": patch
+---
+
+Log plugin is using the `kv_unstable` feature from the `log` crate. That feature was stabilized in [log v0.4.21](https://github.com/rust-lang/log/compare/0.4.20...0.4.21). And this workspace is using [v0.4.27](https://github.com/bajoca05/plugins-workspace/blob/v2/Cargo.lock#L3463), no need to keep using it.
```
FabianLars:
```suggestion
Migrated from the `log` crate's `kv_unstable` feature flag to the `kv` feature flag stabilized in `0.4.21`.
```
no need to mention the lockfile since that's not user facing

Cargo.toml
Legend-Master: Sounds reasonable, let's change it to 0.4.21. Honestly I don't mind it either ways 😂

FabianLars: i don't care either, i just felt like nitpicking today

FabianLars (approved):

PR #3491, opened 2026-07-16
enhance(dialog): use `MaterialAlertDialogBuilder`

Legend-Master:
Use `com.google.android.material.dialog.MaterialAlertDialogBuilder` instead of `AlertDialog` so the dialog follows the app's theme

Also

- Updated the example app's theme
- Re-generated some example app's kotlin files
- Fixed example app's insets on Android so it looks edge to edge

---

Before:

image

After:

image


FabianLars (commented):
untested

this bothered me for so long but not enough to look into it 😂

FabianLars (approved):

PR #3495, opened 2026-07-17
fix(single-instance): close race between mutex creation and event window on Windows

SSS135:
On Windows, `setup` creates the named mutex first and the event target window later. A second launch that lands in between sees `ERROR_ALREADY_EXISTS`, but `FindWindowW` comes back null, so the check falls through to `Ok(())`. The second process doesn't forward anything or exit; it just keeps running as a normal, unguarded instance. It also never creates a window of its own, and its leaked handle keeps the mutex name alive, so a third launch can slip through the same way even after the first instance is gone.

The gap between the two calls is tiny on an idle machine, but it stretches badly under load. We hit this in a real app: two instances started a few hundred milliseconds apart on a busy machine and both passed the check.

The fix keeps the second instance in the check instead of giving up. When the mutex exists but the window isn't there yet, it polls for the window while waiting on the mutex in 50 ms slices, up to 5 seconds total. If the window shows up (the primary finished its setup), it forwards the args and exits, same as before. If `WaitForSingleObject` returns `WAIT_OBJECT_0` or `WAIT_ABANDONED`, the holder released the mutex or died with it, and this process takes over as the primary; that also helps updaters that respawn the app while the old process is still exiting, which currently only works because the old process usually dies before the new one reaches this check. On timeout or `WAIT_FAILED` it fails open like the current code does, so a wedged process that never made a window can't block the app from starting forever.

The primary path (managing the mutex handle and creating the event window) is now unconditional. A fail-open instance therefore still creates its window, so later launches can forward to it, and the mutex handle is managed instead of leaking silently.

Normal startups are unaffected: a fresh start creates the mutex and proceeds immediately, and a second launch against a healthy primary finds the window on the first probe, before any wait. No public API changes, nothing touched on Linux or macOS.

Tested on Windows with `cargo check -p tauri-plugin-single-instance` and clippy, and the same wait logic exercised in a real app: bursts of simultaneous launches leave exactly one process, a launch during an artificially widened mutex-to-window gap forwards and exits instead of duplicating, and killing the primary while a second instance waits hands the primary role over through `WAIT_ABANDONED`.

### DIFF
--- a/plugins/single-instance/src/platform_impl/windows.rs
+++ b/plugins/single-instance/src/platform_impl/windows.rs
@@ -12,11 +12,14 @@ use tauri::{
     AppHandle, Manager, RunEvent, Runtime,
 };
 use windows_sys::Win32::{
-    Foundation::{CloseHandle, GetLastError, ERROR_ALREADY_EXISTS, HWND, LPARAM, LRESULT, WPARAM},
+    Foundation::{
+        CloseHandle, GetLastError, ERROR_ALREADY_EXISTS, HWND, LPARAM, LRESULT, WAIT_ABANDONED,
+        WAIT_OBJECT_0, WAIT_TIMEOUT, WPARAM,
+    },
     System::{
         DataExchange::COPYDATASTRUCT,
         LibraryLoader::GetModuleHandleW,
-        Threading::{CreateMutexW, ReleaseMutex},
+        Threading::{CreateMutexW, ReleaseMutex, WaitForSingleObject},
     },
     UI::WindowsAndMessaging::{
         self as w32wm, CreateWindowExW, DefWindowProcW, DestroyWindow, FindWindowW,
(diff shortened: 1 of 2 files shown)


plugins/single-instance/src/platform_impl/windows.rs
```diff
@@ -70,42 +73,76 @@ pub fn init<R: Runtime>(callback: Box<SingleInstanceCallback<R>>) -> TauriPlugin
...
-                        SendMessageW(hwnd, WM_COPYDATA, 0, &cds as *const _ as _);
+                    if std::time::Instant::now() >= deadline {
+                        // No primary window appeared and nobody released the mutex
+                        // (e.g. the holder is wedged before creating its window).
+                        // Fail open rather than block startup forever.
+                        break;
+                    }
 
-                        app.cleanup_before_exit();
-                        std::process::exit(0);
```
Legend-Master, line 94: I would honestly just keep it simple by always exiting no matter if we found the window or not as long as the mutex exists.

SSS135: That would be simpler, and it does close both holes (the unguarded duplicate and the leaked handle letting a third launch slip through). But it costs three things the wait loop buys:

1. Forwarding: if the second launch lands before the primary has created its window, exiting immediately drops the args/cwd. The callback never fires in the primary, so with the deep-link feature a link clicked while the app is still starting up is silently lost. The loop just parks the second process until the window shows up, then forwards as before.
2. Overlapping relaunches: whenever a new process's CreateMutexW lands while an exiting holder still owns the mutex, "mutex exists, exit" leaves the app not running: the new process exits, then the old one dies. That overlap happens without any bug in user code: the user closes the app and immediately reopens it, a main-thread AppHandle::restart() (which skips RunEvent::Exit delivery, so the mutex is only released by process death), a crash followed by a watchdog respawn, or a custom updater that spawns the new exe and then exits. Waiting on the mutex turns all of these into a clean takeover via WAIT_OBJECT_0/WAIT_ABANDONED. (The stock request_restart()/relaunch() path is unaffected either way, since it releases the mutex before spawning.)
3. A holder that wedged after CreateMutexW but before creating its window would block every future launch until it's killed; today's code fails open there, and the PR keeps that.

The part I'm happy to change is the timeout path. It currently fails open to match the existing code, but if you'd rather have strict single-instance semantics I can make the timeout exit instead. That gives you "always exit while the mutex is held" without losing forwarding or dropping quick relaunches. Can also shorten the 5 s if that feels long.

PR #3504, opened 2026-07-18
fix(docs): deep link platform specific and fs scope

vasfvitor:
closes #3503

PR #3507, opened 2026-07-20
fix(fs): default permissions

Legend-Master:
Reference: #3504


Legend-Master (commented):
@FabianLars Those changes are very much breaking, I'm not sure if we should do it now since they were wrong from the very beginning that we introduced them...

FabianLars:
i'd do it 🤷

Legend-Master:
Sounds fine, the scope was none before so this is more permissive than before.

Legend-Master:
Actually, if we really do this at the deny in command level, doesn't it mean you can't allow access to those directories as long as you have `fs:default`?

FabianLars:
i wondered the same. fine for ebwebview but the linux one is heavy. in v3 we should probably create a similar dir for webkitgtk.

for now though how about we deny the individual directories? This should hopefully cover most of it?
- CacheStorage/
- WebKitCache/
- cookies (file)
- hsts-storage.sqlite (file)
- localstorage/
- mediakeys/
- storage/

Legend-Master:
I would honestly just keep the current behavior, don't allow any directory nor deny any. No security risks nor breaking changes 😂 even for v3.

FabianLars:
let's ping @tweidinger for this. iirc this was one of the sets we wrote in a 1:1 call back then.

FabianLars:
i actually missed the "don't allow any directory" part lol - i can live with that then for now. Since then nothing changes we can still fix it later when he responded.

tweidinger:
Not fully sure about all the implications but we have been tracking the issue itself in this repo's security advisories as a draft 😅 @Legend-Master you may want to check it for the other issues as well. My take is that this was a typo and was due to the assumption that it will be picked up by the global scope the way we wrote it, which it wasn't.

Even if it is potentially breaking, the docs and our expectations were always that default permissions deny WebView folder access and whoever messes with that is able to copy default without the deny block.

We should go to expected behavior and give a heads up on the potential breaking change if you messed with the WebView folder on default permissions.

Legend-Master:
I would like to keep the current behavior (`fs:default` doesn't allow/deny any directories) as we had it for so long. Our commands will not allow any access if there's no allowed scopes so I really don't think this is not secure by any means.

The entire fs permission system is already confusing enough that If we go about this, it becomes even more confusing that all access to `$APPLOCALDATA` on Linux are denied.

FabianLars:
> it becomes even more confusing that all access to $APPLOCALDATA on Linux are denied.

Well, that's a nogo anyway. Like i said we should deny the individual dirs instead.

I do see real value in denying the webview data though. If users allow applocaldata without the webview deny, js could somewhat easily circumvent browser's cookie website isolation by just reading the files directly.

tweidinger:
> I would like to keep the current behavior (fs:default doesn't allow/deny any directories) as we had it for so long.

That was not intended and shouldn't justify us not fixing it.

> Our commands will not allow any access if there's no allowed scopes so I really don't think this is not secure by any means.

Agreed on this one, with the exception that we communicated the previous point in our docs and threat model and default plugin permission docs. 
It's a defense in depth measure to make sure there are less footguns. (Oh the irony just hit me 😂 )

I'm not going to die on this hill but am wondering where you have encountered any usage of the WebView folder in any typical Tauri app? I think that it is a small group that will be affected by us aligning code with docs.

> Like i said we should deny the individual dirs instead.

Full Agreement, if you have context for me why this is a No-Go because we came up with this a while ago 😁 
I remember Linux not having aa clear WebView subfolder but don't remember where applocaldata is used normally.

FabianLars:
> I'm not going to die on this hill but am wondering where you have encountered any usage of the WebView folder in any typical Tauri app? I think that it is a small group that will be affected by us aligning code with docs.

I only ever saw this when devs wanted to change their bundle id and migrate the webview data - but that usually has to use native rust stuff because once tauri runs (which makes tauri's fs apis available) many webview files are locked/inuse

> Full Agreement, if you have context for me why this is a No-Go because we came up with this a while ago 😁
I remember Linux not having aa clear WebView subfolder but don't remember where applocaldata is used normally.

iirc some of our plugins use it? doesn't matter since all appX dirs are used by devs pretty often and we'd deny them all the use of it here since as Tony said, a denied folder can't be allowed again.
idk why i didn't think about this back then but i don't think much in general so maybe that's why 🙃

Legend-Master:
> That was not intended and shouldn't justify us not fixing it.

Well... It really breaks. If we really want to bring it back, I would say let's wait for v3. (and change the docs to reflect on what actually happened)

Legend-Master:
@FabianLars @tweidinger I am going to change the description of `fs:default` and remove `"deny-default"` from the default permissions. We can discuss how to improve this in v3, but I really don't want a breaking change like this in v2 anymore. Let me know your thoughts.

FabianLars:
i told you my thoughts already, i don't see denying access to the webview folders (that are mostly locked/inuse by the webview anyway) like we always intended as breaking. Even if we haven't semi-documented it, i'd consider this an important security fix because again, this can be easily abused to get around a browser's cross site isolation.

I'd much rather make 2 developers that need access to the webview data change their permissions than leave this hole open. Edit: I'd also want us to publish the draft advisory for this then.

Again, this includes fixing the Linux variant to not deny the whole applocaldata folder anymore, cause i agree that this would be too breaking.

Legend-Master:
Do we have an exhaustive list of all the Linux webview files? I could live with just breaking/not allowing those files. (although I find it probably still very confusing that some files in the local app data dir can't be accessed, not matter how you configure it with the `fs:default` permission)

tweidinger:
Full agreement on Fabian's perspective.

We can also offer a permission set (fs:default-legacy) that preserves behavior as a named set to keep the technically breaking migration easy for anyone absolutely relying on this unexpected behavior.

Legend-Master:
Please no `fs:default-legacy`, it's already really confusing 😂 Anyways, ok, we change the Linux deny paths and fix the implementation + release a security advisory, also including the docs for how to access those paths in the description.

FabianLars:
> Do we have an exhaustive list of all the Linux webview files?

In an app i had installed for a while and browsed various websites i have these folders
CacheStorage
cookies 
hsts-storage.sqlite
serviceworkers
WebKitCache
databases
localstorage
mediakeys
storage

Legend-Master:
Updated. Please check and just edit the path that I'm not sure which one is a folder.

FabianLars:
that looks correct, only cookies and hsts-storage.sqlite are files on my end.

according to https://github.com/WebKit/WebKit/blob/df40d92d0fcd348b3fb39bb138e6c5565f6c3d3e/Source/WebKit/UIProcess/WebsiteData/WebsiteDataStore.cpp#L2366 there may be even more:
deviceidhashsalts/
mediakeyshashsalts/
itp/
resourcemonitorthrottler/ (i don't think this exists for us yet)

but i've yet to see anyone of these in the wild

Legend-Master:
We should get this one on to the priority that it is blocking all the other plugin releases

plugins/fs/permissions/autogenerated/reference.md
```diff
@@ -22,7 +22,9 @@ the `mkdir` command.
 
 This default permission set prevents access to critical components
 of the Tauri application by default.
-On Windows the webview data folder access is denied.
+
+- On Windows the webview data folder `$APPLOCALDATA/EBWebView` access is denied.
+- On Linux the webview data folder `$APPLOCALDATA` access is denied.
```
FabianLars: I think you need to regen the autogend files

FabianLars (approved):

PR #3508, opened 2026-07-20
fix(dialog): use parsed MIME type for Android save dialog(#issue 3498)

### LINKED ISSUE #3498: [bug] Android save dialog renames duplicate files as "file.pdf(1)" instead of "file(1).pdf"
### Describe the bug

On Android, when saving a file via `@tauri-apps/plugin-dialog`'s `save()` 
to a location where a file with the same name already exists, the SAF 
(Storage Access Framework) document provider appends the disambiguation 
suffix AFTER the full filename instead of before the extension.

Example:
- Expected: `10mb.pdf` → `10mb(1).pdf`
- Actual:   `10mb.pdf` → `10mb.pdf(1)`

### Root cause

**File:** `plugins/dialog/android/src/main/java/DialogPlugin.kt`
**Function:** `saveFileDialog()`
**Line:** 207

```kotlin
val intent = Intent(Intent.ACTION_CREATE_DOCUMENT)
intent.addCategory(Intent.CATEGORY_OPENABLE)
intent.putExtra(Intent.EXTRA_TITLE, args.fileName ?: "")
intent.type = "*/*"   // ← hardcoded regardless of filters

if (parsedTypes.isNotEmpty()) {
  intent.putExtra(Intent.EXTRA_MIME_TYPES, parsedTypes)
}
```

### Reproduction

Call save({ defaultPath: "10mb.pdf", filters: [{ name: "PDF", extensions: ["pdf"] }] }).
Save to a folder that already contains 10mb.pdf.
Observe the resulting file is named 10mb.pdf(1) instead of 10mb(1).pdf.

### Expected behavior

_No response_

### Full `tauri info` output

```text
Use the first parsed filter's MIME type when available, falling back to
the wildcard only when no filters are given:


intent.type = parsedTypes.firstOrNull() ?: "*/*"
I've tested this fix locally (forked build) and confirmed it produces the
correct name(1).ext format on a real device (Pixel 7, Android 14).
```

### Stack trace

```text

```

### Additional context

Use the first parsed filter's MIME type when available, falling back to
the wildcard only when no filters are given:

intent.type = parsedTypes.firstOrNull() ?: "*/*"
I've tested this fix locally (forked build) and confirmed it produces the
correct name(1).ext format on a real device (Pixel 7, Android 14).

purvsinojiya-inventyv:
## Summary

This PR updates the Android `saveFileDialog` implementation to use the first parsed MIME type as the `Intent` type instead of always using `*/*`.

```kotlin
intent.type = parsedTypes.firstOrNull() ?: "*/*"
```

If no MIME type can be determined, the existing `*/*` fallback is preserved.

## Motivation

The dialog plugin already parses the selected file filters into MIME types using `parseFiltersOption(args.filters)`, but the parsed MIME types are only passed through `Intent.EXTRA_MIME_TYPES` and are not used as the primary `Intent` type.

Using the first parsed MIME type provides Android with a more specific document type while preserving the current behavior for unknown file types.

## Changes

- Use the first parsed MIME type for `intent.type` when available.
- Preserve the existing `*/*` fallback if no MIME type is available.
- No behavior changes for callers that do not provide file filters.

Closes #3498

### DIFF
--- a/plugins/dialog/android/src/main/java/DialogPlugin.kt
+++ b/plugins/dialog/android/src/main/java/DialogPlugin.kt
@@ -204,9 +204,10 @@ class DialogPlugin(private val activity: Activity): Plugin(activity) {
       val intent = Intent(Intent.ACTION_CREATE_DOCUMENT)
       intent.addCategory(Intent.CATEGORY_OPENABLE)
       intent.putExtra(Intent.EXTRA_TITLE, args.fileName ?: "")
-      intent.type = parsedTypes.firstOrNull() ?: "*/*"
-
-      if (parsedTypes.isNotEmpty()) {
+      if (parsedTypes.size == 1) {
+        intent.type = parsedTypes.first()
+      } else {
+        intent.type = "*/*"
         intent.putExtra(Intent.EXTRA_MIME_TYPES, parsedTypes)
       }
 
--- a/plugins/dialog/android/src/main/java/DialogPlugin.kt
+++ b/plugins/dialog/android/src/main/java/DialogPlugin.kt
@@ -204,7 +204,7 @@ class DialogPlugin(private val activity: Activity): Plugin(activity) {
       val intent = Intent(Intent.ACTION_CREATE_DOCUMENT)
       intent.addCategory(Intent.CATEGORY_OPENABLE)
       intent.putExtra(Intent.EXTRA_TITLE, args.fileName ?: "")
-      intent.type = "*/*"
+      intent.type = parsedTypes.firstOrNull() ?: "*/*"
 
       if (parsedTypes.isNotEmpty()) {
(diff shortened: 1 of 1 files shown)


FabianLars:
I think we should only set the mime type if there's only a single one set in the filters. Otherwise you may end up with a mismatch easily.

Also, we should probablyyy look into deriving the mime type from the file extension as well (just for the save dialog) 🤔 but that can be done another time, may even be considered a breaking change.

purvsinojiya-inventyv:
Done. The MIME type is now only set when there's a single MIME type in the filters to avoid mismatches.

purvsinojiya-inventyv:
I've made the requested changes. You can merge it. Thanks!

PR #3510, opened 2026-07-21
refactor(fs): cfg gate some ios code

Legend-Master:
I feel like the entire fs commands are a mess now :(


FabianLars:
> I feel like the entire fs commands are a mess now :(

always have been

FabianLars (approved):

PR #3513, opened 2026-07-24
feat(window-state): add custom state directory support

Lintern:
Add `with_state_dir` builder method to allow users to specify a custom
directory for saving and restoring window state, instead of using the
default app config directory.

This is useful for:
- Storing window state in a custom location
- Sharing window state across multiple apps
- Testing scenarios where isolated state directories are needed


Legend-Master:
Thanks for contributing. There's already a PR up for this, so I'm closing this for now, see https://github.com/tauri-apps/plugins-workspace/pull/3022#issuecomment-3570743741 Thanks anyways.

PR #3526, opened 2026-08-04
feat(updater): add `system-proxy` feature

### LINKED ISSUE #3522: [updater] Not going thouth proxy
With tauri v2 by default the requests are using the system / browser proxy for http requests. 

The requests going out from the updater plugin are not since they came directly from the rust tauri backend. The expected behaviour would be to have go to the system proxy like other request by default, which is not the case.

Is there a way to force the requests of the updater plugin to go through the proxy ?

Legend-Master:
Fix #3522

We can also add this to the default feature


FabianLars:
for the updater plugin i'd always enable it and for the http plugin i'd expose it as a feature flag that's enabled by default. For the http plugin we can also remove macos-system-configuration from the defaults (since system-proxy is its replacement)

Legend-Master:
I would probably add it to the default feature since anything we add here that we don't always need, can't be disabled by the user

FabianLars:
don't think that matters for this feature (maybe depends on whether we want to expose the no_proxy config) but do whatever you prefer.

FabianLars (approved):

Issue #2225, opened 2024-12-19
[feat][clipboard-manager] Add "readImagePNG()" method


WolfBearGames:
I am using the clipboard-manager plugin and readImage() in JS to get an image from the clipboard. The method returns the data as an uncompressed rgba bytes array which results in over 100MB for a screenshot of a UHD Desktop. This results in multiple seconds of wait time and is very inefficient.

A new `readImagePNG()` method could transfer the data as a PNG resulting a less data that need to be transferred. 

The code for this already exists, before it was changed:
https://github.com/tauri-apps/plugins-workspace/commit/d57df4debe7c75cfbd6d6558fff1beb07dbee54c

It should be relatively easy to add a `readImagePNG()` as an additional option in the clipboard-manager plugin.
(Alternatively, the `readImage()` method could accept a paramater which format to return)

PR #3532, opened 2026-08-08
feat(clipboard-manager): add readImagePNG to read clipboard image as PNG bytes

devwez:
## Problem

`readImage()` returns the full uncompressed RGBA buffer — over 100MB for a UHD screenshot — which is slow to transfer over IPC (#2225).

## Fix

Added `readImagePNG()`, which PNG-encodes the same clipboard image (via the `image` crate, `default-features = false, features = ["png"]` to keep the dependency light) before returning it, so large screenshots transfer as a fraction of the size.

- `desktop.rs`: `Clipboard::read_image_png()` — same arboard `get_image()` call as `read_image()`, PNG-encodes the result.
- `mobile.rs`: matching `"Unsupported on this platform"` stub, same as every other image method on mobile.
- `commands.rs` / `lib.rs` / `build.rs`: new `read_image_png` command, registered and wired through `tauri::ipc::Response` (matches the `fs` plugin's `readFile` pattern for binary transfer, rather than JSON-serializing a `Vec<u8>`).
- `guest-js/index.ts`: `readImagePNG(): Promise<Uint8Array>`, handling both the `ArrayBuffer` and `number[]` IPC shapes like `readFile` does.
- Added a changeset (`minor:feat` for both `clipboard-manager` and `clipboard-manager-js`).

I did look at the PNG-handling code linked in the issue (from before #1986 removed it) — that version was actually buggy: it made `read_image` *always* return PNG bytes mislabeled as raw RGBA (see #1985/#1986), which is why it got reverted rather than kept. This PR adds PNG support back as a genuinely separate method instead of reintroducing that bug.

## Testing

- `cargo check` / `cargo clippy -p tauri-plugin-clipboard-manager --all-targets`: clean, no warnings.
- `tsc --noEmit` on the plugin package and `eslint` on the changed file: clean.
- Verified the actual encode path end-to-end against a real OS clipboard (not just compiled): wrote a synthetic RGBA image via `arboard`, ran it through the same `PngEncoder` call `read_image_png` uses, decoded the resulting PNG back with `image::load_from_memory`, and confirmed the pixels are byte-identical to the original buffer.

Per the AI tool policy in CONTRIBUTING.md: I used AI assistance to write this, but reviewed and tested everything above myself before opening the PR — happy to answer questions or make changes in review.

### DIFF
--- a/.changes/clipboard-read-image-png.md
+++ b/.changes/clipboard-read-image-png.md
@@ -1,0 +1,6 @@
+---
+"clipboard-manager": minor:feat
+"clipboard-manager-js": minor:feat
+---
+
+Added `readImagePNG` to read the clipboard image as PNG-encoded bytes instead of a raw decoded RGBA buffer, which is significantly smaller to transfer for large images (e.g. UHD screenshots).
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6496,6 +6496,7 @@ name = "tauri-plugin-clipboard-manager"
 version = "2.3.2"
 dependencies = [
  "arboard",
+ "image",
  "log",
  "serde",
  "serde_json",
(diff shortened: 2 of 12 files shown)


devwez:
Friendly bump — ready for review. Let me know if anything needs adjusting!

PR #3540, opened 2026-08-17
fix(notification): don't force window.Notification.permission to "denied" on Windows (fix #3512)

### LINKED ISSUE #3512: [bug][notification] Windows: init script always sets `window.Notification.permission` to `"denied"` at startup
## Describe the bug

On Windows, the notification plugin's injected init script stamps
`window.Notification.permission = "denied"` on every page load, even though the
desktop backend unconditionally reports `PermissionState::Granted` on Windows.
Any app that reads `window.Notification.permission` (or gates sending on it)
concludes notifications are blocked.

In [`guest-js/init.ts`](https://github.com/tauri-apps/plugins-workspace/blob/v2/plugins/notification/guest-js/init.ts#L12-L18):

```ts
async function isPermissionGranted(): Promise<boolean> {
  if (window.Notification.permission !== 'default' || __TEMPLATE_windows__) {
    return await Promise.resolve(window.Notification.permission === 'granted')
  }
  return await invoke('plugin:notification|is_permission_granted')
}
```

By the time this runs, `window.Notification` has already been replaced by the shim
defined earlier in the same IIFE, whose `permission` is initialized to `'default'`.
With `__TEMPLATE_windows__ === true` the function therefore always returns
`'default' === 'granted'` → `false`, and the bootstrap at the bottom of the file maps
`false` → `'denied'`:

```ts
void isPermissionGranted().then(function (response) {
  if (response === null) {
    setNotificationPermission('default')
  } else {
    setNotificationPermission(response ? 'granted' : 'denied')
  }
})
```

The Windows short-circuit presumably predates the shim replacement (it would make
sense reading the *native* WebView2 permission), but as written it can only ever
observe the shim's own initial value.

## Reproduction

Any Tauri v2 app on Windows with the notification plugin registered:

```js
console.log(window.Notification.permission)          // "denied" — always, fresh profile
await window.Notification.requestPermission()        // "granted"
console.log(window.Notification.permission)          // "granted" — until next launch
```

Observed in the wild in block/buzz (block/buzz#2445), where the settings UI
reads `window.Notification.permission`, sees `"denied"`, and tells users to enable
notifications in system settings (where nothing is disabled).

## Expected behavior

On Windows the bootstrap should reflect the backend (`permission_state()` returns
`Granted`), i.e. initialize to `"granted"` — or fall through to
`invoke('plugin:notification|is_permission_granted')` like other platforms:

```ts
async function isPermissionGranted(): Promise<boolean> {
  if (window.Notification.permission !== 'default') {
    return window.Notification.permission === 'granted'
  }
  return await invoke('plugin:notification|is_permission_granted')
}
```

## Platform and versions

- OS: Windows 11 (10.0.26200), WebView2 Evergreen
- tauri-plugin-notification: 2.3.3 (bug still present on `v2` branch tip)
- tauri: 2.x

AJ-Base44:
## Summary

On Windows, the notification bootstrap always changed `window.Notification.permission` from its initial `default` value to `denied`, even though the desktop backend reports permission as granted. This removes the Windows-only short-circuit so startup asks the backend, matching the existing macOS and Linux path.

## Root cause

The init script replaces `window.Notification` with a shim whose closure-backed `permission` value starts as `default`. The Windows template made `isPermissionGranted()` short-circuit before its backend invoke, so it compared that initial value with `granted`, always resolved `false`, and the bootstrap wrote `denied`.

Repository history shows that the short-circuit was introduced in `4c120d4176d9a59f22bd215a7cec09d92ceab6da`, immediately reverted in `93bdb6d133b89d990ddd5470536edfe4cb617e89`, and restored in `725ff4295e56df9c30c099813bd64b96fe61b945`. Its recorded purpose was to avoid a Windows WebView `STATUS_ACCESS_VIOLATION` when loading remote websites. The shim itself predates that change, having arrived in `e9bbe941819d1dfa8259b758dba7b97cec8cd92f`; history does not explain the deeper reason the backend invoke was associated with the crash.

## Fix

The Windows short-circuit is removed, so Windows now falls through to `plugin:notification|is_permission_granted` like macOS and Linux. The desktop backend already reports `Granted`. The dead `__TEMPLATE_windows__` replacement is removed from `lib.rs`, and `src/init-iife.js` is regenerated.

If maintainers prefer to retain the no-IPC Windows path, I can switch to a hardcoded resolved `granted` result and restore the Rust template hook.

## Blast radius

- Windows now makes one permission IPC call during page initialization and updates the shim from `default` to the desktop backend result.
- macOS and Linux are unchanged because they already used this backend path.
- iOS and Android are unchanged because the template was already false there and `mobile.rs` handles permission through the mobile implementation.
- The `windows7-compat` feature is unaffected because it changes notification display, not permission reporting.

## Alternative considered

The alternative is to return a hardcoded resolved `granted` result on Windows. That avoids startup IPC and preserves the old backend bypass, but it duplicates desktop policy in the frontend and could become stale if the backend later gains real desktop permission handling. The backend path was chosen because it keeps the backend as the source of truth and matches macOS and Linux.

## Behaviour notes

There is one extra IPC call on Windows page load. The permission can briefly remain `default` before becoming `granted`, matching macOS and Linux today. By code inspection, an app that omits `notification:default` and does not separately allow `is-permission-granted` can now have a rejected bootstrap invoke and remain at `default`; this path was not exercised in the runtime reproduction.

The runtime reproduction uses a localhost WebView and does not exercise the remote-site `STATUS_ACCESS_VIOLATION` path recorded in the 2024 history.

## Review notes

The correctness review confirmed the root cause, TypeScript/Rust semantics, generated bundle, and non-Windows blast-radius analysis. It flagged the older remote-site crash path and the missing-capability rejected Promise as unresolved Windows risks; both are disclosed here because the requested minimal design is the backend fall-through without a new catch or fallback.

## Out of scope, deliberately not done

- A hardcoded Windows `granted` result was not substituted because the selected design is to keep desktop permission policy in the backend; it remains the alternative above.
- A bootstrap catch/fallback was not added because that changes existing error handling beyond the four-line minimal fix. The missing-capability path remains documented above.

## Testing

- Before the fix on Windows: `boot: denied`, `1s: denied`, `click: denied`.
- After the fix, full restart 1: `boot: granted`, `1s: granted`, `click: granted`.
- After the fix, full restart 2: `boot: granted`, `1s: granted`, `click: granted`.
- Machine: Windows build 26200.9168 (25H2), WebView2 151.0.4129.86, Tauri 2.11.5, and `tauri-plugin-notification` 2.3.3.
- Passed `cargo fmt --all -- --check`.
- Passed `pnpm lint`.
- Passed `cargo clippy --package tauri-plugin-notification --all-targets -- -D warnings` and the same command with `--all-features`.
- Passed `cargo test --package tauri-plugin-notification --target x86_64-pc-windows-msvc --all-targets --all-features` with the installed stable Rust 1.96.0 MSVC toolchain. CI pins Rust 1.77.2, which is not installed locally.
- Passed the `.changes` extension guard, `node ./.scripts/ci/check-change-files.js`, the license-header invocation, and the notification `pnpm install && pnpm build` determinism check.
- Pinned Prettier passes the touched TypeScript file. The exact repo-wide command did not pass in this Windows checkout, so no repository-wide formatting pass is claimed and no out-of-scope files were rewritten.
- The exact root `pnpm build` did not return exit 0. The isolated notification package build and regeneration check passed repeatedly.
- Taplo is not installed locally. GitHub-action-only jobs and the Linux, macOS, iOS, and Android matrix jobs require their CI event context, operating systems, or SDKs and were not run here.
- The official `examples/api` binary compiled and launched. In its Notifications view, **Send test notification** was clicked; after the click, `Notification.permission` was `granted` immediately before and after screenshot capture, and the UI contained zero exact `Permission is denied` messages.

## Related

block/buzz#2445


plugins/notification/guest-js/init.ts
```diff
@@ -10,8 +10,7 @@ import type { Options } from './index'
   let permissionValue = 'default'
 
   async function isPermissionGranted(): Promise<boolean> {
-    // @ts-expect-error __TEMPLATE_windows__ will be replaced in rust before it's injected.
-    if (window.Notification.permission !== 'default' || __TEMPLATE_windows__) {
+    if (window.Notification.permission !== 'default') {
```
Legend-Master, line 13: @FabianLars I saw the change in #1556 so I would assume this is a no go right? I wonder if we should just return `true` here on all desktop platforms directly?

Edit: opened #3541 see if that works for you

AJ-Base44: Good catch, and you're right to flag it.

I hadn't seen #1556 when I opened this, only found the __TEMPLATE_windows__ short-circuit and traced it to a comparison bug against the shim's own default value. Didn't realize the intent was specifically to avoid the backend invoke on Windows. That's actually the exact risk I flagged as untested in the PR body, I only reproduced against a local dev server, not a remote page or anything Cloudflare-protected.

Rather than falling through to invoke like macOS and Linux, I'll push an update that keeps Windows resolving synchronously without calling the backend, same as #1556 intended, just fixing the actual bug in how that comparison was written. Should have it up shortly.

PR #3541, opened 2026-08-18
fix(notification): always set permission granted on desktop

### LINKED ISSUE #3512: [bug][notification] Windows: init script always sets `window.Notification.permission` to `"denied"` at startup

Legend-Master:
Closes #3512
Closes #3540

I honestly think we should stop injecting to browser APIs at least not as defaults...


plugins/notification/guest-js/init.ts
```diff
@@ -9,10 +9,9 @@ import type { Options } from './index'
   let permissionSettable = false
   let permissionValue = 'default'
 
-  async function isPermissionGranted(): Promise<boolean> {
-    // @ts-expect-error __TEMPLATE_windows__ will be replaced in rust before it's injected.
-    if (window.Notification.permission !== 'default' || __TEMPLATE_windows__) {
-      return await Promise.resolve(window.Notification.permission === 'granted')
+  async function isPermissionGranted(): Promise<boolean | null> {
+    if (window.Notification.permission !== 'default') {
+      return window.Notification.permission === 'granted'
```
FabianLars: isn't that actually incorrect (the prior version as well) for mobile? !default could also be denied, no?

FabianLars (commented):
> I honestly think we should stop injecting to browser APIs at least not as defaults...

then we'd lose desktop notifications on some platforms, maybe just macOS nowadays but i think it was more back when this was added. There were also some webview2 apis missing in 109 for windows 7 but i guess that stops being relevant in 2.12

Legend-Master:
> then we'd lose desktop notifications on some platforms, maybe just macOS nowadays but i think it was more back when this was added.

I meant we shouldn't inject/override the browser APIs by default since they're not exactly 1 to 1 replacements and if the app wants to use the plugin APIs, they can use them directly?

plugins/notification/guest-js/init.ts
Legend-Master: I honestly didn't really get the reason for the check, I just refactored the desktop checks away

Do you know the history on why this was checked at the first place? I can only see the plugin got introduced in #326

FabianLars:
i understood what you mean but some browser apis don't work / don't exist on some platforms. 
>  and if the app wants to use the plugin APIs, they can use them directly?

well, they want to use the browser apis

this would make more sense without the acl tbh cause with it it's not as seamless

Legend-Master:
I get the idea behind them, but if they aren't exact replacements, it could be a big catch/foot gun. That being said, if they're documented well enough, I could live with them.

plugins/notification/guest-js/init.ts
FabianLars: this was written way back in https://github.com/tauri-apps/tauri/commit/2681ad361b4295756bef560b5d74ece2a757749f#diff-cb470e8a8772900c19accd4c5ee3dc6f7f22017fe9fa2935759064b7192c7e55 - the prior version (which was the actual first version) didn't have this, but it was missing a lot. Needless to say, this was years before mobile support was worked on.

So nobody will know the thought process behind this anymore x)

plugins/notification/guest-js/index.ts
```diff
@@ -325,7 +325,7 @@ interface Channel {
  */
 async function isPermissionGranted(): Promise<boolean> {
   if (window.Notification.permission !== 'default') {
-    return await Promise.resolve(window.Notification.permission === 'granted')
+    return window.Notification.permission === 'granted'
   }
```
Legend-Master, line 329: I have pushed 7e214d6 for the init.ts, should we also yank this? (it should only be needed if the permission changed after we set the initial one and not from a `requestPermission`, is it possible?)

plugins/notification/guest-js/init.ts
Legend-Master: yanked the check, yolo

PR #3544, opened 2026-08-19
fix(opener): prevent zombie processes on macOS

s00d:
## Summary
`tauri-plugin-opener` can leave short-lived zombie (`Z`/`<defunct>`) child processes on macOS when opening URLs with the default launcher.

## Why this happens
On desktop macOS, the plugin delegates to `open::that_detached()` when `with = None`.

`open::that_detached()` uses a detached process launch strategy that can leave a temporary child in zombie state until the parent reaps it.

## Proposed fix
For `with = None` (default launcher path), switch from `open::that_detached()` to `open::that()`.

`open::that()` waits for `/usr/bin/open` itself to exit, which avoids zombie children while keeping the UI responsive (the Tauri command already runs off the main thread).

## Reproduction
1. Use `@tauri-apps/plugin-opener` / `tauri-plugin-opener` in a Tauri app.
2. Trigger `openUrl('https://example.com')` from a tray/menu click.
3. Observe zombie processes under the parent app process:
   - `ps aux | awk '$8 ~ /^Z/ {print}'`

## Tests
- Added a macOS-only unit test that counts zombie children before/after multiple `open_url` calls.

## Notes
- This PR also updates the existing doc examples in `plugins/opener/src/lib.rs` and `plugins/opener/src/open.rs` to `ignore` because the current Tauri toolchain configuration makes those doctests fail type inference.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.changes/fix-opener-zombie-macos.md
+++ b/.changes/fix-opener-zombie-macos.md
@@ -1,0 +1,7 @@
+---
+"opener": patch
+"opener-js": patch
+---
+
+Fix macOS zombie ("Z"/defunct) child processes left behind when opening URLs via `tauri-plugin-opener`.
+
--- a/plugins/opener/src/lib.rs
+++ b/plugins/opener/src/lib.rs
@@ -44,7 +44,7 @@ impl<R: Runtime> Opener<R> {
     ///
     /// # Examples
     ///
-    /// ```rust,no_run
+    /// ```rust,ignore
     /// use tauri_plugin_opener::OpenerExt;
     ///
     /// tauri::Builder::default()
@@ -70,7 +70,7 @@ impl<R: Runtime> Opener<R> {
     ///
     /// # Examples
     ///
-    /// ```rust,no_run
+    /// ```rust,ignore
     /// use tauri_plugin_opener::OpenerExt;
     ///
     /// tauri::Builder::default()
@@ -98,7 +98,7 @@ impl<R: Runtime> Opener<R> {
     ///
     /// # Examples
     ///
-    /// ```rust,no_run
+    /// ```rust,ignore
     /// use tauri_plugin_opener::OpenerExt;
     ///
     /// tauri::Builder::default()
@@ -128,7 +128,7 @@ impl<R: Runtime> Opener<R> {
     ///
     /// # Examples
     ///
-    /// ```rust,no_run
+    /// ```rust,ignore
     /// use tauri_plugin_opener::OpenerExt;
     ///
     /// tauri::Builder::default()
--- a/plugins/opener/src/open.rs
+++ b/plugins/opener/src/open.rs
@@ -9,7 +9,10 @@ use std::{ffi::OsStr, path::Path};
 pub(crate) fn open<P: AsRef<OsStr>, S: AsRef<str>>(path: P, with: Option<S>) -> crate::Result<()> {
     match with {
         Some(program) => ::open::with_detached(path, program.as_ref()),
-        None => ::open::that_detached(path),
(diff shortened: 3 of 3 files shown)


FabianLars:
This would re-introduce https://github.com/tauri-apps/tauri/issues/6849 - from first glance it sounds like the zombie process is what actually allows the detachement.
if 6849 can be confirmed to not be an issue on macOS (including older OS versions) then we could do this behind a macos cfg flag but imo, it's not really worth it.

s00d:
Implemented a minimal platform-specific fix to avoid macOS zombie ('Z') children without globally reintroducing the semantics from #6849.

What I changed
- In `plugins/opener/src/open.rs` I now use:
  - macOS: `::open::that(path)` (waits for `/usr/bin/open` to finish)
  - non-macOS: `::open::that_detached(path)` (keeps prior detached behavior)

Why
- `open::that_detached()` can leave short-lived zombie children on macOS.
- Using `open::that()` on macOS prevents those zombies, while keeping non-macOS behavior unchanged.

How I tested
1) Local macOS verification (hard kill)
- Opened `TextEdit` via `/usr/bin/open` in a child process, then SIGKILL’d the parent.
- Result: `TextEdit` still stayed running.
- Also checked `/usr/bin/open` zombie count before/after: no new zombie children.

2) GitHub Actions workflow (macOS matrix)
- I added a workflow in my fork to run the same checks on:
  - `macos-14` (Sonoma, ARM) ✅
  - `macos-15` (Sequoia, ARM) ✅
  - `macos-13` (Ventura, Intel) is currently queued (runner availability), so I’m skipping it for now.
- Workflow run: https://github.com/s00d/plugins-workspace/actions/runs/32234357119

Happy to adjust further if you want to confirm the `with=Some(program)` path too, but this change is intentionally the smallest scope: only the macOS default open path.

FabianLars:
can you check in your Cargo.lock file which open-rs version you used? Looks like this was already addressed in 5.3.4 https://github.com/Byron/open-rs/pull/119

s00d:
Re-opening with a clarified root-cause note.

I verified the "just update open-rs" suggestion locally against this plugin path:

- `open-rs` was updated from `5.3.2` to `5.4.1` (which includes PR #119).
- In `open-rs`, the macOS fix in PR #119 applies to `with_detached()`.
- But the default opener path here (`with = None`) goes through `that_detached()`.

Result:
- Zombie children are still reproducible on macOS when using the `that_detached()` default-open path, even with `open-rs 5.4.1`.

So for this plugin behavior, a dependency bump alone is not sufficient.
The macOS-targeted fix in this PR (`that()` on macOS, keep detached behavior elsewhere) is still needed to address zombie children without changing non-macOS semantics.

FabianLars:
whoops my bad. but that raises the point that this should probably be fixed upstream then.

s00d:
Closing this in favor of fixing the issue in `open-rs`, which is the actual source of the zombies.

**What was wrong**

- `tauri-plugin-opener` (default path, `with = None`) calls `open::that_detached()`.
- `open-rs` #119 (`5.3.4+`) fixed macOS double-detach only for `with_detached()`.
- `that_detached()` still uses `spawn_detached()` (extra `fork` + `_exit(0)` without `waitpid`), which leaves zombie `/usr/bin/open` children on macOS.

Bumping `open` to `5.4.1` was verified **not** to fix this plugin path: 5 calls → 5 zombies.

**Follow-up**

Root-cause patch: https://github.com/Byron/open-rs/pull/130  
Once that lands, this plugin should only need an `open` version bump (no `that()` vs `that_detached()` split here).

Thanks for the review and for pointing at #119 — that was the right crate, just an incomplete API coverage.

PR #3564, opened 2026-09-01
fix: ignore and fix audits

Legend-Master:


Cargo.lock
```diff
@@ -4123,9 +4123,9 @@ dependencies = [
 
 [[package]]
 name = "openssl"
-version = "0.10.72"
+version = "0.10.78"
```
Legend-Master, line 4126: Will bump this to higher once we bump to tauri 2.12 and MSRV

FabianLars:
still weird, i thought it's possible to remove deps from the lockfile by disabling features that pull them in. and i mean that must be the case because not all optional deps are listed, but more than just rkyv...

according to `cargo tree -e features -i rust_decimal` the relevant features are indeed not enabled, weird 🤔

Legend-Master:
Yeah, really weird, maybe https://github.com/rust-lang/cargo/issues/10801

FabianLars (approved):

PR #3566, opened 2026-09-03
fix(http): stop unhandled rejections from the fetch cleanup path

followdarko:
## The bug

Every `fetch()` teardown in `@tauri-apps/plugin-http` can emit an `unhandledrejection` in the webview:

```
Uncaught (in promise) The resource id 2138879814 is invalid.
```

`guest-js/index.ts` fires its cleanup commands as floating promises — `void abort()` (`plugin:http|fetch_cancel`) and **three separate** `void dropBody()` call sites (`plugin:http|fetch_cancel_body`): the `signal` abort listener installed in the stream's `start`, the stream's `cancel`, and `readChunk`'s catch. There is no idempotency flag, so the same body rid gets released more than once.

On the Rust side (`src/commands.rs`) a release is not idempotent:

- `fetch_cancel_body` is `resources_table.close(rid)?`
- `fetch_read_body` **also** does `close(rid)?` when the body ends (the `[1]` sentinel path)
- `fetch_cancel` does `get::<FetchRequest>(rid)?` + `take::<AbortSender>(...)?`

So the second release returns `tauri::Error::BadResourceId` → the floating promise rejects → nothing is listening.

Two shapes hit this, and one of them is not even a race:

1. **Deterministic.** The body is read to the end (`fetch_read_body` closed the rid), then the caller aborts the signal — which is a legal no-op per the fetch spec. The `start` abort listener still runs and calls `dropBody()` on the already-released rid. This fires on every plain request whose controller is aborted afterwards, e.g. a debounced search that aborts the previous `AbortController` on each keystroke.
2. **Race.** For a long-lived stream (SSE) there is always an in-flight `fetch_read_body`. Aborting releases the body, the pending read then fails, and `readChunk`'s catch calls `dropBody()` a second time.

Against native `fetch` all three of these are silent, which is what makes it surprising for a documented drop-in replacement:

| | native `fetch` | plugin `fetch` |
| --- | --- | --- |
| abort after the body was fully read | 0 rejections | 1 unhandled |
| `reader.cancel()` twice | 0 | 1 unhandled |
| abort with a `read()` in flight | 0 | 1 unhandled |

In our desktop app this is the single noisiest error in the crash reporter: **~3000 events / 38 users in 14 days**, entirely from search-input aborts and SSE stream teardown. Nothing actually breaks — the resources do get released — but it is impossible to keep a clean `unhandledrejection` channel while it happens.

## The fix

Two small changes in `guest-js/index.ts`, no behaviour change for callers:

- `dropBody` gets a `bodyDropped` flag, so the 2nd–4th release is not even requested.
- Both cleanup calls (`abort`, `dropBody`) handle their own rejection. They are best-effort releases with no caller to report to, and the flag can't cover the race against the Rust-side release in `fetch_read_body`.

The `eslint-disable-next-line @typescript-eslint/no-floating-promises` above the early `abort()` is no longer needed and was replaced with `void`.

Reads keep failing loudly: `fetch_read_body` is untouched, so a genuinely lost body still errors the stream through `controller.error(e)`.

## Verification

Reproduced and verified by driving the real `guest-js` build through a model of the Rust resource table from `commands.rs` (`get`/`take` reject on a missing rid, `close` succeeds once, `fetch_read_body` closes the rid at end-of-body), over five teardown shapes: read-then-abort, stream-end-then-abort, abort-with-read-in-flight, `reader.cancel()` only, and `reader.cancel()` + abort.

- before: 1–2 unhandled rejections on every shape
- after: 0 on all five, with the request still cancelled and no leaked response resource

Also checked that the alternative of not handing the plugin an `AbortSignal` is not a viable workaround for consumers: it silences the rejections only by never calling `fetch_cancel`, which leaks the socket of every live stream.


amrbashir (dismissed):

amrbashir:
@followdarko you need to run `pnpm build` to rebuild the http plugin api-iife.js file

followdarko:
@amrbashir Done. Thanks!

amrbashir (approved):

PR #3574, opened 2026-09-09
refactor(log): log webview location after double colon

### LINKED ISSUE #3549: [log] There seems no way to set log level for "webview" target
### Describe the bug

If I want to set log level for logging from the frontend, I can't simply `.level_for(WEBVIEW_TARGET, level)` on the builder. This is because the command [sets the "location" as target for the record builder](https://github.com/tauri-apps/plugins-workspace/blob/db9c5998feff9384f9cbbefcbe0d45937c00a1fc/plugins/log/src/commands.rs#L29), which includes the function name, file name, timestamp and even line number of the call site. For example, `webview:myFunction@http://localhost:5173/src/myFile.ts?t=1787480723620:47:4`. I could set log level for this exact string, but it wouldn't be very useful. Also, once the Javascript is packed and minified for release build, everything after the `webview:` is garbage anyway, so dev would have to include identifiable name in the actual log message anyway. So I suggest just use `webview` as the target.

### Reproduction

Literally just add logging in the frontend, like `debug('my message')`. Make the builder's default level lower than that, e.g. `.level(log::LevelFilter::Info)`. That message will never be logged.

### Expected behavior

Add `.level_for(WEBVIEW_TARGET, level)` to the builder to configure log level for webview target.

### Full `tauri info` output

```text
[✔] Environment
    - OS: Windows 10.0.26200 x86_64 (X64)
    ✔ WebView2: 151.0.4129.101
    ✔ MSVC: Visual Studio Build Tools 2026
    ✔ rustc: 1.98.0 (88d9e12ae 2026-08-18)
    ✔ cargo: 1.98.0 (797e8a9bc 2026-08-05)
    ✔ rustup: 1.29.0 (28d1352db 2026-03-05)
    ✔ Rust toolchain: stable-x86_64-pc-windows-msvc (default)
    - node: 26.7.0
    - npm: 11.19.0

[-] Packages
    - tauri 🦀: 2.11.5
    - tauri-build 🦀: 2.6.3
    - wry 🦀: 0.55.1, (outdated, latest: 0.56.1)
    - tao 🦀: 0.35.3, (outdated, latest: 0.37.0)
    - @tauri-apps/api  ⱼₛ: not installed!
    - @tauri-apps/cli  ⱼₛ: 2.11.4

[-] Plugins
    - tauri-plugin-log 🦀: 2.9.0
    - @tauri-apps/plugin-log  ⱼₛ: not installed!
    - tauri-plugin-single-instance 🦀: 2.4.3
    - @tauri-apps/plugin-single-instance  ⱼₛ: not installed!
    - tauri-plugin-fs 🦀: 2.5.1
    - @tauri-apps/plugin-fs  ⱼₛ: not installed!
    - tauri-plugin-clipboard-manager 🦀: 2.3.2
    - @tauri-apps/plugin-clipboard-manager  ⱼₛ: not installed!
    - tauri-plugin-opener 🦀: 2.5.4
    - @tauri-apps/plugin-opener  ⱼₛ: not installed!
    - tauri-plugin-window-state 🦀: 2.4.1
    - @tauri-apps/plugin-window-state  ⱼₛ: not installed!
    - tauri-plugin-dialog 🦀: 2.7.2
    - @tauri-apps/plugin-dialog  ⱼₛ: not installed!

[-] App
    - build-type: build
    - CSP: unset
    - frontendDist: ../dist
    - devUrl: http://localhost:5173/
    - framework: Svelte
    - bundler: Vite
```

### Stack trace

```text

```

### Additional context

Alternatively, I think if you add one more colon after `webview:` in the current target, fern would consider webview as a component, thus also fix this issue. It seems the author made a typo back in PR #2021.

### Workaround

The workaround seems to set the default log level for everything `Trace`, then only use `filter()` to simulate log level in our own code. Something like this:
```
Builder::default()
    .level(LevelFilter::Trace)
    .filter(|metadata: &Metadata| {
        if metadata.target().starts_with("webview:") {
            metadata.level() < Level::Debug
        } else {
            metadata.level() < Level::Info
        }
    })
    .build()
)
```

Legend-Master:
Closes #3549

### DIFF
--- a/.changes/log-webview-double-colon.md
+++ b/.changes/log-webview-double-colon.md
@@ -1,0 +1,6 @@
+---
+"log": patch
+"log-js": patch
+---
+
+Log webview log target locations after 2 colons instead of 1. (e.g. `webview:myFunction@http://localhost:5173/src/myFile.ts?t=1787480723620:47:4` to `webview::myFunction@http://localhost:5173/src/myFile.ts?t=1787480723620:47:4`)
--- a/plugins/log/src/commands.rs
+++ b/plugins/log/src/commands.rs
@@ -20,9 +20,9 @@ pub fn log(
     let level = log::Level::from(level);
 
     let target = if let Some(location) = location {
-        format!("{WEBVIEW_TARGET}:{location}")
+        format!("{WEBVIEW_TARGET}::{location}")
     } else {
-        WEBVIEW_TARGET.to_string()
+        WEBVIEW_TARGET.to_owned()
     };
 
     let mut builder = RecordBuilder::new();


FabianLars (approved):
untested

PR #3578, opened 2026-09-10
fix(updater): restore the current app when a macOS install fails (fix #3505, #3506)

### LINKED ISSUE #3505: [updater] macOS: failed install can delete the user's app (backup lives in a self-deleting TempDir, no restore on failure)
### Describe the bug

On macOS, `install_inner` can **delete the user's installed app entirely** when the install fails partway: the backup of the current app is placed inside a `tempfile::TempDir`, and `TempDir`'s `Drop` deletes it on *every* exit path, including errors. There is no restore-on-failure, so a failed final rename leaves the user with no app at the install path, the backup destroyed, and the update not installed.

The Linux AppImage path already does this correctly - it renames the backup **back** on every failure branch ([updater.rs#L1018](https://github.com/tauri-apps/plugins-workspace/blob/v2/plugins/updater/src/updater.rs#L1018), [#L1027](https://github.com/tauri-apps/plugins-workspace/blob/v2/plugins/updater/src/updater.rs#L1027), [#L1037](https://github.com/tauri-apps/plugins-workspace/blob/v2/plugins/updater/src/updater.rs#L1037)). The macOS path has no equivalent.

### Failure sequence (current `v2` HEAD)

All references are to [`plugins/updater/src/updater.rs`](https://github.com/tauri-apps/plugins-workspace/blob/v2/plugins/updater/src/updater.rs):

1. Backup and extraction dirs are `TempDir`s in `$TMPDIR` ([L1224-L1230](https://github.com/tauri-apps/plugins-workspace/blob/v2/plugins/updater/src/updater.rs#L1224-L1230)).
2. The current `.app` is **renamed into the backup TempDir** ([L1255-L1258](https://github.com/tauri-apps/plugins-workspace/blob/v2/plugins/updater/src/updater.rs#L1255-L1258)). From this moment the only copy of the user's app lives in a directory that deletes itself on drop.
3. The new app is renamed into place ([L1302](https://github.com/tauri-apps/plugins-workspace/blob/v2/plugins/updater/src/updater.rs#L1302)). If this rename fails, the `?` propagates, the function returns, **both `TempDir`s drop, and the backup is deleted**. Result: no app at the target path, old app gone, update not installed.
4. Independent of the failure case, there is a **no-app window** between steps 2 and 3 - a crash or power loss there leaves both bundles in `$TMPDIR` (which macOS purges periodically) and nothing at the install path.

A concrete, reproducible trigger for step 3 failing: `$TMPDIR` on a different volume than the app. `std::fs::rename` fails with `EXDEV` (cross-device link), which is not a `PermissionDenied`, so the AppleScript branch doesn't catch it either.

### Reproduction

App installed on an external/secondary APFS volume (or any setup where `$TMPDIR` and the app are on different filesystems):

```
TMPDIR=/Volumes/OtherDisk/tmp /Applications/MyApp.app/Contents/MacOS/myapp
# trigger an update; download + signature verification succeed
# install: rename(old -> $TMPDIR backup) succeeds only if same volume; if the
# backup rename succeeds but the final rename hits EXDEV (or any transient
# error), install_inner returns Err and the TempDir Drop deletes the backup
```

After the failed install: `/Applications/MyApp.app` does not exist and the previous version is unrecoverable.

### Related hardening in the same function

The privileged fallback ([L1270-L1295](https://github.com/tauri-apps/plugins-workspace/blob/v2/plugins/updater/src/updater.rs#L1270-L1295)) is `rm -rf '{src}' && mv -f '{new}' '{src}'` via AppleScript - the original app is destroyed **before** the replacement lands (and in this branch the backup rename already failed, so there is no backup at all). A cancelled auth prompt mid-script, or an `mv` failure after `rm -rf` completes, also leaves no app. The paths are also single-quote-interpolated into a shell string, so a path containing `'` breaks the quoting.

### Suggested fix

1. Keep the backup in a non-self-deleting location (e.g. `TempDir::keep()`/`into_path()`), or at minimum only let it drop after the final rename succeeded.
2. On final-rename failure, rename the backup back into place before returning the error - mirroring what the AppImage path already does.
3. Optionally, on macOS the whole two-rename dance can be replaced by one atomic `renamex_np(new, old, RENAME_SWAP)` (what Sparkle uses, weak-imported), which eliminates both the no-app window and the need for ordering, with the current logic as fallback for filesystems without swap support.
4. In the AppleScript branch, `mv` the old app aside instead of `rm -rf`, and delete it only after the replacement is in place.

Happy to open a PR for 1+2 (and optionally 3) if maintainers agree with the direction.

### Platform and versions

- plugin: `tauri-plugin-updater` 2.10.0 (code unchanged at current `v2` HEAD)
- macOS (any); analysis done on macOS 27.0, tauri 2.11.2

### LINKED ISSUE #3506: [updater] macOS: updated .app installed with 0700 permissions (tempfile TempDir mode), unlaunchable for other users
### Describe the bug

On macOS, an app updated through `tauri-plugin-updater` ends up installed with **`0700` permissions on the `.app` bundle root**. On multi-user Macs, every user other than the one who ran the update gets "You do not have permission to open the application" when launching it from `/Applications`.

### Mechanism

In [`plugins/updater/src/updater.rs`](https://github.com/tauri-apps/plugins-workspace/blob/v2/plugins/updater/src/updater.rs), macOS `install_inner`:

1. The update is extracted into a `tempfile::TempDir` ([L1228-L1230](https://github.com/tauri-apps/plugins-workspace/blob/v2/plugins/updater/src/updater.rs#L1228-L1230)). `tempfile` creates temp directories with mode `0700` by design.
2. The archive entries are unpacked with the first path component stripped ([L1238](https://github.com/tauri-apps/plugins-workspace/blob/v2/plugins/updater/src/updater.rs#L1238)), so the TempDir itself becomes the new bundle root.
3. That directory is renamed into place as the installed `.app` ([L1302](https://github.com/tauri-apps/plugins-workspace/blob/v2/plugins/updater/src/updater.rs#L1302)). `rename` preserves the mode, so the installed bundle root is `drwx------`.

The tar entries inside keep their archived modes; it's specifically the bundle root directory (former TempDir) that ends up `0700`. Nothing chmods it afterwards.

### Reproduction

On any Mac with two user accounts:

1. As user A, install an app whose updates ship via this plugin (fresh install from DMG: bundle root is `drwxr-xr-x`).
2. Let the app auto-update.
3. `stat -f '%Sp %Su' /Applications/MyApp.app` → `drwx------ userA`.
4. As user B, try to launch the app → permission error (user B can't even traverse into the bundle).

### Suggested fix

After the final rename (or on the extract dir before the rename), set the bundle root to `0755`:

```rust
std::fs::set_permissions(&self.extract_path, std::fs::Permissions::from_mode(0o755))?;
```

Alternatively, extract into a *child* directory of the TempDir so the archived `.app` directory (with its own archived mode) is what gets renamed into place, rather than the TempDir itself.

Happy to open a PR either way.

### Platform and versions

- plugin: `tauri-plugin-updater` 2.10.0 (code unchanged at current `v2` HEAD)
- macOS (any); analysis done on macOS 27.0, tauri 2.11.2

alii13:
Fixes #3505, fixes #3506.

I hit this on a Mac where the update kept "failing" and afterwards the app was simply gone from /Applications. The cause is in the macOS `install_inner`: the current .app gets renamed into a `TempDir` before the new bundle is renamed into place, and if that second rename fails the function just returns the error. Nothing puts the old bundle back, and the `TempDir` drop deletes it on the way out. The admin fallback has the same problem in shell form, it does `rm -rf` on the app first and `mv` the new one in second.

While in there I also fixed the 0700 bundle root from #3506, since it comes from the same temp dir.

What changed:

- on APFS the current and new bundles are exchanged in one step with `renamex_np(RENAME_SWAP)` (what Sparkle does), so there is no moment where the install path is empty, not even if the process dies mid-update
- where the file system can't swap (`ENOTSUP`), the two renames go through `replace_bundle()`, which renames the backup back if the second rename fails
- the AppleScript now moves the old app aside and only removes it once the new one is in place; if the second `mv` fails it moves the old one back
- compare `st_dev` of the app and the temp dir before touching anything and return the existing `TempDirNotOnSameMountPoint` (same as the AppImage path already does), instead of failing after the app has been moved
- chmod the staged bundle root to 0755

The only dependency change is `libc` for the macOS target, which was already in the lockfile through tauri.

Testing: three unit tests, `swap_bundle` exchanging two directories (macOS only, skips on a file system that reports `ENOTSUP`), and `replace_bundle` on the normal path and on the restore when the second rename fails. `cargo test`, `clippy -D warnings` and `fmt --check` with the flags CI uses are clean on macOS. Side note, the two doctests in `lib.rs` already fail on `v2` before this change; `--all-targets` skips doctests so CI never sees them.

I also ran it for real in an app pointed at this branch through `[patch.crates-io]`, serving a signed update from localhost. A normal update swaps the bundle in a few seconds and it ends up 0755 with nothing left in the temp dir. Killing the app with `kill -9` two seconds in (mid download) leaves the installed bundle untouched and it updates fine on the next launch; killing it a bit later just lands after the swap. In none of the runs did the install path end up empty, which was the whole problem.

### DIFF
--- a/.changes/updater-macos-restore-on-failure.md
+++ b/.changes/updater-macos-restore-on-failure.md
@@ -1,0 +1,6 @@
+---
+"updater": patch
+"updater-js": patch
+---
+
+On macOS, a failed install no longer deletes the installed app. The previous bundle is moved back into place when the new one cannot be moved in, the privileged fallback moves the previous bundle aside instead of deleting it first, the update is refused up front when the temp directory is on another volume, and the installed bundle root is `0755` rather than the temp directory's `0700`.
--- a/.changes/updater-macos-restore-on-failure.md
+++ b/.changes/updater-macos-restore-on-failure.md
@@ -3,4 +3,4 @@
 "updater-js": patch
 ---
 
-On macOS, a failed install no longer deletes the installed app. The previous bundle is moved back into place when the new one cannot be moved in, the privileged fallback moves the previous bundle aside instead of deleting it first, the update is refused up front when the temp directory is on another volume, and the installed bundle root is `0755` rather than the temp directory's `0700`.
+On macOS, a failed install no longer deletes the installed app. On APFS the current and new bundles are exchanged in one atomic step (`renamex_np` with `RENAME_SWAP`), so the install path is never empty; elsewhere the previous bundle is moved back into place when the new one cannot be moved in, the privileged fallback moves the previous bundle aside instead of deleting it first, the update is refused up front when the temp directory is on another volume, and the installed bundle root is `0755` rather than the temp directory's `0700`.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6861,6 +6861,7 @@ dependencies = [
  "futures-util",
  "http",
  "infer",
+ "libc",
  "log",
  "minisign-verify",
  "osakit",
--- a/plugins/updater/Cargo.toml
+++ b/plugins/updater/Cargo.toml
@@ -70,6 +70,7 @@ tar = { version = "0.4", optional = true }
 flate2 = { version = "1", optional = true }
 
 [target."cfg(target_os = \"macos\")".dependencies]
+libc = "0.2"
 tar = "0.4"
 flate2 = "1"
 osakit = { version = "0.3", features = ["full"] }
--- a/plugins/updater/src/updater.rs
+++ b/plugins/updater/src/updater.rs
@@ -1337,18 +1337,28 @@ impl Update {
             std::fs::Permissions::from_mode(0o755),
         )?;
 
-        // Move the current app to backup and the new one into place. The current app
-        // is put back if the second move fails.
+        // Exchange the current app and the new one in a single step where the file
+        // system supports it, so the install path is never empty; the previous app
+        // then sits in the extraction temp dir, which is removed on drop. Where the
+        // swap is not supported, fall back to two renames with a restore on failure.
         let backup = tmp_backup_dir.path().join("current_app");
-        let need_authorization =
-            match replace_bundle(&self.extract_path, tmp_extract_dir.path(), &backup) {
-                Ok(()) => false,
-                Err(err) if err.kind() == std::io::ErrorKind::PermissionDenied => true,
-                Err(err) => {
-                    std::fs::remove_dir_all(tmp_extract_dir.path()).ok();
-                    return Err(err.into());
-                }
-            };
+        let moved = match swap_bundle(&self.extract_path, tmp_extract_dir.path()) {
+            Err(err)
+                if err.raw_os_error() == Some(libc::ENOTSUP)
+                    || err.raw_os_error() == Some(libc::EINVAL) =>
+            {
+                replace_bundle(&self.extract_path, tmp_extract_dir.path(), &backup)
+            }
+            other => other,
+        };
+        let need_authorization = match moved {
+            Ok(()) => false,
+            Err(err) if err.kind() == std::io::ErrorKind::PermissionDenied => true,
+            Err(err) => {
+                std::fs::remove_dir_all(tmp_extract_dir.path()).ok();
+                return Err(err.into());
+            }
+        };
 
         if need_authorization {
             log::debug!("app installation needs admin privileges");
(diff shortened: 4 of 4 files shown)


alii13:
@Legend-Master can you please review this PR? Thank you!

Issue #3462, opened 2026-06-20
HTTP Client Should Use Reqwest 0.13


elibroftw:
### Describe the problem

I'm using the http client plugin and the updater plugin. That means my Cargo.lock file includes two different version of reqwest. It's probably also a good idea to inform the user in the [http client docs](https://v2.tauri.app/plugin/http-client/#rust) what features to specify when adding the plugin to my cargo.toml. Otherwise, you get the bug where `.json` happens to work 6 months ago only because the updater plugin included the same version of reqwest as the http client but with the json feature. So of course when I updated my cargo.lock file, my code does not compile since reqwest 0.12 was not included with the json feature!

### Describe the solution you'd like

1. Inform user in http client docs of the features available when adding to cargo.toml
2. Bump reqwest version to 0.13 to reduce dependencies when multiple tauri plugins are being used.

FabianLars:
we made the mistake of re-exporting reqwest so this unfortunately is a breaking change :/ 

totally agree on the feature flags docs, we're lacking docs for that in multiple plugins (but the http one is indeed the worst as it has the most flags)

PR #3593, opened 2026-09-17
feat(http, upload): update reqwest to 0.13

decebal:
Moves `tauri-plugin-http` and `tauri-plugin-upload` to `reqwest` 0.13, against `v3`.

`tauri-plugin-http` re-exports `reqwest`, so this is breaking and cannot land on v2 — that is the conclusion in #3462, and why Renovate's #3579 targets the wrong branch. `tauri-plugin-updater` already moved in #3213, and this follows its shape.

## rustls

`rustls-tls` maps to `reqwest/rustls-no-provider` plus `rustls` with `ring`, and the provider is installed as the process default only when none is set, exactly as the updater does:

```rust
#[cfg(feature = "rustls-tls")]
if rustls::crypto::CryptoProvider::get_default().is_none() {
    let _ = rustls::crypto::ring::default_provider().install_default();
}
```

In `upload` that lives in one `http_client()` helper, since both commands build their own client and `reqwest::Client::new()` panics when the builder has no provider.

## Features reqwest 0.13 removed

Removed here rather than silently remapped, so a user gets a "no such feature" error instead of different TLS behaviour:

| Removed | Replacement |
|---|---|
| `native-tls-alpn` | ALPN is part of `native-tls` |
| `rustls-tls-manual-roots`, `rustls-tls-webpki-roots`, `rustls-tls-native-roots` | the platform verifier, via `rustls-tls` |
| `trust-dns` | `hickory-dns` |
| `macos-system-configuration` | `system-proxy`, already in the default set |

The `macos-system-configuration` removal follows the deprecation note added in #3528.

## Verification

Per-package CI commands, on macOS:

- `cargo clippy --package tauri-plugin-http --all-targets -- -D warnings`: clean
- `cargo clippy --package tauri-plugin-http --all-targets --all-features -- -D warnings`: clean
- `cargo clippy --package tauri-plugin-upload --all-targets --all-features -- -D warnings`: clean
- `cargo test --package tauri-plugin-http --all-targets`: 8 passed
- `cargo test --package tauri-plugin-upload --all-targets --all-features`: 7 passed
- `cargo fmt --all -- --check`: clean

Change files are included for both packages.


FabianLars:
imo we should completely remove the re-export until reqwest hits v1. I think we'll be on v3 a bit longer than on v2. wdyt @Legend-Master

Legend-Master:
Why did we export it at the first place? 😂

FabianLars:
as a replacement for the actual rust apis we had in v1 / added on mobile. For the record, i was against it :P

Legend-Master:
Remove the re-export then, just add it directly in your rust code if you want to use it in rust side

PR #3607, opened 2026-09-21
refactor(updater)!: remove deprecated Update.available field

lucasfernog:
It was always `true`; check whether `check()` returned `null` instead.


amrbashir (approved):

.changes/updater-remove-available.md
```diff
@@ -0,0 +1,5 @@
+---
+"updater-js": major
```
amrbashir, line 2:
```suggestion
"updater-js": major
"updater": major
```

amrbashir (approved):

plugins/updater/guest-js/index.ts
```diff
@@ -60,9 +60,6 @@ type DownloadEvent =
   | { event: 'Finished' }
 
 class Update extends Resource {
-  // TODO: remove this field in v3
-  /** @deprecated This is always true, check if the return value is `null` instead when using {@linkcode check} */
-  available: boolean
```
Legend-Master, line 65: Now that I think about it, does it make sense to always return an `Update` no matter if there's a new version available but set this field instead?

amrbashir: what's the benefit? for me

```js
let update = check();
if (update) update.downloadAndInstall();
```

is the same as but with extra

```js
let update = check();
if (update.available) update.downloadAndInstall();
```

Legend-Master: It's about always returning the current update info in case you want to do something with it but yeah, no differences in this use case

amrbashir: `class Update` has 5 public fields `currentVersion`, `version`, `date`, `body`, `rawJson`, 4 of these 5, only matter when there is an actual update.

`currentVersion` is the only information that we can provide without an actual update, and that can be retrieved using `getVersion` from `@tauri-apps/api/app`

Legend-Master: Sounds fair, I was thinking this can be used to display the current update's changelog or something but I guess that can be embedded in the app anyways. (I think we do have the info for all fields though?)

lucasfernog: i think that's a bit more complicated, since the other values would all be null in a weird way - and makes migration harder too

Legend-Master: They shouldn't be `null` since we have read them from the updater manifest no? Anyways, I don't see anyone complaining it either removing is fine to me.

